PR #60, opened 2019-12-05
Add initContainers to default configuration's `templateRules`

lirsacc:
This change fixes https://github.com/k14s/kapp/issues/59 by adding entries to the default config which mirror the ones for `containers`. Before that the ConfigMap and Secret references would not be updated when using versioned resources. Also adds `initContainers` to the existing e2e test.


cppforlife:
@lirsacc thanks!

PR #88, opened 2020-03-18
Bump version to 0.22.0

dbirks:
Just noticed the version was behind when building the most recent release. Or maybe this should be `0.22.1` or `0.23.0` to catch the next release? Whatever you'd like of course, just wanted to make the issue since I noticed it. Thanks :slightly_smiling_face:

### DIFF
--- a/pkg/kapp/cmd/version.go
+++ b/pkg/kapp/cmd/version.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	Version = "0.21.0"
+	Version = "0.22.0"
 )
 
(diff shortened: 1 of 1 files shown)


cppforlife:
looks like commit pushing failed on my side so i tagged commit before version commit. v0.22.0 published binaries do have a correct version. just re-run git push so latest develop should be good. thanks for noticing.

dbirks:
Sounds good, thanks :+1:

dbirks:
Sounds good, thanks :+1:

PR #97, opened 2020-04-06
[KAPP-71] Watch pods in specific namespace when needed

jessehu:
When a namespace-scoped RBAC user is used by kubectl, and this
user doesn't have privilege to watch pods in all namespaces,
we should only watch the namespace specified by '-n' flag or
default to the namespace specified in kubeconfig.

Manually tested on Mac OS and the error msg is gone:
`./kapp deploy -a myapp -f examples/gitops/guestbook/`

Close https://github.com/k14s/kapp/issues/71

### DIFF
--- a/pkg/kapp/resources/identified_resources_pods.go
+++ b/pkg/kapp/resources/identified_resources_pods.go
@@ -2,15 +2,15 @@ package resources
 
 import (
 	"fmt"
-
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 )
 
 func (r IdentifiedResources) PodResources(labelSelector labels.Selector) UniquePodWatcher {
-	return UniquePodWatcher{labelSelector, r.resources.fallbackAllowedNamespaces, r.coreClient}
+	return UniquePodWatcher{labelSelector, r.fallbackAllowedNamespaces, r.coreClient}
 }
 
 type PodWatcherI interface {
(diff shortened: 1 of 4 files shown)


pkg/kapp/resources/identified_resources_pods.go
```diff
@@ -28,8 +29,15 @@ func (w UniquePodWatcher) Watch(podsToWatchCh chan corev1.Pod, cancelCh chan str
 	nonUniquePodsToWatchCh := make(chan corev1.Pod)
 
 	go func() {
+		namespace := ""
+		if len(w.fallbackAllowedNamespaces) > 0 {
```
cppforlife, line 33: fallbackAllowedNamespaces is always set which means that we would always limit it to this namespace. we should use it when `errors.IsForbidden(err)` is true for the Watch call.

pkg/kapp/resources/identified_resources_pods.go
```diff
@@ -10,16 +10,17 @@ import (
 )
 
 func (r IdentifiedResources) PodResources(labelSelector labels.Selector) UniquePodWatcher {
-	return UniquePodWatcher{labelSelector, r.coreClient}
+	return UniquePodWatcher{labelSelector, r.resources.fallbackAllowedNamespaces, r.coreClient}
```
cppforlife, line 13: instead of reaching into resources object, let's save off fallbackAllowedNamespaces onto IdentifiedResources directly and use it here: https://github.com/k14s/kapp/blob/0a2087f8225c9ea650c6244d71c3b5af20fbbfc2/pkg/kapp/resources/identified_resources.go#L21

jessehu: Thanks @cppforlife. Do you mean like this? Then fallbackAllowedNamespaces is duplicated with resources.fallbackAllowedNamespaces.

`return IdentifiedResources{coreClient, resourceTypes, resources, fallbackAllowedNamespaces, logger.NewPrefixed("IdentifiedResources")}`

jessehu: Or remove fallbackAllowedNamespaces from `resources` additionaly?

pkg/kapp/resources/identified_resources_pods.go
jessehu: Got it. Can we assume fallbackAllowedNamespaces[0] should be used for watch pods if fallbackAllowedNamespaces[0] is not "default"?

jessehu: @cppforlife I added a new commit to follow your advise.

cppforlife:
closed-reopened PR for test-gh workflow to run.

Issue #103, opened 2020-04-20
Tail specific container logs


eLco:
Hi,

this isn't much of the bug, but more like a feature.

In the case of multiple containers per pod, someone may want to tail logs of specific containers in pod.
This can be the case during deployment of application with an Istio sidecar container for example, where developer may only want to see his/her application container logs, not istio-proxy wall of logs.

This relates to both kapp logs command and to kapp deploy --logs flag.
For the kapp deploy command new annotation can be used to include container names -
"kapp.k14s.io/container-logs: "container-name1, container2-name,.." OR current annotation can be reused 'kapp.k14s.io/deploy-logs: ""' for all container logs and "kapp.k14s.io/deploy-logs: "container-name1, container2-name,.." for specific ones.

I'm happy to work on this feature.

cppforlife:
> For the kapp deploy command new annotation can be used to include container names -
"kapp.k14s.io/container-logs: "container-name1, container2-name,.." OR current annotation can be reused 'kapp.k14s.io/deploy-logs: ""' for all container logs and "kapp.k14s.io/deploy-logs: "container-name1, container2-name,.." for specific ones.

let's add a new annotation `kapp.k14s.io/deploy-logs-container-names: "name1,name2..."`

eLco:
Both parts of this feature were merged.

cppforlife:
included in https://github.com/k14s/kapp/releases/tag/v0.25.0

PR #104, opened 2020-04-20
Kapp logs option to tail only specific containers in pod

eLco:
Adding initial support for specific container logs in kapp logs command output.

Related to #103.

### DIFF
--- a/pkg/kapp/cmd/app/logs_flags.go
+++ b/pkg/kapp/cmd/app/logs_flags.go
@@ -8,11 +8,11 @@ import (
 )
 
 type LogsFlags struct {
-	Follow        bool
-	Lines         int64
-	ContainerName []string
-	ContainerTag  bool
-	PodName       string
+	Follow         bool
+	Lines          int64
+	ContainerNames []string
+	ContainerTag   bool
+	PodName        string
 }
 
 func (s *LogsFlags) Set(cmd *cobra.Command) {
@@ -20,7 +20,7 @@ func (s *LogsFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().Int64Var(&s.Lines, "lines", 10, "Limit to number of lines (use -1 to remove limit)")
 	cmd.Flags().BoolVar(&s.ContainerTag, "container-tag", true, "Include container tag")
 	cmd.Flags().StringVarP(&s.PodName, "pod-name", "m", "", "Set pod name to filter logs (% acts as wildcard, e.g. 'app%')")
-	cmd.Flags().StringSliceVarP(&s.ContainerName, "container-name", "c", nil, "Set container names to filter logs in the pod, separated by comma")
+	cmd.Flags().StringSliceVarP(&s.ContainerNames, "container-name", "c", nil, "Set container names to filter logs in the pod, separated by comma")
 }
 
 func (s *LogsFlags) PodLogOpts() (ctllogs.PodLogOpts, error) {
@@ -29,7 +29,7 @@ func (s *LogsFlags) PodLogOpts() (ctllogs.PodLogOpts, error) {
 			"Expected --lines to be greater than zero since --follow is not specified")
 	}
 
-	opts := ctllogs.PodLogOpts{Follow: s.Follow, ContainerName: s.ContainerName, ContainerTag: s.ContainerTag}
+	opts := ctllogs.PodLogOpts{Follow: s.Follow, ContainerNames: s.ContainerNames, ContainerTag: s.ContainerTag}
 
 	if s.Lines >= 0 {
 		opts.Lines = &s.Lines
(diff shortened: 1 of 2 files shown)


pkg/kapp/cmd/app/logs_flags.go
```diff
@@ -8,17 +8,19 @@ import (
...
+	ContainerTag   bool
+	PodName        string
 }
 
 func (s *LogsFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().BoolVarP(&s.Follow, "follow", "f", false, "As new pods are added, new pod logs will be printed")
 	cmd.Flags().Int64Var(&s.Lines, "lines", 10, "Limit to number of lines (use -1 to remove limit)")
 	cmd.Flags().BoolVar(&s.ContainerTag, "container-tag", true, "Include container tag")
 	cmd.Flags().StringVarP(&s.PodName, "pod-name", "m", "", "Set pod name to filter logs (% acts as wildcard, e.g. 'app%')")
+	cmd.Flags().StringSliceVarP(&s.ContainerNames, "container-names", "c", []string{}, "Set container names to filter logs in the pod, separated by comma")
```
cppforlife, line 23: actual flag name should be singular "container-name" to be more consistent with rest of kapp flags. default could be nil, instead of []string{} (since it's a typed field)

pkg/kapp/cmd/app/logs_flags.go
```diff
@@ -33,5 +35,9 @@ func (s *LogsFlags) PodLogOpts() (ctllogs.PodLogOpts, error) {
 		opts.Lines = &s.Lines
 	}
 
+	if len(s.ContainerNames) > 0 {
```
cppforlife, line 38: i dont think we need to check for 0 length here.

pkg/kapp/logs/pod_log.go
```diff
@@ -76,3 +89,12 @@ func (l PodLog) isWaitingContainer(cont corev1.Container, statuses []corev1.Cont
 	}
 	return false
 }
+
+func (l PodLog) isContainerInSlice(cont corev1.Container, containers []string) bool {
```
cppforlife, line 93: let's change this to return true if there are 0 containers. otherwise check existence in slice.

pkg/kapp/logs/pod_log.go
```diff
@@ -41,12 +42,24 @@ func (l PodLog) TailAll(ui ui.UI, cancelCh chan struct{}) error {
...
+				}
+				continue
+			}
 			conts = append(conts, cont)
 		}
 	}
 
 	for _, cont := range l.pod.Spec.Containers {
 		if !(podInTerminalState && l.isWaitingContainer(cont, l.pod.Status.ContainerStatuses)) {
+			if len(l.opts.ContainerNames) > 0 {
```
cppforlife, line 57: given change below in isContainerInSlice, we can simplify to append if isContainerInSlice return true.

pkg/kapp/logs/pod_log.go
cppforlife: may be also rename isContainerInSlice to isWatchingContainer

eLco:
Thank you for your feedback @cppforlife, I've made suggested changes.

eLco:
Thank you for your feedback @cppforlife, I've made suggested changes.

pkg/kapp/cmd/app/logs_flags.go
```diff
@@ -8,17 +8,19 @@ import (
...
+	ContainerTag  bool
+	PodName       string
 }
 
 func (s *LogsFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().BoolVarP(&s.Follow, "follow", "f", false, "As new pods are added, new pod logs will be printed")
 	cmd.Flags().Int64Var(&s.Lines, "lines", 10, "Limit to number of lines (use -1 to remove limit)")
 	cmd.Flags().BoolVar(&s.ContainerTag, "container-tag", true, "Include container tag")
 	cmd.Flags().StringVarP(&s.PodName, "pod-name", "m", "", "Set pod name to filter logs (% acts as wildcard, e.g. 'app%')")
+	cmd.Flags().StringSliceVarP(&s.ContainerName, "container-name", "c", nil, "Set container names to filter logs in the pod, separated by comma")
```
cppforlife, line 23: oops, should have mentioned that actual field should stay plural (eg ContainerNames), only flag is singular.

eLco: No problem :)

cppforlife:
thanks for contributing @eLco. change will be included in upcoming release. if you are feeling adventerous it probably also makes sense to add something like this `matcher.NewStringMatcher(o.LogsFlags.PodName).Matches(pod.Name)` but for container names to support wildcarding.

eLco:
Good idea, thank you! I will continue to work on this feature.

PR #107, opened 2020-04-21
kapp logs: add support for wildcard filtering in container names

eLco:
Add support for wildcard filtering in container names in kapp logs command output.

Related to #103.


eLco:
Decided to merge this functionality into another PR related to #103 .

PR #108, opened 2020-04-21
Deploy per container logs

eLco:
Add support for wildcard filtering for container names in kapp logs cli command.
Add support for new annotations kapp.k14s.io/deploy-logs-container-names: "name1,name2..." for kapp deploy command to filter log output by container names.

Related #103

### DIFF
--- a/pkg/kapp/cmd/app/logs.go
+++ b/pkg/kapp/cmd/app/logs.go
@@ -71,7 +71,14 @@ func (o *LogsOptions) Run() error {
 		supportObjs.IdentifiedResources.PodResources(labelSelector),
 	}
 
-	logsView := ctllogs.NewView(logOpts, podWatcher, supportObjs.CoreClient, o.ui)
+	contFilter := func(pod corev1.Pod) []string {
+		if len(o.LogsFlags.ContainerNames) > 0 {
+			return o.LogsFlags.ContainerNames
+		}
+		return nil
+	}
+
+	logsView := ctllogs.NewView(logOpts, podWatcher, contFilter, supportObjs.CoreClient, o.ui)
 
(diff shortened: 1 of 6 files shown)


pkg/kapp/cmd/app/logs.go
```diff
@@ -71,7 +71,14 @@ func (o *LogsOptions) Run() error {
 		supportObjs.IdentifiedResources.PodResources(labelSelector),
 	}
 
-	logsView := ctllogs.NewView(logOpts, podWatcher, supportObjs.CoreClient, o.ui)
+	contFilter := func(pod corev1.Pod) []string {
+		if len(o.LogsFlags.ContainerNames) > 0 {
```
cppforlife, line 75: dont think we need to check > 0, can just return.

pkg/kapp/logs/view.go
```diff
@@ -66,6 +68,8 @@ func (v View) Show(cancelCh chan struct{}) error {
 				return fmt.Sprintf("%s > %s", pod.Name, cont.Name)
 			}
 
+			v.tailOpts.ContainerNames = v.contFilterFunc(pod)
```
cppforlife, line 71: im not certain how go works this out with go func()... "modifying" concurrently. let's change it so that we copy tailOpts to be more explicit:

```
tailOpts := v.tailOpts
tailOpts.ContainerNames = v.contFilterFunc(pod)
```

and use it below in NewPodLog

eLco: Agree.

pkg/kapp/cmd/app/logs.go
eLco: Too paranoid :)

PR #110, opened 2020-05-05
Update Secret and ConfigMap reference paths in default rules

acosta11:
## Context
Kapp versioned Secrets and ConfigMaps are not updated within [projected volumes](https://kubernetes.io/docs/concepts/storage/volumes/#projected) or ServiceAccounts by default. Similarly, Secrets of type ImagePullSecrets can have a different key/path in Pod and ServiceAccount specs.

## Proposed Change
Update the default template rules to include object references to the above resource paths so that versioned Secrets and ConfigMaps are updated by default.

## Open Question
Are ImagePullSecrets paths a necessary default directly on Pod specs? I don't know how prevalent their use is.

## Validation
I validated the first two commits by deploying cf-for-k8s with the kapp versioned annotation on all non-sds secrets and kapp was able to deploy successfully with the app functional. I didn't have any instances of nested imagePullSecrets in the deployment I happened to use, so the last commit is just based on the api definitions.

### DIFF
--- a/pkg/kapp/config/default.go
+++ b/pkg/kapp/config/default.go
@@ -244,6 +244,8 @@ templateRules:
       resourceMatchers: *withPodTemplate
     - path: [spec, template, spec, initContainers, {allIndexes: true}, envFrom, {allIndexes: true}, secretRef]
       resourceMatchers: *withPodTemplate
+    - path: [spec, template, spec, imagePullSecrets, {allIndexes: true}]
+      resourceMatchers: *withPodTemplate
     - path: [spec, template, spec, volumes, {allIndexes: true}, secret]
       resourceMatchers: *withPodTemplate
       nameKey: secretName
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@acosta11 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kapp&pullRequestId=110)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=k14s/kapp&pullRequestId=110) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@acosta11 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kapp&pullRequestId=110)!

pkg/kapp/config/default.go
```diff
@@ -242,13 +245,31 @@ templateRules:
       resourceMatchers: *withPodTemplate
     - path: [spec, template, spec, initContainers, {allIndexes: true}, envFrom, {allIndexes: true}, secretRef]
       resourceMatchers: *withPodTemplate
+    - path: [spec, template, spec, imagePullSecrets, {allIndexes: true}]
+      resourceMatchers: *withPodTemplate
+      nameKey: name
```
cppforlife, line 250: `nameKey: name` is default, so no need to add it.

Issue #87, opened 2020-03-17
[apply retry] retry on ServiceUnavailable error


cppforlife:
> kapp: Error: waiting on delete podmetrics/capi-clock-747b748bbb-89pn4 (metrics.k8s.io/v1beta1) namespace: cf-system: errored: Checking existance of resource podmetrics/capi-clock-747b748bbb-89pn4 (metrics.k8s.io/v1beta1) namespace: cf-system: the server is currently unable to handle the request (reason: ServiceUnavailable)

it's not quite clear if ServiceUnavailable is a terminal condition for Exists()

full logs: https://gist.github.com/cppforlife/0cd4c527ab03c68685c32705ebffef88

cppforlife:
included in https://github.com/k14s/kapp/releases/tag/v0.31.0

Issue #82, opened 2020-03-09
[apply retry] Retry if API-server instance is shut down


jkbschmid:
If a k8s instance has multiple API-servers, one of which is being scaled down, existing websocket connections will be terminated and kapp will fail with:

> kapp: Error: waiting on reconcile destinationrule/istio-policy (networking.istio.io/v1alpha3) namespace: istio-system: errored: Listing schema.GroupVersionResource{Group:"apiextensions.k8s.io", Version:"v1beta1", Resource:"customresourcedefinitions"}, namespaced: false: Get https://api.cf4k8s.istio.shoot.canary.k8s-hana.ondemand.com/apis/apiextensions.k8s.io/v1beta1/customresourcedefinitions?labelSelector=kapp.k14s.io%2Fassociation%3Dv1.4ea2426af853c41012e6a53e19d4a3b1: http2: server sent GOAWAY and closed the connection; LastStreamID=64781, ErrCode=NO_ERROR, debug=""

# Feature request

kapp should use retries to cope with API server instances being scaled down.

cppforlife:
interesting, i was pretty sure we already do retries, will check what's going on in this case.

cppforlife:
> http2: server sent GOAWAY and closed the connection

it's interesting that this is how api-server "drains" connection.

jkbschmid:
> it's interesting that this is how api-server "drains" connection.

IMO, the only thing the API-server can do to properly drain connections is to 1) wait for a certain period for the connections to be terminated and, once this period is over 2) send a GOAWAY and terminate connections itself. 
Therefore, kapp should be able to handle such connection terminations.

cppforlife:
included in https://github.com/k14s/kapp/releases/tag/v0.31.0

PR #122, opened 2020-06-29
Improve apply retry

kramerul:
This PR is mostly related to issues #82 and #87.

* `util.Retry` was replaced by `util.RetryOnError` which is a copy of `k8s.io/client-go/util/retry.OnError`. It seems to be the standard way to do retries with newer versions of the k8s client library.
* The two errors `http2.GoAwayError` and `errors.StatusError` with `err.ErrStatus.Reason == metav1.StatusReasonServiceUnavailable` are automatically retried.
* List functions are also retried now

### DIFF
--- a/pkg/kapp/clusterapply/cluster_change.go
+++ b/pkg/kapp/clusterapply/cluster_change.go
@@ -143,7 +143,7 @@ func (c *ClusterChange) Apply() (bool, []string, error) {
 		return false, descMsgs, err
 	}
 
-	retryable := err != nil && ctlres.IsRetriableWebhookError(err)
+	retryable := err != nil && ctlres.IsResourceChangeBlockedErr(err)
 	if retryable {
 		descMsgs = append(descMsgs, uiWaitMsgPrefix+"Retryable error: "+err.Error())
 	}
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@kramerul Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kapp&pullRequestId=122)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=k14s/kapp&pullRequestId=122) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@kramerul Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/kapp&pullRequestId=122)!

cppforlife:
few comments:

IsRetryableErr was meant to represent error that could be retried at a higher level due to resource conflicts (ie one resource waiting for the other). i think we should keep it that way. http2 errors should be retried at a lower level and not leak out to higher level retry mechanisms.

i would also like to stick with util.Retry to keep things very simple for now.

cppforlife:
IsRetryableErr should probably be renamed to something like IsResourceChangeBlockedErr to indicate better that resource change is blocked on somethign else in the system (like another resource being changed).

kramerul:
Hi @cppforlife,

I renamed `IsRetryableErr` to `IsResourceChangeBlockedErr` and reverted the usage of `RetryOnError` although in my point of view the code looked simpler and an exponential backoff is missing now.

PR #162, opened 2020-11-17
Update kubernetes deps

benmoss:
This project is on [a version of k8s from 2018](https://github.com/kubernetes/client-go/releases/tag/v8.0.0) 😱 .

Should we also bump the go version at some point?


benmoss:
Looking into the failures 😸

benmoss:
Bleh, don't have the time to fix this much

PR #168, opened 2020-12-02
Point the URLs from k14s github org to vmware-tanzu

joaopapereira:
Update the website to download the installer to carvel.dev


StevenLocke:
From a quick search with GoLand there's still a few hundred usages of "k14s". We didn't verify whether most of those were accurate, but things like imports will probably need to be updated eventually.

Is this PR scoped only to github URLs and the install.sh script? If so, lgtm. We'll want to make another pass for non-links.

joaopapereira:
We are not changing the package's location for now. This should be addressed in the future after we have a release created.

PR #176, opened 2020-12-18
WIP: dumps resources that do not have labels

### LINKED ISSUE #160: kapp should double check that k8s api returns labeled resources
currently kapp trusts that k8s api returns resources based on the label query specified. it appears that if aggregated api impl come into play there is no guarantee that they return resources filted by label as requested (antrea for example did not implement label selector).

paulczar:
This is a very naive solution to #160 that I'm using in the interim.  Probably too naive to be merged in, but could be useful for folks in the meantime.

Signed-off-by: Paul Czarkowski <username.taken@gmail.com>

### DIFF
--- a/pkg/kapp/resources/identified_resources_list.go
+++ b/pkg/kapp/resources/identified_resources_list.go
@@ -55,6 +55,16 @@ func (r IdentifiedResources) List(labelSelector labels.Selector, resRefs []Resou
 		}
 	}
 
+	// Dump any resources that do not contain Labels
+	// as they cannot be kapp resources
+	var newResources []Resource
+	for _, res := range resources {
+		if len(res.Labels()) > 0 {
+			newResources = append(newResources, res)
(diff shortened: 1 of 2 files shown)


pkg/kapp/resources/identified_resources_list.go
```diff
@@ -55,6 +55,16 @@ func (r IdentifiedResources) List(labelSelector labels.Selector, resRefs []Resou
 		}
 	}
 
+	// Dump any resources that do not contain Labels
+	// as they cannot be kapp resources
+	var newResources []Resource
+	for _, res := range resources {
+		if len(res.Labels()) > 0 {
```
cppforlife, line 62: hmm, this would break usage of `kapp -a label:` or other negative-matching variants.

gcheadle-vmware:
Closing due to fix in #198

PR #177, opened 2020-12-20
feat: added in ability for kapp delete to use custom rules loaded in cluster ConfigMap and extended annotation matcher to optionally consider annotation values as well as keys

chrismellard:


chrismellard:
@cppforlife who is best placed for a review of this PR please?

pkg/kapp/cmd/app/delete.go
```diff
@@ -81,7 +81,7 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
-	_, conf, err := ctlconf.NewConfFromResourcesWithDefaults(nil)
+	_, conf, err := ctlconf.NewConfFromResourcesWithDefaults(existingResources)
```
cppforlife, line 84: this line change makes sense in general though it would only work cases when kapp config is stored in a configmap (not when it's stored with top level kapp kind). i would be interested in seeing what it takes to save deploy config at the time of deploy into app record.

can you think of any downsides of applying config against deletion of objects? any possible ordering snafus?

pkg/kapp/config/resource_matchers.go
```diff
@@ -56,7 +56,7 @@ type KindNamespaceNameMatcher struct {
 }
 
 type HasAnnotationMatcher struct {
-	Keys []string
+	KeyValues []ctlres.KeyValue
```
cppforlife, line 59: this is a backwards incompatible change. let's add KeyValues as an additional key instead of replacing Keys.

cppforlife: could you also separate keyvalues change into its own PR?

pkg/kapp/cmd/app/delete.go
chrismellard: Can  you elaborate on what you mean by an "app record" please? The kapp config in a ConfigMap on deploy feels like what you're describing when you say "record on deploy".

In terms of downsides of  applying config against deletion of objects - perhaps I'm missing something in trying to achieve what I'm trying to achieve - but without the change above I was unable to get any of the deletion ordering rules to  work on `kapp delete` - i.e.  there seemed to be noway for `kapp` to recognise the `kapp` config which was used on `deploy`. Therefore this PR was deliberately raised so I could get deletion ordering rules recognised by `kapp` on delete. Specific need for deletion ordering in my immediate use case is unpicking my way through finalisers and ensuring we don't delete resources that need to hang around to run the finalisers of other resources

pkg/kapp/config/resource_matchers.go
chrismellard: Yep - I'll get this split out in to a non breaking change PR of its own. Just that both of these changes were required for use to try start using `kapp` on the jenkins-x project hence my laziness ;)

pkg/kapp/cmd/app/delete.go
cppforlife: > Can you elaborate on what you mean by an "app record" please

yeah, there is two ways to provide kapp config to kapp today:

- directly via resource
```
kind: Config
...
```

- or, wrapped as a ConfigMap
```
apiVersion: v1
kind: ConfigMap
data:
  config.yml: |
    kind: Config
    ...
```

in the case of ConfigMap it does get saved into the cluster and pulled out with your change. in the case of direct resource it does get discarded hence not available to delete command. the idea here would be to save Config resources into record that kapp maintains per each app (when you do kapp deploy -a foo -f ... kapp will create foo ConfigMap that records generated label for foo app. there we can also save deploy config).

cppforlife:
going to close this PR given it's been sitting here for a while without changes. feel free to reopen if progress is made.

PR #198, opened 2021-03-04
Filter resources by label to catch aggregated api mistakes

gcheadle-vmware:


pkg/kapp/resources/resources.go
```diff
@@ -37,7 +37,17 @@ const (
...
+	All([]ResourceType, AllOpts) ([]Resource, error)
+	Delete(Resource) error
+	Exists(Resource) (bool, error)
+	Get(Resource) (Resource, error)
+	Patch(Resource, types.PatchType, []byte) (Resource, error)
+	Update(Resource) (Resource, error)
+	Create(resource Resource) (Resource, error)
+}
+
+type Impl struct {
```
joaopapereira: Usually, in Go, the `Impl` term is not seen as a good naming practice. Given what this struct is trying to accomplish it looks like a struct that will hold Kubernetes resources so my suggestion is that we rename this struct to `KubernetesResources`

pkg/kapp/resources/resources.go
```diff
@@ -49,10 +59,10 @@ type Resources struct {
 	logger logger.Logger
 }
 
-func NewResources(resourceTypes ResourceTypes, coreClient kubernetes.Interface,
-	dynamicClient dynamic.Interface, fallbackAllowedNamespaces []string, logger logger.Logger) *Resources {
+func NewImpl(resourceTypes ResourceTypes, coreClient kubernetes.Interface,
+	dynamicClient dynamic.Interface, fallbackAllowedNamespaces []string, logger logger.Logger) *Impl {
```
joaopapereira:
```suggestion
	dynamicClient dynamic.Interface, fallbackAllowedNamespaces []string, logger logger.Logger) Resources {
```

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package resources
+
+import (
+	"github.com/cppforlife/go-cli-ui/ui"
```
joaopapereira, line 8: Imports need reorder, the convention we have been following is:
```go
import (
  "strings"
  "testing"
  ....
  
  "github.com/one/two"
)
```

joaopapereira (commented):
Added some comments on the PR. The implementation looks good, more some nitpicks.
Also not 100% convinced on the usage of testify/mock.

pkg/kapp/resources/identified_resources_list.go
```diff
@@ -47,6 +47,15 @@ func (r IdentifiedResources) List(labelSelector labels.Selector, resRefs []Resou
 		return nil, err
 	}
 
+	// Filter resources by label in case of aggregated api https://github.com/vmware-tanzu/carvel-kapp/issues/160
```
cppforlife: "in case of aggregated api" feels like sentence didnt end. how about:

```
// Check returned resources against label selector 
// in case of Kubernetes APIs returned resources that do not match.
// This can happen if custom aggregated APIs did not implement label selector filtering.
// (https://github.com/vmware-tanzu/carvel-kapp/issues/160)
```

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+    spec:
+      containers:
+      - image: nginx:1.14.2
+        name: nginx
+        ports:
+        - containerPort: 80
+          protocol: TCP
+`
+
+	antreaRes, err := NewResourceFromBytes([]byte(antreaBs))
```
cppforlife: you can use MustNewResourceFromBytes here to avoid error checking.

pkg/kapp/resources/resources.go
cppforlife: current implementation is good. there is common saying: "Accept Interfaces and Return Concrete Types". this class can implement other methods that are outside of scope of Resources interface. in general there is no need to couple a constructor to particular interface.

pkg/kapp/resources/resources.go
cppforlife: given the context that we are dealing with Kubernetes to me KubernetesResources is not that appealing. it's not necessary common, but plenty of places use Impl as a suffix for a default implementations. my personal choice would be to roll with ResourcesImpl until there is something more defining about this class that would give us a more specific prefix than Kubernetes.

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+	t.Helper()
+	antreaBs := `---
+apiVersion: clusterinformation.antrea.tanzu.vmware.com/v1beta1
+kind: AntreaControllerInfo
+metadata:
+  name: antrea-controller
+version: v0.10.1
+`
+
+	deploymentBs := `---
```
cppforlife, line 46: it seems that this resource has a lot of stuff that's not really relevant to this test. i would try to significantly trim it to more or less a resources with metadata.labels per our requirement.

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func TestIdentifiedResourcesListReturnsLabeledResources(t *testing.T) {
+	mockedResTypes := new(ResourceTypesMock)
+	mockedRess := new(ResourcesMock)
+
+	mockedRess.Mock.On("All").Return(makeNewResources(t), nil)
+
+	ui := ui.NewNoopUI()
```
cppforlife: can inline this ui into logger since it's not used anywhere.

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+	logger := logger2.NewUILogger(ui)
+	identifiedResources := NewIdentifiedResources(nil, mockedResTypes, mockedRess, []string{}, logger)
+
+	kappLabel := make(map[string]string)
+	kappLabel["kapp.k14s.io/app"] = "app-name"
+	sel := labels.Set(kappLabel).AsSelector()
```
cppforlife: you can write these 3 lines as:

```
sel := labels.Set(map[string]string{"kapp.k14s.io/app": "app-name"}).AsSelector()
```

this gets rid of the noise of constructing objects and gets to the meat of what we are trying to do.

i would also change the label to something else so that it's not implying that this is a special label.

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+		require.Contains(t, res.Labels(), "kapp.k14s.io/app")
+		require.Equal(t, res.Labels()["kapp.k14s.io/app"], kappLabel["kapp.k14s.io/app"])
+	}
+}
+
+type ResourcesMock struct {
+	mock.Mock
+}
+
+func (r *ResourcesMock) All([]ResourceType, AllOpts) ([]Resource, error) {
```
cppforlife: instead of having this object be a mock, i would recommend for now to just inline `makeNewResources` functionality into here. it's just a "dummy" implementation of Resources interfaces that always returns 2 resources for this test.

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+	kappLabel["kapp.k14s.io/app"] = "app-name"
+	sel := labels.Set(kappLabel).AsSelector()
+
+	resources, err := identifiedResources.List(sel, nil)
+
+	mockedRess.AssertCalled(t, "All")
+	require.Nil(t, err)
+	require.NotNil(t, resources)
+
+	for _, res := range resources {
```
cppforlife: given the test setup we know exactly the length of the resources slice (1). the less possibility of assertions silently being skipped the better -- that's pretty common when loops are used without checking on length of the slice.

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+		require.Contains(t, res.Labels(), "kapp.k14s.io/app")
+		require.Equal(t, res.Labels()["kapp.k14s.io/app"], kappLabel["kapp.k14s.io/app"])
```
cppforlife: the more hardcodedness in test assertions the better == less ways it can go wrong. in this case this could be

`require.Equal(t, res.Labels()["kapp.k14s.io/app"], "app-name")`

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,118 @@
...
+}
+func (r *ResourcesMock) Delete(Resource) error                                     { return nil }
+func (r *ResourcesMock) Exists(Resource) (bool, error)                             { return true, nil }
+func (r *ResourcesMock) Get(Resource) (Resource, error)                            { return nil, nil }
+func (r *ResourcesMock) Patch(Resource, types.PatchType, []byte) (Resource, error) { return nil, nil }
+func (r *ResourcesMock) Update(Resource) (Resource, error)                         { return nil, nil }
+func (r *ResourcesMock) Create(Resource) (Resource, error)                         { return nil, nil }
+
+type ResourceTypesMock struct {
+	mock.Mock
```
cppforlife: this doesnt have to be a mock.

pkg/kapp/resources/resources.go
joaopapereira: Yep you are right

pkg/kapp/resources/identified_resources_list_test.go
gcheadle-vmware: Changed to a generic label: `"some-label": "value"`. Hopefully this does not imply anything special about the label.

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,74 @@
...
+	require.Contains(t, resources[0].Labels(), "some-label")
+	require.Equal(t, resources[0].Labels()["some-label"], "value")
+}
+
+type FakeResources struct {
+	t *testing.T
+}
+
+func (r *FakeResources) All([]ResourceType, AllOpts) ([]Resource, error) {
+	return makeNewResources(r.t), nil
```
cppforlife: any reason why not inline makeNewResources into All?

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package resources
+
+import (
+	"testing"
+
+	"github.com/cppforlife/go-cli-ui/ui"
+	logger2 "github.com/k14s/kapp/pkg/kapp/logger"
```
cppforlife: is aliasing logger package intentional?

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+package resources
```
joaopapereira: Is there any particular reason for this test to be in `resources` vs `resources_test`?

joaopapereira (commented):
Just a minor comment on the test location

PR #203, opened 2021-03-12
Adding Linters for Checking for Deadcode

danielhelfand:
This pull request adds linters to detect if any deadcode is in any changes being added to kapp. Currently, there is very little deadcode project, but might be helpful to have a way to protect against it moving forward.

### DIFF
--- a/pkg/kapp/logger/ui.go
+++ b/pkg/kapp/logger/ui.go
@@ -11,7 +11,9 @@ import (
 )
 
 const (
+	// nolint: deadcode,unused
 	loggerLevelError = "error"
+	// nolint: deadcode,unused
 	loggerLevelInfo  = "info"
 	loggerLevelDebug = "debug"
 )
--- a/pkg/kapp/logger/ui.go
+++ b/pkg/kapp/logger/ui.go
@@ -11,9 +11,7 @@ import (
 )
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/kapp/logger/ui.go
```diff
@@ -11,7 +11,9 @@ import (
 )
 
 const (
+	// nolint: deadcode,unused
```
cppforlife: i think we should use these below in Info and Error methods.

pkg/kapp/resources/association_label.go
```diff
@@ -12,7 +12,6 @@ import (
 
 const (
 	kappAssociationLabelKey = "kapp.k14s.io/association"
-	kappAssociationLabelV1  = "v1"
```
cppforlife: this one is actually looks like should be used in v1Value() of AssociationLabel. it's currently using kappIdentityAnnV1 which has a same value but technically isnt for this class.

danielhelfand: Added back and now using `kappAssociationLabelV1`

pkg/kapp/logger/ui.go
danielhelfand: Using in associated methods now.

PR #221, opened 2021-05-26
expose --kube-api-qps and --kube-api-burst flags

cppforlife:
these flags are useful to control how many requests at once are sent to the api server.


pkg/kapp/cmd/core/config_factory.go
```diff
@@ -69,6 +78,11 @@ func (f *ConfigFactoryImpl) RESTConfig() (*rest.Config, error) {
 		return nil, fmt.Errorf("Building Kubernetes config%s: %s%s", prefixMsg, err, hintMsg)
 	}
 
+	if f.qps > 0.0 {
```
pivotaljohn, line 81: So, if `qps` is not set, we ignore burst config. Is that intentional?

pivotaljohn (commented):
Two questions:

1. Given the rather opaque nature of this feature, `kapp` should acknowledge the configuration to the user somehow so _they_ have confidence the tool is using that config. How else does the user have confidence that if there _is_ a problem, _they_ aren't contributing to it?

For example, perhaps a concise parenthetical in `DepsFactoryImpl.printTarget()`?

2. When's the right moment to document a feature like this? I don't mean describing the flag (the CLI's help does that), but the purpose and use of the feature?

pkg/kapp/cmd/core/config_factory.go
cppforlife: intentional as a form of sanity check. if qps is 0.0, ConfigureClient wasnt probably called and we just dealing with zero values from the struct -- in which case we dont want to set them to 0s.

cppforlife:
> kapp should acknowledge the configuration to the user somehow so they have confidence the tool is using that config

im not sure im catching your meaning here. if user specified --kube-api-* then it's expected its getting used. if it doesnt we have a bug.

> For example, perhaps a concise parenthetical in DepsFactoryImpl.printTarget()?

this is a super low level feature, most likely used for debugging/working-with slow clusters. not something that user needs to bother with unless they are setting flags themselves and if they are then they know what they are doing.

> When's the right moment to document a feature like this? I don't mean describing the flag (the CLI's help does that), but the purpose and use of the feature?

imho probably never beyond -h output. (wouldnt want 99.9% of users using these flags)

pivotaljohn:
> > kapp should acknowledge the configuration to the user somehow so they have confidence the tool is using that config
> 
> im not sure im catching your meaning here. if user specified --kube-api-* then it's expected its getting used. if it doesnt we have a bug.

When I'm using a piece of software, it's a kind of conversation. To say, "please save my work," I click the "Save" button. I _expect_ that it will save the work. But I've had things not work enough times that there's this lingering doubt ... a doubt that pretty much goes away when the software responds, "Saved."

In this case, I'm asking `kapp` to rate limit itself. Unlike other inputs — like `-a` — where there is obvious evidence that `kapp` is respecting that input, unless it specifically reports receipt of these arguments, I've got lingering doubts. Of course, I can trust and verify (setup a proxy or network sniffer). But why not acknowledge it, proactively?

It's like doctor's employing so-called "bedside manners." They do that, in part, to build/retain trust with their patients.

> > For example, perhaps a concise parenthetical in DepsFactoryImpl.printTarget()?
> 
> this is a super low level feature, most likely used for debugging/working-with slow clusters. not something that user needs to bother with unless they are setting flags themselves and if they are then they know what they are doing.
> 
> > When's the right moment to document a feature like this? I don't mean describing the flag (the CLI's help does that), but the purpose and use of the feature?
> 
> imho probably never beyond -h output. (wouldnt want 99.9% of users using these flags)

oh. okay. That isn't obvious to me.

cppforlife:
i see where you were going with it. im not sure it makes sense to do that for such small configurations -- it's just going to be noise. i would favor not doing anything until somebody asks for that (my bet here, nobody would).

pivotaljohn:
> ... it's just going to be noise. ...

Yeah, I loosely had thought that it could be displayed if the settings were given. But that's now particularly dubious ROI.

> i would favor not doing anything until somebody asks for that (my bet here, nobody would).

Pragmatic... and if that does happen, can get the UX nuances that would lead to a useful share by teasing those details from the requestor(s). 👍🏻

pivotaljohn (approved):

PR #226, opened 2021-06-11
Automate creating github release

DennisDenuto:
- Uses goreleaser to upload and generate draft release notes with assets
attached, and preliminary release notes generated

Authored-by: Dennis Leon <leonde@vmware.com>


hack/build-binaries.sh
```diff
@@ -2,16 +2,24 @@
...
-GOOS=linux GOARCH=amd64 go build $repro_flags -o kapp-linux-amd64 ./cmd/kapp/...
-GOOS=linux GOARCH=arm64 go build $repro_flags -o kapp-linux-arm64 ./cmd/kapp/...
-GOOS=windows GOARCH=amd64 go build $repro_flags -o kapp-windows-amd64.exe ./cmd/kapp/...
+GOOS=darwin GOARCH=amd64 go build -ldflags="$LDFLAGS" -trimpath -o kapp-darwin-amd64 ./cmd/kapp/...
+GOOS=darwin GOARCH=arm64 go build -ldflags="$LDFLAGS" -trimpath -o kapp-darwin-arm64 ./cmd/kapp/...
+GOOS=linux GOARCH=amd64 go build -ldflags="$LDFLAGS" -trimpath -o kapp-linux-amd64 ./cmd/kapp/...
+GOOS=linux GOARCH=arm64 go build -ldflags="$LDFLAGS" -trimpath -o kapp-linux-arm64 ./cmd/kapp/...
+GOOS=windows GOARCH=amd64 go build -ldflags="$LDFLAGS" -trimpath -o kapp-windows-amd64.exe ./cmd/kapp/...
 
-shasum -a 256 ./kapp-*
```
100mik, line 17: We probably wanna keep this? Seems to have the same effect

pkg/kapp/version/version.go
```diff
@@ -3,6 +3,4 @@
 
 package version
 
-const (
-	Version = "0.39.0"
-)
+var Version = "develop"
```
100mik, line 6: I guess we do not wanna update this value manually, but is this change required for the automation of the release process? This variable seems to be used in a number of places and is parsed by `semver` in `pkg/kapp/config.go` as well. This will cause kapp to error out if that code path is traversed. We should ensure this change does not affect other usages?

hack/build-binaries.sh
DennisDenuto: i'm not sure if the order is guaranteed. (it might be alphabetical, in which case you would need to make sure the later steps in the GH action generate the same checksum in alphabetical order too)

The order needs to match https://github.com/vmware-tanzu/carvel-kapp/pull/226/files/80ab962ddf17464474bed5b84d286a501e14dbc6#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R79-R83

100mik: Makes sense to keep it this way and avoid the overhead!

pkg/kapp/version/version.go
DennisDenuto: > but is this change required for the automation of the release process

It is currently a manual step to update the version number before releasing a new version.

> We should ensure this change does not affect other usages

Sure, if you can identify any other usages that would break from this, definitely please list them out here. Also I think adding test coverage for those cases will help with future discoverability.

DennisDenuto:
@100mik the changes you made LGTM

PR #238, opened 2021-06-28
make diff-exit-status work for app-group deployments

### LINKED ISSUE #166: Unexpected result of diff-exit-status for app group deploy
When passing the `diff-exit-status` flag to an app group deployment, kapp exits after checking the first app inside the app group.

When applied to app groups, I would expect kapp to exit after checking all apps inside the group and returning the exit status for the complete group (unchanged, any resource has changed etc.).

twiechert:
fixes https://github.com/vmware-tanzu/carvel-kapp/issues/166 by making app-group deployments not fail if a single app returns exit status 2 or 3.

### DIFF
--- a/pkg/kapp/cmd/appgroup/deploy.go
+++ b/pkg/kapp/cmd/appgroup/deploy.go
@@ -102,7 +102,7 @@ func (o *DeployOptions) Run() error {
 	}
 
 	if o.AppFlags.DiffFlags.Run && o.AppFlags.DiffFlags.ExitStatus {
-		var hasNoChanges = exitCode == 3
+		var hasNoChanges = exitCode == 2
 		return cmdapp.DeployDiffExitStatus{HasNoChanges: hasNoChanges}
 	}
 
--- a/pkg/kapp/cmd/appgroup/deploy.go
+++ b/pkg/kapp/cmd/appgroup/deploy.go
@@ -74,10 +74,10 @@ func (o *DeployOptions) Run() error {
 		err := o.deployApp(appGroupApp)
 		if err != nil {
             if deployErr, ok := err.(cmdapp.DeployDiffExitStatus); ok {
-				exitCode = math.Max(exitCode, float64(deployErr.ExitStatus()))
-			} else {
-				return err
-			}
+                exitCode = math.Max(exitCode, float64(deployErr.ExitStatus()))
+            } else {
+                return err
+            }
 		}
 	}
 
--- a/pkg/kapp/cmd/appgroup/deploy.go
+++ b/pkg/kapp/cmd/appgroup/deploy.go
@@ -110,7 +110,7 @@ func (o *DeployOptions) Run() error {
 	}
 
 	if o.AppFlags.DiffFlags.Run && o.AppFlags.DiffFlags.ExitStatus {
-		var hasNoChanges = exitCode < 1
+		var hasNoChanges = exitCode < 3
 		return cmdapp.DeployDiffExitStatus{HasNoChanges: hasNoChanges}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -68,11 +70,22 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	var exitCode float64 = 0
 	// TODO is there some order between apps?
 	for _, appGroupApp := range updatedApps {
 		err := o.deployApp(appGroupApp)
 		if err != nil {
-			return err
+			if reflect.TypeOf(err).Name() == "DeployDiffExitStatus" {
```
cppforlife: i think we should be able to check err like so:

```
if deployErr, ok := err.(cmdapp.DeployDiffExitStatus); ok {
  ...
}
```

pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -68,11 +70,22 @@ func (o *DeployOptions) Run() error {
...
+				var deployErr cmdapp.DeployDiffExitStatus = err.(cmdapp.DeployDiffExitStatus)
+
+				if deployErr.ExitStatus() == 1 {
+					return err
+				}
```
cppforlife: it never return 1 today. instead of checking on ExitStatus, if we were able to type cast into DeployDiffExitStatus then we are working with either no changes or pending changes. no need to double check that.

pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -68,11 +70,22 @@ func (o *DeployOptions) Run() error {
...
+				exitCode = math.Max(exitCode, float64(deployErr.ExitStatus()))
+
```
cppforlife: nit: remove newline

pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -96,6 +109,11 @@ func (o *DeployOptions) Run() error {
 		}
 	}
 
+	if o.AppFlags.DiffFlags.Run && o.AppFlags.DiffFlags.ExitStatus {
+		var hasNoChanges = exitCode < 3
```
cppforlife: lets do a more explicit check here: cmdapp.DeployDiffExitStatus{HasNoChanges: exitCode == 2}

twiechert:
very valid points, thanks for the review.

pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -68,11 +68,16 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	var exitCode float64 = 0
 	// TODO is there some order between apps?
 	for _, appGroupApp := range updatedApps {
 		err := o.deployApp(appGroupApp)
 		if err != nil {
-			return err
+            if deployErr, ok := err.(cmdapp.DeployDiffExitStatus); ok {
```
cppforlife: nit: looks like formatting is a bit off

twiechert: image
I think the diff view is misleading, also the linters don't indicate anything malformed.

cppforlife: it seems that if is indented via spaces whereas the rest of the code is indented via tabs, and thats why it shows up at different invent levels in github and not in your editor.

twiechert: replaced whitestaces by tabs.

does anything speak against adding - gofmt to the linters section as part of the .golangci.yml?

Issue #241, opened 2021-06-29
kapp inspect --raw should not show managed fields by default.


GrahamDumpleton:
**Describe the problem/challenge you have**

Using ``kapp inspect --raw`` and it outputs overly verbose managed fields from resource.

**Describe the solution you'd like**

Should adopt convention ``kubectl`` has taken of not showing managed fields by default and only show them when ``--show-managed-fields=true`` is passed as an option.

See ``show-managed-fields`` option in:

* https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#get

**Anything else you would like to add:**

Nope.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

cc @jorgemoralespou

jorgemoralespou:
This was included in kubectl 1.21 as a response to the community (https://github.com/kubernetes/kubernetes/issues/90066).

PR #247, opened 2021-07-14
adding show-managed-fields flag for insepct with raw output

renuy:
This addresses #241


pkg/kapp/cmd/app/inspect.go
```diff
@@ -23,9 +23,10 @@ type InspectOptions struct {
 	ResourceFilterFlags cmdtools.ResourceFilterFlags
 	ResourceTypesFlags  ResourceTypesFlags
 
-	Raw    bool
-	Status bool
-	Tree   bool
+	Raw               bool
+	Status            bool
+	Tree              bool
+	ShowManagedFields bool
```
cppforlife: carvel tools typically do not use verb in the flag name. so in this case ManagedFields=true/false would be more consistent

pkg/kapp/cmd/app/inspect.go
```diff
@@ -48,6 +49,7 @@ func NewInspectCmd(o *InspectOptions, flagsFactory cmdcore.FlagsFactory) *cobra.
 	cmd.Flags().BoolVar(&o.Raw, "raw", false, "Output raw YAML resource content")
 	cmd.Flags().BoolVar(&o.Status, "status", false, "Output status content")
 	cmd.Flags().BoolVarP(&o.Tree, "tree", "t", false, "Tree view")
+	cmd.Flags().BoolVar(&o.ShowManagedFields, "show-managed-fields", false, "Keep the managedFields when printing objects")
```
cppforlife: lets say metadata.managedFields instead of managedFields

pkg/kapp/cmd/app/inspect.go
```diff
@@ -91,8 +93,14 @@ func (o *InspectOptions) Run() error {
 			if err != nil {
 				return err
 			}
-
-			resBs, err := historylessRes.AsYAMLBytes()
+			resWithoutManagedFields := historylessRes
+			if !o.ShowManagedFields {
+				resWithoutManagedFields, err = ctldiff.NewResourceWithHistory(historylessRes, nil, nil).ResourceWithoutManagedFields()
```
cppforlife: managed fields are not really related to concept of a history within kapp (a set of annotation). we can make a new class that represents resource with/without managed fields.

cppforlife: instead of checking on managed fields bool here, we could pass it down into such class so that it owns the behaviour whether to remove or keep managed fields.

pkg/kapp/diff/resource_with_history.go
```diff
@@ -214,3 +231,12 @@ func (resourceWithoutHistory) removeAppliedResAnnKeysMods() []ctlres.ResourceMod
 		},
 	}
 }
+
+func (resourceWithoutHistory) removeManagedFieldsResMods() []ctlres.ResourceMod {
+	return []ctlres.ResourceMod{
```
cppforlife: good use of ResourceMod. i would recommend instead of having a dedicated method just keeping this as a variable within the method that applies that mod.

also since there is only one mod there is no need to do it in the loop above.

pkg/kapp/cmd/app/inspect.go
renuy: Should the flag also be renamed to managed-fields (from show-managed-fields, tried to use the same flag that  kubectl uses.)?

pkg/kapp/cmd/app/inspect.go
```diff
@@ -48,6 +49,7 @@ func NewInspectCmd(o *InspectOptions, flagsFactory cmdcore.FlagsFactory) *cobra.
 	cmd.Flags().BoolVar(&o.Raw, "raw", false, "Output raw YAML resource content")
 	cmd.Flags().BoolVar(&o.Status, "status", false, "Output status content")
 	cmd.Flags().BoolVarP(&o.Tree, "tree", "t", false, "Tree view")
+	cmd.Flags().BoolVar(&o.ManagedFields, "show-managed-fields", false, "Keep the metadata.managedFields when printing objects")
```
cppforlife: let's also adjust show-managed-fields -> managed-fields

pkg/kapp/cmd/app/inspect.go
```diff
@@ -91,8 +93,12 @@ func (o *InspectOptions) Run() error {
 			if err != nil {
 				return err
 			}
+			resManagedFields, err := ctldiff.NewManagedFieldsResource(historylessRes, o.ManagedFields).Resource()
```
cppforlife: NewManagedFieldsResource -> NewResourceWithManagedFields reads a little better.

pkg/kapp/diff/managed_fields_resource.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
```
cppforlife: lets move this class into pkg/kapp/cmd/app package since it's only used by inspect command.

pkg/kapp/cmd/app/inspect.go
renuy: done

pkg/kapp/cmd/app/inspect.go
renuy: done

pkg/kapp/cmd/app/inspect.go
renuy: done

pkg/kapp/diff/managed_fields_resource.go
renuy: done

PR #260, opened 2021-08-03
Making file flag mandatory for deploy option (#249)

### LINKED ISSUE #249: Better warning message when kapp things you gave it empty input.

Run:

```
cat resources.yaml | kapp deploy -a website -y
```

One will get an error message saying:

```
kapp: Error: Trying to apply empty set of resources will result in deletion of resources on cluster. Refusing to continue unless --dangerous-allow-empty-list-of-resources is specified.
```

This isn't useful in understanding the obvious mistake you likely made.

Reword the message to something like:

```
kapp: Error: Trying to apply empty set of resources will result in deletion of resources on cluster. Did you forget to supply the -f/--file option? Refusing to continue unless --dangerous-allow-empty-list-of-resources is specified.
```

In other words suggest maybe the issue is they forgot to provide the ``-f`` or ``--file`` option.

In this case what was actually required was ``-f -`` and if it can detect that ``stdin`` wasn't closed but something was being piped to ``kapp`` then maybe it can be even more specific and ask if ``-f -`` was forgotten.

I get confused by this message every time when I forget the ``-f`` option. :-(

---
Vote on this request

cc @jorgemoralespou

renuy:


cppforlife:
i would actually recommend adding error check in newResourcesFromFiles (https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/cmd/app/deploy.go#L249) instead, so that we have better control over error message.

renuy:
@cppforlife , will change accordingly (checking for len(o.FileFlags.Files) < 1 to error out. 
Also do suggest the error message as we are not going with the default (required error message)
suggestions:
1.  Expected -f (file) flag to be non-empty
2. Please supply the -f (file) flag with atleast one file.
3. Expected at least one file with 'file' flag. Did you forget to add the -f flag?

Issue #261, opened 2021-08-09
Test cases do not consider EndpointSlices mirrored from Endpoints


100mik:
**What steps did you take:**
Built the application, created the needed endpoint variables and ran `./hack.test-all.sh` locally (Minikube with Kubernetes v1.21)

**What happened:**
The following tests failed : `TestTransientResourceSwitchToNonTransient` , `TestTransientResourceInspectDelete`,  `TestInspect`

**What did you expect:**
The tests should expect an EndpointSlice mirroring the Endpoint which is created with the instantiation of a Service.

**Anything else you would like to add:**
This issue can be solved by adding the expected EndpointSlice resource to the list of expected resources if the response contains one in the aforementioned tests.

Kubernetes version : v1.21
kapp version : 0.37
Operating System: MacOS 11.4

100mik:
Resolved with #265

PR #262, opened 2021-08-09
Modified tests using Service resources to expect EndpointSlices

100mik:
Modified tests to check for expected values of an EndpointSlice resource if one is present in the output.
Makes sure that tests pass in environments using newer versions of k8s.

Addresses #261


100mik:
Reopening this issue as a new one now that CLA bot is back.

PR #266, opened 2021-08-10
Add --warnings flag (#257)

### LINKED ISSUE #257: Show k8s API Deprecation Warnings

With the addition of #245, kapp by default will not show API deprecation notices as shown in this commit: https://github.com/vmware-tanzu/carvel-kapp/commit/4f554b03fa078119597f2005b43aff77c886040a

Users may want to see these deprecation notices as part of using kapp, so it might be helpful to provide an option to show deprecation notices in output. This will help users catch deprecation notices in advance sooner as described [here](https://kubernetes.io/blog/2020/09/03/warnings/).

My initial thought is to provide a command line option (e.g. `--show-deprecation-warnings`) with kapp deploy/delete commands that will show warnings like below:

```
W0719 09:22:53.305568   47369 warnings.go:67] storage.k8s.io/v1beta1 VolumeAttachment is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 VolumeAttachment
W0719 09:22:53.305624   47369 warnings.go:67] storage.k8s.io/v1beta1 CSIDriver is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 CSIDriver
W0719 09:22:53.305568   47369 warnings.go:67] networking.k8s.io/v1beta1 IngressClass is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 IngressClassList
W0719 09:22:53.305569   47369 warnings.go:67] scheduling.k8s.io/v1beta1 PriorityClass is deprecated in v1.14+, unavailable in v1.22+; use scheduling.k8s.io/v1 PriorityClass
W0719 09:22:53.306062   47369 warnings.go:67] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0719 09:22:53.306093   47369 warnings.go:67] coordination.k8s.io/v1beta1 Lease is deprecated in v1.19+, unavailable in v1.22+; use coordination.k8s.io/v1 Lease
W0719 09:22:53.309686   47369 warnings.go:67] storage.k8s.io/v1beta1 StorageClass is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 StorageClass
W0719 09:22:53.312761   47369 warnings.go:67] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0719 09:22:53.317088   47369 warnings.go:67] rbac.authorization.k8s.io/v1beta1 Role is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 Role
W0719 09:22:53.318580   47369 warnings.go:67] storage.k8s.io/v1beta1 CSINode is deprecated in v1.17+, unavailable in v1.22+; use storage.k8s.io/v1 CSINode
W0719 09:22:53.318716   47369 warnings.go:67] apiregistration.k8s.io/v1beta1 APIService is deprecated in v1.19+, unavailable in v1.22+; use apiregistration.k8s.io/v1 APIService
W0719 09:22:53.319085   47369 warnings.go:67] rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
W0719 09:22:53.333891   47369 warnings.go:67] admissionregistration.k8s.io/v1beta1 MutatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 MutatingWebhookConfiguration
W0719 09:22:53.334257   47369 warnings.go:67] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
W0719 09:22:53.401374   47369 warnings.go:67] admissionregistration.k8s.io/v1beta1 ValidatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 ValidatingWebhookConfiguration
W0719 09:22:53.403840   47369 warnings.go:67] rbac.authorization.k8s.io/v1beta1 ClusterRole is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRole
W0719 09:22:53.404188   47369 warnings.go:67] certificates.k8s.io/v1beta1 CertificateSigningRequest is deprecated in v1.19+, unavailable in v1.22+; use certificates.k8s.io/v1 CertificateSigningRequest
W0719 09:22:53.406843   47369 warnings.go:67] rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRoleBinding
```
So the final command might look like:

`kapp deploy -a app -f app.yml --show-deprecation-warnings`

It might also be nice to format the deprecation notices in a way that fits kapp's current UX.

Information on client handling: https://kubernetes.io/blog/2020/09/03/warnings/#customize-client-handling

---
Vote on this request

praveenrewar:
Add `--warning` flag with default set to `true`
Add test which uses a CRD with a custom deprecated version and warning.
Skip test if version less than 1.19 as CLI warnings were introduced in v1.19 (https://kubernetes.io/blog/2020/09/03/warnings/).

### DIFF
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -6,6 +6,7 @@ package core
 import (
 	"context"
 	"fmt"
+	"os"
 	"sync"
 
 	"github.com/cppforlife/go-cli-ui/ui"
@@ -46,9 +47,7 @@ func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
 
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
-	if !f.Warnings {
-		cpConfig.WarningHandler = rest.NoWarnings{}
-	}
+	cpConfig.WarningHandler = f.newWarningHandler()
 
 	clientset, err := dynamic.NewForConfig(cpConfig)
 	if err != nil {
@@ -117,4 +116,15 @@ func (f *DepsFactoryImpl) summarizeNodes(config *rest.Config) string {
 		}
 		return fmt.Sprintf("%s, %d+", oldestNode.Name, len(nodes.Items)-1)
 	}
+}
+
+func (f *DepsFactoryImpl) newWarningHandler() rest.WarningHandler {
+	if !f.Warnings {
+		return rest.NoWarnings{}
+	}
+	options := rest.WarningWriterOptions{
+		Deduplicate: true,
+	}
+	warningWriter := rest.NewWarningWriter(os.Stderr, options)
+	return warningWriter
 }
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -6,7 +6,6 @@ package core
 import (
 	"context"
 	"fmt"
-	"os"
 	"sync"
 
 	"github.com/cppforlife/go-cli-ui/ui"
@@ -124,7 +123,17 @@ func (f *DepsFactoryImpl) newWarningHandler() rest.WarningHandler {
 	}
 	options := rest.WarningWriterOptions{
 		Deduplicate: true,
+		Color:       true,
 	}
-	warningWriter := rest.NewWarningWriter(os.Stderr, options)
+	warningWriter := rest.NewWarningWriter(uiWriter{ui: f.ui}, options)
 	return warningWriter
+}
+
+type uiWriter struct {
+	ui ui.UI
+}
+
+func (w uiWriter) Write(data []byte) (int, error) {
+	w.ui.BeginLinef("%s", data)
+	return len(data), nil
 }
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -18,18 +18,24 @@ import (
 type DepsFactory interface {
 	DynamicClient() (dynamic.Interface, error)
 	CoreClient() (kubernetes.Interface, error)
+	ConfigureDeprecationWarning(noDeprecationWarnings bool)
 }
 
 type DepsFactoryImpl struct {
 	configFactory   ConfigFactory
 	ui              ui.UI
 	printTargetOnce *sync.Once
+
+	NoDeprecationWarnings bool
 }
 
 var _ DepsFactory = &DepsFactoryImpl{}
 
 func NewDepsFactoryImpl(configFactory ConfigFactory, ui ui.UI) *DepsFactoryImpl {
-	return &DepsFactoryImpl{configFactory, ui, &sync.Once{}}
+	return &DepsFactoryImpl{
+		configFactory:   configFactory,
+		ui:              ui,
+		printTargetOnce: &sync.Once{}}
 }
 
 func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
@@ -40,8 +46,9 @@ func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
 
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
-	// set the warning handler for this client to ignore warnings
-	cpConfig.WarningHandler = rest.NoWarnings{}
+	if f.NoDeprecationWarnings {
+		cpConfig.WarningHandler = rest.NoWarnings{}
+	}
 
 	clientset, err := dynamic.NewForConfig(cpConfig)
 	if err != nil {
@@ -67,6 +74,10 @@ func (f *DepsFactoryImpl) CoreClient() (kubernetes.Interface, error) {
 	f.printTarget(config)
 
 	return clientset, nil
+}
+
+func (f *DepsFactoryImpl) ConfigureDeprecationWarning(noDeprecationWarnings bool) {
+	f.NoDeprecationWarnings = noDeprecationWarnings
 }
 
 func (f *DepsFactoryImpl) printTarget(config *rest.Config) {
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -18,7 +18,7 @@ import (
 type DepsFactory interface {
 	DynamicClient() (dynamic.Interface, error)
 	CoreClient() (kubernetes.Interface, error)
-	ConfigureDeprecationWarning(noDeprecationWarnings bool)
+	ConfigureWarnings(warnings bool)
 }
 
 type DepsFactoryImpl struct {
@@ -26,7 +26,7 @@ type DepsFactoryImpl struct {
 	ui              ui.UI
 	printTargetOnce *sync.Once
 
-	NoDeprecationWarnings bool
+	Warnings bool
 }
 
 var _ DepsFactory = &DepsFactoryImpl{}
@@ -46,7 +46,7 @@ func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
 
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
-	if f.NoDeprecationWarnings {
+	if !f.Warnings {
 		cpConfig.WarningHandler = rest.NoWarnings{}
 	}
 
@@ -76,8 +76,8 @@ func (f *DepsFactoryImpl) CoreClient() (kubernetes.Interface, error) {
 	return clientset, nil
 }
 
-func (f *DepsFactoryImpl) ConfigureDeprecationWarning(noDeprecationWarnings bool) {
-	f.NoDeprecationWarnings = noDeprecationWarnings
+func (f *DepsFactoryImpl) ConfigureWarnings(warnings bool) {
+	f.Warnings = warnings
 }
 
 func (f *DepsFactoryImpl) printTarget(config *rest.Config) {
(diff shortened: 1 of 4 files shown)


test/e2e/warnings_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"strings"
+	"testing"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func TestWarningsFlag(t *testing.T) {
+	minorVersion, _ := getServerMinorVersion()
+	//skip the tests if k8s version is less than 1.19 as warnings were introduced in v1.19
```
praveenrewar: Comment seems unnecessary after adding the skip message
```suggestion
```

test/e2e/warnings_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"strconv"
+	"strings"
+	"testing"
+
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func TestWarningsFlag(t *testing.T) {
+	minorVersion, _ := getServerMinorVersion()
```
cppforlife: check that err is not nil?

test/e2e/warnings_test.go
```diff
@@ -0,0 +1,143 @@
...
+  cronSpec: "* * * * */5"
+  image: my-awesome-cron-image
+`
+	testcases := []struct {
+		logMessage   string
+		yaml         string
+		options      []string
+		expected     string
+		errorMessage string
+	}{
```
cppforlife: for this type of test i think its a lot less readable to do a table test than just having unrolled version.

test/e2e/warnings_test.go
praveenrewar: I thought of skipping the test if version was not found, but failing it makes more sense. I will update it accordingly.

test/e2e/warnings_test.go
praveenrewar: Makes sense. I will update it.

pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -40,8 +46,9 @@ func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
 
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
-	// set the warning handler for this client to ignore warnings
-	cpConfig.WarningHandler = rest.NoWarnings{}
+	if !f.Warnings {
```
cppforlife: within kapp there is a UI object that handles all interactions with stdout/stderr. when warnings are enabled lets redirect them to ui.PrintLinef (e.g. https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/cmd/app/deploy.go#L190)

pkg/kapp/cmd/warning_flags.go
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	cmdcore "github.com/k14s/kapp/pkg/kapp/cmd/core"
+	"github.com/spf13/cobra"
+)
+
+type WarningFlags struct {
```
cppforlife, line 11: what do you think about moving this flag into KubeAPIFlags (https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/cmd/core/kube_api_flags.go). they all seem to be related to control kapp's interaction with kube api.

test/e2e/warnings_test.go
```diff
@@ -0,0 +1,126 @@
...
+metadata:
+  name: crontabs.stable.example.com
+spec:
+  group: stable.example.com
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: false
+    deprecated: true
+    deprecationWarning: "<customWarning>"
```
cppforlife, line 49: very cool way to test deprecations! 💯

pkg/kapp/cmd/core/deps_factory.go
praveenrewar: Sounds good. I will take a look at the example.

pkg/kapp/cmd/warning_flags.go
praveenrewar: That does sound reasonable. I was thinking of using this as an extension to include more flags related to warnings in the near future(if necessary), for example deduplication, warnings as error, etc,. We can definitely use kube api flags for the new flags as well.
Let me know which one sounds better.

cppforlife: > for example deduplication

i dont think this needs a separate flag. we can just do it by default?

> warnings as error

that's an interesting idea.

cppforlife: this leads me to thinking should we actually call this --kube-api-warnings since these warnings are coming from kube api and not general warnings from kapp?

praveenrewar: > i dont think this needs a separate flag. we can just do it by default? 

Yes, even I haven't seen a separate flag for deduplication anywhere.

> that's an interesting idea.

This is also provided by kube APIs.

> this leads me to thinking should we actually call this --kube-api-warnings since these warnings are coming from kube api and not general warnings from kapp?

I am just worried that the names would be too long then, for warnings as error, it would be something like --kube-api-warnings-as-error. However if we continue with --warnings, then we should definitely show/hide warnings produced by kapp as well (if there are any).

pkg/kapp/cmd/core/deps_factory.go
praveenrewar: We would have to write our own custom warning handler if we would want to use ui.PrintLinef, which seems fine to me. The only concern is that we would also have to implement deduplication etc,. on our own (duplicating already existing code in vendor).
I will see if we can somehow use their existing handler which accepts an io.Writer to write the warnings, or we will proceed with our own custom warning handler.

pkg/kapp/cmd/warning_flags.go
praveenrewar: Based on discussion, continuing with --warning. In future kapp's own warning will be included under this.

pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -107,3 +117,14 @@ func (f *DepsFactoryImpl) summarizeNodes(config *rest.Config) string {
...
 }
+
+func (f *DepsFactoryImpl) newWarningHandler() rest.WarningHandler {
+	if !f.Warnings {
+		return rest.NoWarnings{}
+	}
+	options := rest.WarningWriterOptions{
+		Deduplicate: true,
+	}
+	warningWriter := rest.NewWarningWriter(os.Stderr, options)
```
praveenrewar: @cppforlife Redirecting the warnings to Stderr as that's what k8s does, should we redirect it to Stdout instead?

pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -107,3 +117,14 @@ func (f *DepsFactoryImpl) summarizeNodes(config *rest.Config) string {
 		return fmt.Sprintf("%s, %d+", oldestNode.Name, len(nodes.Items)-1)
 	}
 }
+
+func (f *DepsFactoryImpl) newWarningHandler() rest.WarningHandler {
+	if !f.Warnings {
+		return rest.NoWarnings{}
+	}
+	options := rest.WarningWriterOptions{
+		Deduplicate: true,
```
praveenrewar, line 125: This will print warning header in a yellowish colour.
```suggestion
		Deduplicate: true,
		Color: true,
```

pkg/kapp/cmd/core/deps_factory.go
cppforlife: should be using o.ui.PrintLinef as the destination.

praveenrewar: I had tried adding a custom warning handler so that we could use ui.PrintLinef, but didn't proceed with it because of the following 2 reasons:

- Code duplication: I almost wrote the exact same code as provided by [client-go](https://github.com/kubernetes/client-go/blob/2961e1de2c1350e8532945d8aca4a0b1082f2232/rest/warnings.go#L94), except that they used fmt.Printf() and I used ui.PrintLinef(). (I mentioned this [previously](https://github.com/vmware-tanzu/carvel-kapp/pull/266#discussion_r690256775) and I am really sorry for adding a new comment for an already existing discussion).
- New Lines: The warning messages provided by k8s have new line appended at the end and since ui.PrintLinef uses fmt.Println(), every warning had an extra line in between them (unless I removed the new lines).

However, I will add it again and then we can decide whether to move forward with it or not.

praveenrewar: We can also provide an intermediate buffer to client go warning writer and then provide the stringified version to ui.PrintLinef, althought I am not convinced with this approach as either.
Let me know your suggestions as I might have missed on some better alternatives :)

cppforlife: NewWarningWriter takes an io.Writer. so we can add a little adapter

```
type uiWriter struct { ui UI }
func (w uiWriter) Write(data []byte) (int, error) { w.ui.BeginLinef("%s", data); return len(data), nil }
```

praveenrewar: Done. And updated the test to check Stdout instead of Stderr.

Issue #208, opened 2021-03-29
Add per change wait timeout


ewrenn8:
**Describe the problem/challenge you have**
I want to lower the amount of time I wait for a change to take effect, but still allow longer running deploys to continue successfully.

**Describe the solution you'd like**
Currently, lowering the timeout lowers the amount of time allowed for the entire deploy to finish, regardless of if it is continuing successfully or not. I would like a way to specify a per change wait period, so that if any single change takes longer than Xs it will exit with an error, but if changes are continuing successfully, kapp will not error.

**Anything else you would like to add:**

---

Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this bug fixed as soon as possible"
👎 "There are more important bugs to focus on right now"

We are also happy to receive and review Pull Requests if you want to help fix this issue.

aaronshurley:
Hey @ewrenn8, could you share a bit more about a specific use case that you'd like to improve?

ewrenn8:
This is to improve kapp's ability to determine when a deployment has failed. Currently, the timeout toggle isn't granular enough to distinguish between a failing deployment and a successful deployment that takes a long time. By introducing a per wait timeout, users will be able to leave the overall timeout long enough to accommodate long running deployments without having to wait a long amount of time before kapp decides that deployment has failed.

I can think of a deployment of something like cf-for-k8s as an example of where this could be useful. This deploy usually takes a couple minutes to complete, but if I don't expect any change to take longer than 30s, I could get an error from kapp in 30s instead of 8 min.

aaronshurley:
Discussed at our community meeting on 4/5. For this issue:
- We're interested in keeping the timeout as a general timeout per change and not something more granular (i.e. specific change).
- We don't intend to change the global timeout as a part of this issue.
- Add this configuration as a command-line flag (similar to the existing global timeout).
- We didn't come to a conclusion on a whether this timeout should be for (a) deploy only or (b) for deploy and delete.

joaopapereira:
Given the above outcome, going to change the label to accept.
This feature includes the addition of the flag `--apply-resource-timeout`.

---

**Acceptance criteria**

🟢 Happy path
**Given** I have some manifests to install
**When** I execute `kapp deploy -a app1 -f manifests --apply-resource-timeout 1m`
**Then** I should see `Succeed`

🔴 Changes take longer than the timeout
[...]

PR #267, opened 2021-08-10
[Draft] #208 per change wait timeout

sethiyash:


sethiyash:
Have raised a new one.

Issue #269, opened 2021-08-12
Increase Timeout of golangci-lint 


sethiyash:
**Describe the problem/challenge you have** - 
golangci-lint check fails sometimes on new PR's throwing timeout error

**Describe the solution you'd like**
Increasing the timeout to maybe 5 mins as we already have in kapp-controller 

**Anything else you would like to add:**

Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this bug fixed as soon as possible"
👎 "There are more important bugs to focus on right now"

We are also happy to receive and review Pull Requests if you want to help fix this issue.

cppforlife:
PR was merged.

PR #270, opened 2021-08-12
#269 increasing timeout in golangci-lint

sethiyash:


.github/workflows/golangci-lint.yml
```diff
@@ -17,4 +17,4 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         with:
           version: v1.34
-          args: -v
+          args: -v --timeout 2m
```
danielhelfand: You may want to bump this higher. We have 5m for kapp-controller. The golangci-lint action can be flakey some times, so probably best to be conservative.

danielhelfand (commented):
Also, this should probably be added to `.golangci.yml` config instead of action. Like shown here:

https://github.com/vmware-tanzu/carvel-kapp-controller/blob/adae427d34fcfd20b56721ad09e03299ab56758f/.golangci.yml#L2

.github/workflows/golangci-lint.yml
sethiyash: Sure. Thanks

sethiyash:
> Also, this should probably be added to `.golangci.yml` config instead of action. Like shown here:
> 
> https://github.com/vmware-tanzu/carvel-kapp-controller/blob/adae427d34fcfd20b56721ad09e03299ab56758f/.golangci.yml#L2

Ah! Got it. Thanks

PR #274, opened 2021-08-16
GitHub Workflow altered to use latest version of Minikube

### LINKED ISSUE #268: GitHub workflow test-gh.yml does not run tests against latest version of k8s
The script seems to be fetching an older version of Minikube (v1.10.0) to run tests against :

```
wget -O- https://github.com/kubernetes/minikube/releases/download/v1.10.0/minikube-linux-amd64 > /tmp/bin/minikube
chmod +x /tmp/bin/minikube
```

This version does not use the latest version of K8s by default.
We should change the script to get the latest stable version of Minikube.

Tests that were failing on K8s v1.21 (#261 ) should have ideally failed in the pipeline.

### LINKED ISSUE #298: Make result of tree inspect deterministic
Tree inspect has indeterministic output due to UIDs being used to track parent resources.

This is an issue tracking ongoing work on [this PR](https://github.com/vmware-tanzu/carvel-kapp/pull/274).

100mik:
Fetches latest binary as instructed [here](https://minikube.sigs.k8s.io/docs/start/)

Addresses #268

### DIFF
--- a/pkg/kapp/cmd/tools/inspect_tree_view.go
+++ b/pkg/kapp/cmd/tools/inspect_tree_view.go
@@ -189,7 +189,7 @@ func (a *assocSortingValue) uidOwnersStr() string {
 }
 
 func (a *assocSortingValue) resIdentifier(resource ctlres.Resource) string {
-	return fmt.Sprintf("%s-%s", resource.Kind(), resource.Name())
+	return fmt.Sprintf("%s$%s$%s$%s", resource.Namespace(), resource.APIGroup(), resource.Kind(), resource.Name())
 }
 
 func (a *assocSortingValue) Depth() int {
--- a/pkg/kapp/cmd/tools/inspect_tree_view.go
+++ b/pkg/kapp/cmd/tools/inspect_tree_view.go
@@ -167,7 +167,7 @@ func (a *assocSortingValue) labelAssocStr() string {
 }
 
 func (a *assocSortingValue) uidOwnersStr() string {
-	uids := []string{a.resource.UID()}
+	identifiers := []string{a.resIdentifier(a.resource)}
 	nextRes := &a.resource
 
 	for nextRes != nil {
@@ -178,14 +178,18 @@ func (a *assocSortingValue) uidOwnersStr() string {
 			foundRes, found := a.rsByUID[string(ref.UID)]
 			if found {
 				// only nest into first object that we find
-				uids = append([]string{string(ref.UID)}, uids...)
+				identifiers = append([]string{a.resIdentifier(foundRes)}, identifiers...)
 				nextRes = &foundRes
 				break
 			}
 		}
 	}
 
-	return "ref-" + strings.Join(uids, "/")
+	return "ref-" + strings.Join(identifiers, "/")
+}
+
+func (a *assocSortingValue) resIdentifier(resource ctlres.Resource) string {
+	return fmt.Sprintf("%s-%s", resource.Kind(), resource.Name())
 }
 
 func (a *assocSortingValue) Depth() int {
--- a/test/e2e/inspect_test.go
+++ b/test/e2e/inspect_test.go
@@ -126,9 +126,6 @@ spec:
 				"reconcile_info":  "",
 				"reconcile_state": "ok",
 			})
-
-			sortByKind(expected)
-			sortByKind(respRows)
 		}
 
 		if !reflect.DeepEqual(replaceAge(respRows), expected) {
--- a/test/e2e/inspect_test.go
+++ b/test/e2e/inspect_test.go
@@ -126,6 +126,9 @@ spec:
 				"reconcile_info":  "",
 				"reconcile_state": "ok",
 			})
+
+			sortByKind(expected)
+			sortByKind(respRows)
 		}
 
 		if !reflect.DeepEqual(replaceAge(respRows), expected) {
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -31,8 +31,8 @@ jobs:
 
         wget -O- https://carvel.dev/install.sh | K14SIO_INSTALL_BIN_DIR=/tmp/bin bash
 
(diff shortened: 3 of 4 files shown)


test/e2e/inspect_test.go
```diff
@@ -126,6 +126,9 @@ spec:
 				"reconcile_info":  "",
 				"reconcile_state": "ok",
 			})
+
+			sortByKind(expected)
+			sortByKind(respRows)
```
cppforlife: what prompted this change? is the diff summary not deterministic?

100mik: While inspecting an app with the `--tree` flag, the result seemed to be deterministic and the order was generally, `Service, EndpointSlice, Endpoints` . The test initially had a different order but was passing erratically. Ordering seemed to be a viable option as we are checking for the existence of the resources and the test results were reproducible once the values were sorted in a deterministic manner. 
I was mainly concerned about the case where the output did have a different order and passed the test.

100mik:
@praveenrewar The test seems to be failing due to warnings on latest version of K8s after I rebased on develop. Should we set the warnings flag for these tests?

praveenrewar:
> @praveenrewar The test seems to be failing due to warnings on latest version of K8s after I rebased on develop. Should we set the warnings flag for these tests?

The tests could be failing because we have now redirected the warnings to Stdout instead of Stderr. 
@rohitagg2020  @cppforlife Should we redirect them to Stderr or should we update the test to use the warning flag?

pkg/kapp/cmd/tools/inspect_tree_view.go
```diff
@@ -133,7 +134,7 @@ func newAssocSortingValue(resource ctlres.Resource, rsByUID map[string]ctlres.Re
 }
 
 func (a *assocSortingValue) Value() string {
-	return strings.Join(a.Values(), "@")
+	return a.Values()[0] + "-" + fmt.Sprintf("%d", a.Depth())
```
praveenrewar: Should we check the length of the array before using an index?

test/e2e/inspect_test.go
```diff
@@ -125,7 +125,8 @@ spec:
 				"owner":           "cluster",
 				"reconcile_info":  "",
 				"reconcile_state": "ok",
-			})
+			}
+			expected = append(append([]map[string]string{expected[0]}, expectedEndpointSlice), expected[1])
```
praveenrewar: Can we make it more readable somehow?
I can think of 2 ways: 
One is to have different values for Service, Endpoint and EndpointSlice and append them based on hasEndpointSlice.
Or we can use copy and do something like [this](https://play.golang.org/p/THJRkJtahsx).

pkg/kapp/cmd/tools/inspect_tree_view.go
100mik: It does make sense to. I think this might be an issue if a resource that should be displayed does not have a `kapp.k14s.io/association` label. Looking into this and a better approach as this one seems to miss the case where there are multiple resources at the same depth.

test/e2e/inspect_test.go
100mik: I actually did what you've done in the playground initially as well. I am not sure which one would be more suitable for legibility. I guess we can revisit this after the sorting issue is resolved as the order might change and this might be unnecessary.

praveenrewar: Basically, I was thinking that the test should be clear that we are expecting the array elements in certain order and that order should be clear. Even I am not satisfied with either of the approaches. But yeah, we can visit this later.

100mik: Yes that makes complete sense

pkg/kapp/cmd/tools/inspect_tree_view.go
```diff
@@ -178,14 +178,18 @@ func (a *assocSortingValue) uidOwnersStr() string {
...
 			}
 		}
 	}
 
-	return "ref-" + strings.Join(uids, "/")
+	return "ref-" + strings.Join(identifiers, "/")
+}
+
+func (a *assocSortingValue) resIdentifier(resource ctlres.Resource) string {
+	return fmt.Sprintf("%s-%s", resource.Kind(), resource.Name())
```
100mik: Replacing UIDs with a `<Kind>-<Name>` value, this results in deterministic sorting. In the examples I considered, the resources with same name and kind in different namespaces had a different `kapp.k14s.io/association` label as well and the got grouped separately. We could certainly consider namespaces too.

```suggestion
	return fmt.Sprintf("%s-%s-%s", resource.Namespace(), resource.Name(), resource.Kind())
```

cppforlife: afaik it is currently using owner uid, which means that it's able to place related objects (aka owned) one after another (hence it forms a tree within a table). if we turn into ns/name/kind then it's not going to place related objects one after another and it's not going to be a tree. is this right, or am i missing something?

100mik: We are calling this function for each owner resource. The function `uidOwnersStr()` returned a path of sorts using UIDs. Arranging one resource after the other.Replaced those UIDs with ns/name/kind  for each of the owners.
Instead of `UID1/UID2/UID3` it is now `ns-kind1-name1/ns-kind2-name2/ns-kind3-name3` (In case, this suggestion is accepted)

For example for a pod the entire group header value looks like:
`lbl-v1.0e236d647aa83adeae8c4c5bf1556f4d-2/child@ref-default-nginx-deployment-2-Deployment/default-nginx-deployment-2-cbdd67c76-ReplicaSet/default-nginx-deployment-2-cbdd67c76-k6pf4-Pod`

Maybe, we should consider stricter test cases for tree view?

100mik:
```suggestion
	return fmt.Sprintf("%s+%s+%s+%s", resource.Namespace(), resource.APIGroup(), resource.Kind(), resource.Name())
```
Ensures that the identifier is unique and avoids edge cases where the '-' separator might be confused for a dash used in a name.

100mik:
@cppforlife rebasing to ensure all new changes on develop pass on latest version of k8s

PR #275, opened 2021-08-16
#208 resource wait timeout

sethiyash:


cppforlife:
@sethiyash are we not able to extend waiting_changes.go (https://github.com/vmware-tanzu/carvel-kapp/blob/e0bf1c9d485c5c59a23b4cfe75915782e7fb4b42/pkg/kapp/clusterapply/waiting_changes.go) to include this feature? current change set seems to touch quite a bit of code which to me is an indicator that there might be easier ways.

PR #279, opened 2021-08-17
#208 resource wait timeout

### LINKED ISSUE #208: Add per change wait timeout

sethiyash:


pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -13,9 +14,10 @@ import (
 )
 
 type WaitingChangesOpts struct {
-	Timeout       time.Duration
-	CheckInterval time.Duration
-	Concurrency   int
+	Timeout             time.Duration
+	ResourceWaitTimeout time.Duration
```
cppforlife: ResourceWaitTimeout -> ResourceTimeout (we are already under waiting struct and the other field is just called Timeout)

pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -62,13 +65,19 @@ func (c *WaitingChanges) WaitForAny() ([]WaitingChange, error) {
 		for _, change := range c.trackedChanges {
 			change := change // copy
 
-			go func() {
+			go func(timeout time.Duration) {
```
cppforlife: any specific reason why we are taking this in like this?

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,258 @@
+package e2e
```
cppforlife, line 4: see golint error below

pkg/kapp/cmd/app/apply_flags.go
```diff
@@ -55,6 +55,8 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 
 	cmd.Flags().DurationVar(&s.WaitingChangesOpts.Timeout, prefix+"wait-timeout",
 		mustParseDuration("15m"), "Maximum amount of time to wait in wait phase")
+	cmd.Flags().DurationVar(&s.WaitingChangesOpts.ResourceWaitTimeout, prefix+"resource-wait-timeout",
```
cppforlife: flag should be --wait-resource-timeout (all wait flags start with a wait prefix)

pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -81,6 +90,10 @@ func (c *WaitingChanges) WaitForAny() ([]WaitingChange, error) {
 			desc := fmt.Sprintf("waiting on %s", change.Cluster.WaitDescription())
 			c.ui.Notify(descMsgs)
 
+			if err != nil && err.Error() == "resource wait timeout" {
```
cppforlife: below is a check for err != nil. why dont we just use that instead of making a special case for this?

pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -62,13 +65,19 @@ func (c *WaitingChanges) WaitForAny() ([]WaitingChange, error) {
...
+				//To start tracking the time for the new resource
+				if change.startTime.IsZero() {
+					change.startTime = time.Now()
+				}
+
 				waitThrottle.Take()
 				defer waitThrottle.Done()
 
 				state, descMsgs, err := change.Cluster.IsDoneApplying()
+				err = checkForTimeout(change, timeout, err)
```
cppforlife: lets inline logic for checkForTimeout here.

pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -126,3 +139,12 @@ func (c *WaitingChanges) Complete() error {
 func (c *WaitingChanges) stats() string {
 	return fmt.Sprintf("[%d/%d done]", c.numWaited, c.numTotal)
 }
+
+func checkForTimeout(change WaitingChange, timeout time.Duration, err error) error {
+	if time.Now().Sub(change.startTime) > timeout {
+		return errors.New("resource wait timeout")
+	} else if err != nil {
+		return err
+	}
```
cppforlife: we should prefer "actual" error over timeout error when we have it.

pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -126,3 +139,12 @@ func (c *WaitingChanges) Complete() error {
 func (c *WaitingChanges) stats() string {
 	return fmt.Sprintf("[%d/%d done]", c.numWaited, c.numTotal)
 }
+
+func checkForTimeout(change WaitingChange, timeout time.Duration, err error) error {
+	if time.Now().Sub(change.startTime) > timeout {
+		return errors.New("resource wait timeout")
```
cppforlife: this is where should build full error message instead of doing it later (e.g. fmt.Errorf("Resource timed out waiting after %s", c.opts.ResourceWaitTimeout))

pkg/kapp/clusterapply/applying_changes.go
```diff
@@ -95,7 +95,7 @@ func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange
 			}
 
 			c.markApplied(result.Change)
-			appliedChanges = append(appliedChanges, WaitingChange{result.Change, result.ClusterChange})
+			appliedChanges = append(appliedChanges, WaitingChange{result.Change, result.ClusterChange, time.Time{}})
```
cppforlife: we should just initiate this time field with time.Now() here instead of doing it later.

sethiyash: Yeah this will be. better. Done

pkg/kapp/clusterapply/waiting_changes.go
sethiyash: Done

pkg/kapp/clusterapply/waiting_changes.go
sethiyash: Removed this function and making sure actual error will get the preference always over timeout

pkg/kapp/clusterapply/waiting_changes.go
sethiyash: Done

pkg/kapp/clusterapply/waiting_changes.go
sethiyash: Done. Changed the logic accordingly

pkg/kapp/cmd/app/apply_flags.go
sethiyash: Ah! got it. Done.

pkg/kapp/clusterapply/waiting_changes.go
sethiyash: Yeah we can remove this better to use it everywhere using `c.opts.ResourceTimeout`

pkg/kapp/clusterapply/waiting_changes.go
sethiyash: Done

test/e2e/wait_timeout_test.go
sethiyash: Got it. Done

pkg/kapp/cmd/app/apply_flags.go
```diff
@@ -55,6 +55,8 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 
 	cmd.Flags().DurationVar(&s.WaitingChangesOpts.Timeout, prefix+"wait-timeout",
 		mustParseDuration("15m"), "Maximum amount of time to wait in wait phase")
+	cmd.Flags().DurationVar(&s.WaitingChangesOpts.ResourceTimeout, prefix+"wait-resource-timeout",
+		mustParseDuration("15m"), "Maximum amount of time to wait for a resource in wait phase")
```
cppforlife: i was thinking a bit more about this. i think we should set this to default to 0s (also add a note to flag description, 0s means no timeout). check in WaitingChanges if duration is 0 and skip timing out if it is. reasoning behind this is that we currently dont have per resource timeout, and i dont think we want to add a timeout since it may affect some existing deployments.

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("Resource wait timeout", func() {
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--wait-timeout",
+			"2m", "--wait-resource-timeout", "2ms", "--json"},
+			RunOpts{IntoNs: true, AllowError: true, StdinReader: strings.NewReader(yaml1)})
+
+		expected := []map[string]string{{
```
cppforlife: i dont think we need to check out diff in this test (diff isnt really related to this feature). your assertion below about the error is good.

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,261 @@
...
+	name := "test-wait-timeout"
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("Resource wait timeout", func() {
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--wait-timeout",
```
cppforlife: should have _, err := ... and check that err is not nil.

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,261 @@
...
+			"reconcile_state": "",
+			"wait_to":         "reconcile",
+		}}
+
+		resp := uitest.JSONUIFromBytes(t, []byte(out))
+
+		validateChanges(t, resp.Tables, expected, "Op:      6 create, 0 delete, 0 update, 0 noop",
+			"Wait to: 6 reconcile, 0 delete, 0 noop", out)
+
+		if len(resp.Lines) > 0 && !strings.Contains(resp.Lines[len(resp.Lines)-1] ,"Resource timed out waiting after 2ms") {
```
cppforlife: instead of looking at lines lets check err.Error() contains that string.

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,261 @@
...
+	})
+
+	cleanUp()
+
+	logger.Section("Global wait timeout", func() {
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--wait-timeout",
+			"2ms", "--wait-resource-timeout", "2m", "--json"},
+			RunOpts{IntoNs: true, AllowError: true, StdinReader: strings.NewReader(yaml1)})
+
+		expected := []map[string]string{{
```
cppforlife: same for this test as well.

pkg/kapp/cmd/app/apply_flags.go
sethiyash: Yeah sure sounds reasonable. Done

test/e2e/wait_timeout_test.go
sethiyash: Yeah sure. Done

test/e2e/wait_timeout_test.go
sethiyash: Done

test/e2e/wait_timeout_test.go
sethiyash: Yeah Done

test/e2e/wait_timeout_test.go
sethiyash: Done

pkg/kapp/cmd/app/apply_flags.go
```diff
@@ -55,6 +55,8 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 
 	cmd.Flags().DurationVar(&s.WaitingChangesOpts.Timeout, prefix+"wait-timeout",
 		mustParseDuration("15m"), "Maximum amount of time to wait in wait phase")
+	cmd.Flags().DurationVar(&s.WaitingChangesOpts.ResourceTimeout, prefix+"wait-resource-timeout",
+		mustParseDuration("0"), "Maximum amount of time to wait for a resource in wait phase. 0s is the default value which indicates no timeout")
```
cppforlife: `Maximum amount of time to wait for a resource in wait phase. 0s is the default value which indicates no timeout` -> `Maximum amount of time to wait for a resource in wait phase (0s means no timeout)`

pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -67,6 +72,10 @@ func (c *WaitingChanges) WaitForAny() ([]WaitingChange, error) {
 				defer waitThrottle.Done()
 
 				state, descMsgs, err := change.Cluster.IsDoneApplying()
+				// check for resource timeout
+				if err == nil && c.opts.ResourceTimeout != DefaultResourceTimeout && time.Now().Sub(change.startTime) > c.opts.ResourceTimeout {
+					err = fmt.Errorf("Resource timed out waiting after %s", c.opts.ResourceTimeout)
+				}
```
cppforlife: i would recommend splitting it across multiple conditionals for better readability:

```
if err == nil {
	if c.opts.ResourceTimeout != 0 && time.Now().Sub(change.startTime) > c.opts.ResourceTimeout {
		err = fmt.Errorf("Resource timed out waiting after %s", c.opts.ResourceTimeout)
	}
}
```

i would also recommend inlining DefaultResourceTimeout since it's not providing more clarity of what it is via its name.

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,125 @@
...
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ns2
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: ns3
```
cppforlife: is there a reason why we need namespaces? can we just be using default default test namespace that kapp.Deploy automatically sets.

sethiyash: No I was just creating that much namespaces such that Deployments will wait for the namespaces to get created and the changes will not get finished instantly and I can provide resource time value and make it fail. 

Yeah we can just use default namespace

pkg/kapp/clusterapply/waiting_changes.go
sethiyash: Sure. Done

pkg/kapp/cmd/app/apply_flags.go
sethiyash: @cppforlife can be have default value as `0` instead of `0s` because time parser is treating differently if we give `0s` and the condition I have put in `pkg/kapp/clusterapply/waiting_changes.go` line 77 will work only for `0` not for `0s` if we are planning to have it as `0s` then we need to change the condition.

cppforlife: how is it different? https://play.golang.org/p/5QqqYr9Iy6D

sethiyash: oh okay got it my bad.

cppforlife: np. was just trying to see where you were stuck.

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"testing"
+)
+
+func TestWaitTimeout(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+
+	yaml1 := `
+apiVersion: apps/v1
```
cppforlife: one last thing to consider here is how flakey deployments would be. eg first time it might take some and tests would succeed, but subsequent times it may be quite fast because image is already cached. so how do we make it more deterministic? we could for example use a resource where we explicitly control how long it takes to reconcile e.g. Job with a sleep (https://github.com/vmware-tanzu/carvel-kapp/blob/e0bf1c9d485c5c59a23b4cfe75915782e7fb4b42/test/e2e/create_fallback_on_update_test.go#L26-L43)

rohitagg2020: can we use imagePullPolicy flag of k8s and set its value to "Always"?. This will ensure that we never pick the cached image and it will always download from the hub/repositories

My bad. This might not work always as image wont be fetched if there is no digest change.[https://kubernetes.io/docs/concepts/configuration/overview/#container-images](https://kubernetes.io/docs/concepts/configuration/overview/#container-images)

cppforlife: > command: ["/bin/sh", "-c", "sleep 10"] 

for jobs, this part is what making things deterministic, not image pulling.

sethiyash: Sounds reasonable @cppforlife. I have made the changes

test/e2e/wait_timeout_test.go
```diff
@@ -0,0 +1,86 @@
...
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("Resource timed out waiting", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--wait-timeout",
+			"2m", "--wait-resource-timeout", "3s", "--json"},
```
cppforlife: how come we are using different timeout for different purposes? lets standardize...

1s - for failing timeout
100s - for cases when resource succeeds before timeout
10000s - for cases when specific timeout should not play a role

sethiyash: Sure done

PR #280, opened 2021-08-19
[Draft] Adding resource filter for labels (#272)

### LINKED ISSUE #272: support filtering resources by label in kapp deploy
**What steps did you take:**
```bash
:; kapp deploy -a 'label:!undefined' -f path/to/yaml
```

**What happened:**
```
kapp: Error: Expected label selector to check for equality
```

**What did you expect:**
I expected kapp to deploy all the resources in the yaml since none have a label w/ the key "undefined", and the expression in the attempted command line is supposed to be valid.

**Environment:**

- kapp version (use `kapp --version`): 0.37.0

---
Vote on this request

renuy:
@rohitagg2020


pkg/kapp/resources/resource_filter.go
```diff
@@ -92,6 +94,26 @@ func (f ResourceFilter) Matches(resource Resource) bool {
...
+			expectedLabelKey, expectedLabelVal, err := NewSimpleLabel(labelSelector).KV()
+			if err != nil {
+				return false
+			}
+			if val, found := resource.Labels()[expectedLabelKey]; found {
+				if val == expectedLabelVal {
+					matched = true
+					break
+				}
+			}
```
cppforlife: instead of doing manual matching, lets use https://github.com/vmware-tanzu/carvel-kapp/blob/e0bf1c9d485c5c59a23b4cfe75915782e7fb4b42/vendor/k8s.io/apimachinery/pkg/labels/selector.go#L38. this gives us ability to work with label selectors like: `!key`, `key=foo`, `key!=foo` etc.

pkg/kapp/resources/resource_filter.go
```diff
@@ -5,6 +5,7 @@ package resources
 
 import (
 	"encoding/json"
+	"k8s.io/apimachinery/pkg/labels"
```
cppforlife: nit: third party exports go into the next section

pkg/kapp/resources/resource_filter.go
```diff
@@ -92,6 +94,26 @@ func (f ResourceFilter) Matches(resource Resource) bool {
 		}
 	}
 
+	if len(f.Labels) > 0 {
+		var matched bool
```
praveenrewar, line 99: Should we introduce a separate method for this flag like BoolFilter? Maybe we can do this later and do it for a few other flags as well.

praveenrewar: However, the code is readable, so we can definitely skip it.

pkg/kapp/resources/resource_filter.go
renuy: done. Please check

pkg/kapp/resources/resource_filter.go
renuy: done

rohitagg2020 (approved):

pkg/kapp/resources/resource_filter.go
```diff
@@ -92,6 +94,24 @@ func (f ResourceFilter) Matches(resource Resource) bool {
 		}
 	}
 
+	if len(f.Labels) > 0 {
+		var matched bool = false
```
cppforlife: no need to set `= false`

pkg/kapp/resources/resource_filter.go
```diff
@@ -92,6 +94,24 @@ func (f ResourceFilter) Matches(resource Resource) bool {
...
+		for _, label := range f.Labels {
+			labelSelector, err := labels.Parse(label)
+			if err != nil {
+				return false
```
cppforlife: i wonder if we should panic here for now, since otherwise we are swallowing error and skipping other matchers. eventually we could add Validate() may be to ResourceFilter.

pkg/kapp/resources/resource_filter.go
```diff
@@ -92,6 +94,24 @@ func (f ResourceFilter) Matches(resource Resource) bool {
...
+			}
+			labelsSet := labels.Set(resource.Labels())
+			if labelSelector.Matches(labelsSet) {
```
cppforlife: lets inline labelsSet since it's just a cast. eg

```
if labelSelector.Matches(labels.Set(resource.Labels())) {
```

test/e2e/label_filter_test.go
```diff
@@ -0,0 +1,90 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestLabelFilter(t *testing.T) {
```
cppforlife: for this functionality having an e2e test is quite heavy. i would recommend adding a unit test against ResourceFilter class directly.

if this is the only e2e test that touches filtering i would modify it a bit to be more generic, eg
- test 2 filter flags in combination
- test --filter

also we do not need to actually deploy things, just have to check against diff summary since that's what filter applies to.

test/e2e/label_filter_test.go
```diff
@@ -0,0 +1,90 @@
...
+	})
+
+	logger.Section("deploy initial", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--filter-labels", "x!=y"}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+
+		NewMissingClusterResource(t, "service", "redis-primary", env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", "redis-config", env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", "redis-config2", env.Namespace, kubectl)
+	})
+
```
cppforlife: nit: extra newline

100mik (approved):

pkg/kapp/resources/resource_filter.go
renuy: will panic.

pkg/kapp/resources/resource_filter.go
renuy: ok

test/e2e/label_filter_test.go
renuy: if out of the two filter flags, one is "--filter" then it will always take precedence. other filter flag (like labels or kind or nz etc) will not get applied. Hence adding test with filter-label and filter-kind together.
Found testing with table output better than testing with json output.

sethiyash (approved):

test/e2e/filter_test.go
```diff
@@ -0,0 +1,127 @@
...
+			RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+
+		if !strings.Contains(out, expectedOutput2) {
+			t.Fatalf("Did not find expected diff output >>%s<< in >>%s<<", out, expectedOutput2)
+		}
+	})
+
+	logger.Section("test filter flag", func() {
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--diff-run",
+			"--filter", "{\"resource\":{\"kinds\":[\"Service\"]}}"},
```
cppforlife: tip: this is a good case to use ` for wrapping string so that you dont have escape " with slashes.

test/e2e/filter_test.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"strings"
+	"testing"
+)
+
+func TestFilter(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	expectedOutput1 := `
```
cppforlife: good pattern to follow here is to put expected "material" as close as possible to the assertion itself.

renuy: Done

test/e2e/filter_test.go
renuy: done

PR #289, opened 2021-08-24
Set default value of warnings flag to false

praveenrewar:
This is a temporary solution, we should revert it once we find a concrete solution to the deprecation warnings being produced due to the list api call.

### DIFF
--- a/pkg/kapp/cmd/warning_flags.go
+++ b/pkg/kapp/cmd/warning_flags.go
@@ -13,7 +13,7 @@ type WarningFlags struct {
 }
 
 func (f *WarningFlags) Set(cmd *cobra.Command, flagsFactory cmdcore.FlagsFactory) {
-	cmd.PersistentFlags().BoolVar(&f.Warnings, "warnings", true, "Show warnings")
+	cmd.PersistentFlags().BoolVar(&f.Warnings, "warnings", false, "Show warnings")
 }
 
 func (f *WarningFlags) Configure(depsFactory cmdcore.DepsFactory) {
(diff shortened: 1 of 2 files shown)


pkg/kapp/cmd/warning_flags.go
```diff
@@ -13,7 +13,7 @@ type WarningFlags struct {
 }
 
 func (f *WarningFlags) Set(cmd *cobra.Command, flagsFactory cmdcore.FlagsFactory) {
-	cmd.PersistentFlags().BoolVar(&f.Warnings, "warnings", true, "Show warnings")
+	cmd.PersistentFlags().BoolVar(&f.Warnings, "warnings", false, "Show warnings")
```
praveenrewar, line 16: Should we change the message?

test/e2e/warnings_test.go
```diff
@@ -79,7 +79,7 @@ spec:
 	})
 	logger.Section("deploying without --warnings flag", func() {
```
praveenrewar:
```suggestion
	logger.Section("deploying with --warnings flag", func() {
```

test/e2e/warnings_test.go
```diff
@@ -88,7 +88,7 @@ spec:
 	})
 	logger.Section("deploying with --warnings flag", func() {
```
praveenrewar:
```suggestion
	logger.Section("deploying without --warnings flag", func() {
```

pkg/kapp/cmd/warning_flags.go
cppforlife: i think its fine.

PR #296, opened 2021-08-27
Checking the UID to confirm the deletion of resource

### LINKED ISSUE #229: kapp delete change should precondition on uid when waiting for resource to be gone

currently kapp when waiting for resource to be fully deleted off cluster, ignores that resource might have changed its uid. when uid changes it implies that resource has been deleted, and replaced by a different resource with a same name.

we should _probably_ check on uid difference and see that if resource uid changed, then consider it as deleted. there is a nuance in this change though, we should probably continue waiting for the resource to be gone if uid changed, but resource still matches criteria for being within the app (currently happens today). 

this came up during following steps:
- with kapp-controller 0.20.0-rc.1
- add pkg repo to a default ns
- add pkg repo to global packaging ns
- delete pkg repo from a default ns (this hangs for default wait timeout) because deleted package was replaced by package that is derived from global ns

---
Vote on this request

rohitagg2020:


pkg/kapp/resources/resources.go
```diff
@@ -381,7 +381,7 @@ func (c *ResourcesImpl) Exists(resource Resource) (bool, error) {
 	var found bool
 
 	err = util.Retry(time.Second, time.Minute, func() (bool, error) {
-		_, err = resClient.Get(context.TODO(), resource.Name(), metav1.GetOptions{})
+		item, err := resClient.Get(context.TODO(), resource.Name(), metav1.GetOptions{})
```
cppforlife: nit naming: item -> fetchedRes

pkg/kapp/resources/resources.go
```diff
@@ -381,7 +381,7 @@ func (c *ResourcesImpl) Exists(resource Resource) (bool, error) {
 	var found bool
```
cppforlife, line 385: instead of modifying existing behavious of Exists() (since it might be used in different places that do not care about UID), i was thinking should we add ExistsOpts{SameUID: true} optional arg?

rohitagg2020:
Should we have constants file so that if same value is used at two places, then we pick it from constants file. It will help in future enhancement as we can change the value in the constants file only.

pkg/kapp/resources/resources.go
rohitagg2020: Done

pkg/kapp/resources/resources.go
```diff
@@ -362,7 +362,7 @@ func (c *ResourcesImpl) Get(resource Resource) (Resource, error) {
 	return NewResourceUnstructured(*item, resType), nil
 }
 
-func (c *ResourcesImpl) Exists(resource Resource) (bool, error) {
+func (c *ResourcesImpl) Exists(resource Resource, existsOpts ...string) (bool, error) {
```
100mik: Should we be using a simple boolean instead of allowing any number of string arguments to be passed to the `Exists()` function? That would lead to a lot of refactoring but I think it makes it more robust? Maybe create a struct like `ExistsOpts` if we wanna accommodate more options in the future?

rohitagg2020: Changed to have a struct instead of string variadac.

pkg/kapp/resources/resources.go
```diff
@@ -401,6 +401,18 @@ func (c *ResourcesImpl) Exists(resource Resource) (bool, error) {
 			return isDone, c.resourceErr(err, "Checking existence of", resource)
 		}
 
+		if existsOpts.checkForSameUID() {
```
cppforlife: instead of having a method that returns the field value, it's probably better to just use existsOps.SameUID directly.

pkg/kapp/resources/resources.go
```diff
@@ -401,6 +401,18 @@ func (c *ResourcesImpl) Exists(resource Resource) (bool, error) {
...
+				// If fetchedRes(i.e. resource from K8s) is not null and its UID didn't match with the
+				// UID of resource we are trying to delete, then it means resource has been deleted
+				// successfully.
+				if fetchedRes != nil {
+					if string(fetchedRes.GetUID()) != resource.UID() {
```
cppforlife: we should be able to use .UID() instead of GetUID()

pkg/kapp/resources/resources.go
rohitagg2020: I was thinking adding a function with name existsOpts.checkForSameUID() gives more context on what we want to do as compared to existsOps.SameUID. Your thoughts ?

pkg/kapp/resources/resources.go
rohitagg2020: As it is of type Unstructured struct from k8s/apimachinery, it doesn't have UID() function. It has SetUID() and GetUID() functions. https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured

pkg/kapp/resources/exists_options.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package resources
+
+type ExistsOpts struct {
+	SameUID bool
+}
```
cppforlife: i would move this to resources.go next to Resources interface (since it's part of the interface).

pkg/kapp/resources/resources.go
```diff
@@ -401,6 +405,16 @@ func (c *ResourcesImpl) Exists(resource Resource) (bool, error) {
 			return isDone, c.resourceErr(err, "Checking existence of", resource)
 		}
 
+		// Check if we have to compare the UID's also to confirm if it is same resource.
+		if existsOpts.SameUID {
+			if fetchedRes != nil {
```
praveenrewar, line 410: Should we return in case `fetchedRes` is nil to avoid nested if statements?
Although it does feel more readable this way.

cppforlife: > Although it does feel more readable this way.

i would stick with current setup as well.

pkg/kapp/resources/resources.go
```diff
@@ -401,6 +405,16 @@ func (c *ResourcesImpl) Exists(resource Resource) (bool, error) {
...
+				if string(fetchedRes.GetUID()) != resource.UID() {
+					found = false
+					return true, nil
+				}
```
rohitagg2020, line 414: There is 1 more case needs to be handled. What if fetched resource is nil. Will include that as well.

pkg/kapp/resources/resources.go
cppforlife: > Although it does feel more readable this way.

i would agree current form is more readable.

pkg/kapp/resources/resources.go
cppforlife: it just falls through to lines 418-419

PR #299, opened 2021-08-31
Improving how warnings are displayed

### LINKED ISSUE #287: All K8s warning displayed while running any command
**What steps did you take:**
Run a `deploy`, `delete` or `inspect` command. 

**What happened:**
All k8s deprecation warnings for the version are displayed.

```zsh
Warning: admissionregistration.k8s.io/v1beta1 ValidatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 ValidatingWebhookConfiguration
Warning: rbac.authorization.k8s.io/v1beta1 ClusterRole is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRole
Warning: apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
Warning: networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
Warning: extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
Warning: rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
Warning: storage.k8s.io/v1beta1 VolumeAttachment is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 VolumeAttachment
Warning: apiregistration.k8s.io/v1beta1 APIService is deprecated in v1.19+, unavailable in v1.22+; use apiregistration.k8s.io/v1 APIService
Warning: rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRoleBinding
Warning: admissionregistration.k8s.io/v1beta1 MutatingWebhookConfiguration is deprecated in v1.16+, unavailable in v1.22+; use admissionregistration.k8s.io/v1 MutatingWebhookConfiguration
Warning: policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
Warning: coordination.k8s.io/v1beta1 Lease is deprecated in v1.19+, unavailable in v1.22+; use coordination.k8s.io/v1 Lease
Warning: storage.k8s.io/v1beta1 CSINode is deprecated in v1.17+, unavailable in v1.22+; use storage.k8s.io/v1 CSINode
Warning: networking.k8s.io/v1beta1 IngressClass is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 IngressClassList
Warning: discovery.k8s.io/v1beta1 EndpointSlice is deprecated in v1.21+, unavailable in v1.25+; use discovery.k8s.io/v1 EndpointSlice
Warning: storage.k8s.io/v1beta1 CSIDriver is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 CSIDriver
Warning: scheduling.k8s.io/v1beta1 PriorityClass is deprecated in v1.14+, unavailable in v1.22+; use scheduling.k8s.io/v1 PriorityClass
Warning: rbac.authorization.k8s.io/v1beta1 Role is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 Role
Warning: policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
Warning: batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
Warning: storage.k8s.io/v1beta1 StorageClass is deprecated in v1.19+, unavailable in v1.22+; use storage.k8s.io/v1 StorageClass
Warning: certificates.k8s.io/v1beta1 CertificateSigningRequest is deprecated in v1.19+, unavailable in v1.22+; use certificates.k8s.io/v1 CertificateSigningRequest
```

**What did you expect:**
Warnings should be shown only when a resource being mentioned in the warnings are being operated on.

**Environment:**

- kapp version : 0.38.0
- OS : MacOS 11.5.2 
- Kubernetes version 1.21

---
Vote on this request

100mik:
Changes:
- `deps_factory.go` now has a function which returns a `DynamicClient` which does not produce warnings. Called `MutedDynamicClient`.
- The `ResourcesImpl` now has a `MutedDynamicClient` , `NewResourcesImpl()` and all references updated accordingly.
- `All()` in `resources.go` uses a `MutedDynamicClient` to make API calls.

### DIFF
--- a/pkg/kapp/cmd/app/inspect.go
+++ b/pkg/kapp/cmd/app/inspect.go
@@ -11,6 +11,7 @@ import (
 	cmdtools "github.com/k14s/kapp/pkg/kapp/cmd/tools"
 	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
 	"github.com/k14s/kapp/pkg/kapp/logger"
+	"github.com/k14s/kapp/pkg/kapp/resources"
 	"github.com/spf13/cobra"
 )
 
@@ -86,6 +87,8 @@ func (o *InspectOptions) Run() error {
 	resources = resourceFilter.Apply(resources)
 	source := fmt.Sprintf("app '%s'", app.Name())
 
+	o.checkDeprecation(resources, supportObjs)
+
 	switch {
 	case o.Raw:
 		for _, res := range resources {
@@ -118,4 +121,10 @@ func (o *InspectOptions) Run() error {
 	}
 
 	return nil
+}
+
+func (o *InspectOptions) checkDeprecation(resources []resources.Resource, supportingObjs FactorySupportObjs) {
+	for _, r := range resources {
+		supportingObjs.IdentifiedResources.Get(r)
+	}
 }
--- a/pkg/kapp/cmd/app/inspect.go
+++ b/pkg/kapp/cmd/app/inspect.go
@@ -11,7 +11,6 @@ import (
 	cmdtools "github.com/k14s/kapp/pkg/kapp/cmd/tools"
 	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
 	"github.com/k14s/kapp/pkg/kapp/logger"
-	"github.com/k14s/kapp/pkg/kapp/resources"
 	"github.com/spf13/cobra"
 )
 
@@ -87,8 +86,6 @@ func (o *InspectOptions) Run() error {
 	resources = resourceFilter.Apply(resources)
 	source := fmt.Sprintf("app '%s'", app.Name())
 
-	o.checkDeprecation(resources, supportObjs)
-
 	switch {
 	case o.Raw:
 		for _, res := range resources {
@@ -121,10 +118,4 @@ func (o *InspectOptions) Run() error {
 	}
 
 	return nil
-}
-
-func (o *InspectOptions) checkDeprecation(resources []resources.Resource, supportingObjs FactorySupportObjs) {
-	for _, r := range resources {
-		supportingObjs.IdentifiedResources.Get(r)
-	}
 }
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -16,8 +16,7 @@ import (
 )
 
 type DepsFactory interface {
-	DynamicClient() (dynamic.Interface, error)
-	MutedDynamicClient() (dynamic.Interface, error)
+	DynamicClient(warnings bool) (dynamic.Interface, error)
 	CoreClient() (kubernetes.Interface, error)
 	ConfigureWarnings(warnings bool)
 }
@@ -39,7 +38,7 @@ func NewDepsFactoryImpl(configFactory ConfigFactory, ui ui.UI) *DepsFactoryImpl 
 		printTargetOnce: &sync.Once{}}
 }
 
-func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
+func (f *DepsFactoryImpl) DynamicClient(warnings bool) (dynamic.Interface, error) {
 	config, err := f.configFactory.RESTConfig()
 	if err != nil {
 		return nil, err
@@ -47,27 +46,12 @@ func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
 
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
-	cpConfig.WarningHandler = f.newWarningHandler()
 
-	clientset, err := dynamic.NewForConfig(cpConfig)
-	if err != nil {
-		return nil, fmt.Errorf("Building Dynamic clientset: %s", err)
+	if warnings {
+		cpConfig.WarningHandler = f.newWarningHandler()
+	} else {
+		cpConfig.WarningHandler = rest.NoWarnings{}
 	}
-
-	f.printTarget(config)
-
-	return clientset, nil
-}
-
-func (f *DepsFactoryImpl) MutedDynamicClient() (dynamic.Interface, error) {
-	config, err := f.configFactory.RESTConfig()
-	if err != nil {
-		return nil, err
-	}
-
-	// copy to avoid mutating the passed-in config
-	cpConfig := rest.CopyConfig(config)
-	cpConfig.WarningHandler = rest.NoWarnings{}
 
 	clientset, err := dynamic.NewForConfig(cpConfig)
 	if err != nil {
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -132,7 +132,7 @@ func (f *DepsFactoryImpl) newWarningHandler() rest.WarningHandler {
 	}
 	options := rest.WarningWriterOptions{
 		Deduplicate: true,
-		Color:       true,
+		Color:       false,
 	}
 	warningWriter := rest.NewWarningWriter(uiWriter{ui: f.ui}, options)
 	return warningWriter
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -16,7 +16,7 @@ import (
 )
 
 type DepsFactory interface {
-	DynamicClient(warnings bool) (dynamic.Interface, error)
+	DynamicClient(opts DynamicClientOpts) (dynamic.Interface, error)
 	CoreClient() (kubernetes.Interface, error)
 	ConfigureWarnings(warnings bool)
 }
@@ -38,7 +38,11 @@ func NewDepsFactoryImpl(configFactory ConfigFactory, ui ui.UI) *DepsFactoryImpl 
 		printTargetOnce: &sync.Once{}}
 }
 
-func (f *DepsFactoryImpl) DynamicClient(warnings bool) (dynamic.Interface, error) {
+type DynamicClientOpts struct {
+	Warnings bool
+}
+
+func (f *DepsFactoryImpl) DynamicClient(opts DynamicClientOpts) (dynamic.Interface, error) {
 	config, err := f.configFactory.RESTConfig()
 	if err != nil {
 		return nil, err
@@ -47,7 +51,7 @@ func (f *DepsFactoryImpl) DynamicClient(warnings bool) (dynamic.Interface, error
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
 
-	if warnings {
+	if opts.Warnings {
 		cpConfig.WarningHandler = f.newWarningHandler()
 	} else {
 		cpConfig.WarningHandler = rest.NoWarnings{}
--- a/pkg/kapp/cmd/core/deps_factory.go
+++ b/pkg/kapp/cmd/core/deps_factory.go
@@ -17,6 +17,7 @@ import (
 
 type DepsFactory interface {
 	DynamicClient() (dynamic.Interface, error)
+	MutedDynamicClient() (dynamic.Interface, error)
 	CoreClient() (kubernetes.Interface, error)
 	ConfigureWarnings(warnings bool)
 }
@@ -47,6 +48,26 @@ func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
 	cpConfig.WarningHandler = f.newWarningHandler()
+
+	clientset, err := dynamic.NewForConfig(cpConfig)
+	if err != nil {
+		return nil, fmt.Errorf("Building Dynamic clientset: %s", err)
+	}
+
+	f.printTarget(config)
+
+	return clientset, nil
+}
+
+func (f *DepsFactoryImpl) MutedDynamicClient() (dynamic.Interface, error) {
+	config, err := f.configFactory.RESTConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	// copy to avoid mutating the passed-in config
+	cpConfig := rest.CopyConfig(config)
+	cpConfig.WarningHandler = rest.NoWarnings{}
 
 	clientset, err := dynamic.NewForConfig(cpConfig)
 	if err != nil {
--- a/pkg/kapp/cmd/tools/list_labels.go
+++ b/pkg/kapp/cmd/tools/list_labels.go
@@ -122,12 +122,12 @@ func (o *ListLabelsOptions) listResources() ([]ctlres.Resource, error) {
 		return nil, err
 	}
 
-	dynamicClient, err := o.depsFactory.DynamicClient(true)
+	dynamicClient, err := o.depsFactory.DynamicClient(cmdcore.DynamicClientOpts{Warnings: true})
 	if err != nil {
 		return nil, err
 	}
 
-	mutedDynamicClient, err := o.depsFactory.DynamicClient(false)
+	mutedDynamicClient, err := o.depsFactory.DynamicClient(cmdcore.DynamicClientOpts{Warnings: false})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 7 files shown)


pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -17,6 +17,7 @@ import (
 
 type DepsFactory interface {
 	DynamicClient() (dynamic.Interface, error)
+	MutedDynamicClient() (dynamic.Interface, error)
```
praveenrewar: IMO we should not call it `MutetdDynamicClient` as we are planning to use it for all the internal(internal to kapp like list) client-go calls.

100mik: Something like `InternalDynamicClient` ? Do you have any other suggestions ?

pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -58,6 +59,26 @@ func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
 	return clientset, nil
 }
 
+func (f *DepsFactoryImpl) MutedDynamicClient() (dynamic.Interface, error) {
+	config, err := f.configFactory.RESTConfig()
```
rohitagg2020, line 63: As the code is almost similar to Dynamic Client, should we have a common function being called from both MutedDynamicClient and DynamicClient by passing the boolean flag for enabling or disabling the warning?

rohitagg2020, line 46: As the code is almost similar to Dynamic Client, should we have a common function being called from both MutedDynamicClient and DynamicClient by passing the boolean flag for enabling or disabling the warning?

pkg/kapp/cmd/tools/list_labels.go
```diff
@@ -127,8 +127,13 @@ func (o *ListLabelsOptions) listResources() ([]ctlres.Resource, error) {
 		return nil, err
 	}
 
+	mutedDynamicClient, err := o.depsFactory.DynamicClient()
```
rohitagg2020: Will it be o.depsFactory.DynamicClient() or o.depsFactory.MutedDynamicClient()?

pkg/kapp/cmd/core/deps_factory.go
100mik: Yes well do that. Doing that and updating references for `DynamicClient()` makes sense

pkg/kapp/cmd/tools/list_labels.go
100mik: Yep it will be `MutedDynamicClient`, will change this

pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -39,36 +38,21 @@ func NewDepsFactoryImpl(configFactory ConfigFactory, ui ui.UI) *DepsFactoryImpl
 		printTargetOnce: &sync.Once{}}
 }
 
-func (f *DepsFactoryImpl) DynamicClient() (dynamic.Interface, error) {
+func (f *DepsFactoryImpl) DynamicClient(warnings bool) (dynamic.Interface, error) {
```
100mik: Do we wanna have something like `DynamicClientOpts` which we pass to this function  or is a parameter fine because we are not looking to add more options right now?

pkg/kapp/cmd/tools/list_labels.go
```diff
@@ -122,13 +122,18 @@ func (o *ListLabelsOptions) listResources() ([]ctlres.Resource, error) {
 		return nil, err
 	}
 
-	dynamicClient, err := o.depsFactory.DynamicClient()
```
praveenrewar, line 125: Is this the only place where we are calling this method?

pkg/kapp/resources/resources.go
```diff
@@ -52,6 +52,7 @@ type ResourcesImpl struct {
 	resourceTypes             ResourceTypes
 	coreClient                kubernetes.Interface
 	dynamicClient             dynamic.Interface
+	mutedDynamicClient        dynamic.Interface
```
praveenrewar, line 55: Are we planning to use mutex so that we don't use both the clients at the same time? If so, is there any way we can avoid making changes in all the places they are used?

100mik: Had a discussion about this, but seems like we won't need the mutex unless the clients are used in parallel, which is not the case right now. I tried figuring out a way with lesser files changed but since we instantiate clients in the `reps_factory.go` but use the instances that are stored in the struct in `resources.go` while API calls are made adding a common wrapper might be a bit complicated. Certainly worth digging into but might add unnecessary complexity

pkg/kapp/cmd/app/inspect.go
```diff
@@ -119,3 +122,9 @@ func (o *InspectOptions) Run() error {
 
 	return nil
 }
+
+func (o *InspectOptions) checkDeprecation(resources []resources.Resource, supportingObjs FactorySupportObjs) {
```
100mik: Can checking for deprecation of filtered resources on a remote cluster in this manner be too much of an overhead if the number of resources increases?
It attempts to `Get()` each filtered resource.

pkg/kapp/cmd/app/inspect.go
```diff
@@ -86,6 +87,8 @@ func (o *InspectOptions) Run() error {
 	resources = resourceFilter.Apply(resources)
 	source := fmt.Sprintf("app '%s'", app.Name())
 
+	o.checkDeprecation(resources, supportObjs)
```
cppforlife: TODO remove this bit since we do not want to make additional Get calls.

pkg/kapp/resources/resources.go
```diff
@@ -472,12 +475,20 @@ func (c *ResourcesImpl) resourceErr(err error, action string, resource Resource)
 	return resourcePlainErr{err, action, resource}
 }
 
-func (c *ResourcesImpl) resourceClient(resource Resource) (dynamic.ResourceInterface, ResourceType, error) {
+type resourceClientOpts struct {
+	warnings bool
```
cppforlife: even though resourceClientOpts is private, it's field warnings _should_ be public since you are setting it in functions above.

pkg/kapp/resources/resources.go
```diff
@@ -472,12 +475,20 @@ func (c *ResourcesImpl) resourceErr(err error, action string, resource Resource)
...
+}
+
+func (c *ResourcesImpl) resourceClient(resource Resource, opts resourceClientOpts) (dynamic.ResourceInterface, ResourceType, error) {
 	resType, err := c.resourceTypes.Find(resource)
 	if err != nil {
 		return nil, ResourceType{}, err
 	}
 
+	if !opts.warnings {
+		return c.mutedDynamicClient.Resource(resType.GroupVersionResource).Namespace(resource.Namespace()), resType, nil
```
cppforlife: instead of doing the same line below, lets have a variable dynamicClient which is set to mutedDynamicClient and if !opts.Warnings then it gets set to dynamicClient. makes it easier to tell that same logic is applied on both clients.

test/e2e/warnings_test.go
```diff
@@ -77,18 +77,18 @@ spec:
...
 			RunOpts{StdinReader: strings.NewReader(yaml)})
 	})
-	logger.Section("deploying with --warnings flag", func() {
+	logger.Section("deploying without --warnings flag", func() {
 		yaml := strings.Replace(crYaml, "<cr-name>", "cr-1", 1)
-		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", crName, "--warnings=true"},
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", crName},
 			RunOpts{StdinReader: strings.NewReader(yaml)})
 
 		if !strings.Contains(out, customWarning) {
```
cppforlife: should we assert on full output to see where warnings are coming up? (for the test below)

pkg/kapp/resources/resources.go
100mik: Made the changes!

pkg/kapp/resources/resources.go
100mik: Improved this conditional!

pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -38,15 +38,24 @@ func NewDepsFactoryImpl(configFactory ConfigFactory, ui ui.UI) *DepsFactoryImpl
...
 	config, err := f.configFactory.RESTConfig()
 	if err != nil {
 		return nil, err
 	}
 
 	// copy to avoid mutating the passed-in config
 	cpConfig := rest.CopyConfig(config)
-	cpConfig.WarningHandler = f.newWarningHandler()
+
+	if opts.Warnings {
```
praveenrewar, line 54: Should we check for `f.Warnings` here itself instead of checking inside the `newWarningHandler()` function, so that we don't repeat using `rest.NoWarnings{}`?
We might want to rename `opts.Warnings` to something else in that case.

pkg/kapp/resources/resources.go
```diff
@@ -472,13 +475,24 @@ func (c *ResourcesImpl) resourceErr(err error, action string, resource Resource)
 	return resourcePlainErr{err, action, resource}
 }
 
-func (c *ResourcesImpl) resourceClient(resource Resource) (dynamic.ResourceInterface, ResourceType, error) {
+type resourceClientOpts struct {
+	Warnings bool
+}
+
+func (c *ResourcesImpl) resourceClient(resource Resource, opts resourceClientOpts) (dynamic.ResourceInterface, ResourceType, error) {
```
praveenrewar, line 482: Should we make the options optional?

pkg/kapp/cmd/core/deps_factory.go
100mik: Does make sense, I considered passing the value of `opts.Warnings` to `newWarningHandler()` as well. But I think it also makes sense to keep the conditional checks due to CLI flags and functional parameters separate.

pkg/kapp/resources/resources.go
100mik: All the usages are within this file so updating them wasn't a hassle. Doesn't a struct lead to a more robust function signature that way?

test/e2e/warnings_test.go
100mik: Updated the test cases

pkg/kapp/resources/resources.go
praveenrewar: It's just that the struct is not required except for one place and we are kind of forcing it's usage everywhere.

test/e2e/warnings_test.go
```diff
@@ -72,27 +81,62 @@ spec:
...
+
+Changes
+
+Namespace  Name  Kind     Conds.  Age  Op      Op st.  Wait to    Rs  Ri  
+kapp-test  <cr-name>  CronTab  -       -    create  -       reconcile  -   -  
+
+Op:      1 create, 0 delete, 0 update, 0 noop
+Wait to: 1 reconcile, 0 delete, 0 noop
+
+<replaced>: ---- applying 1 changes [0/1 done] ----<output-warning>
```
100mik: Could have `<output-warning>` in the next line and use a backspace escape character for the case where warning is not shown? Might help if this is not legible enough

pkg/kapp/resources/resources.go
100mik: That is fair. I am still torn about using the triple dot operator to mark an optional parameter as it's actual intention is to accept a slice of values

cppforlife: ... for optional params is really handy for public APIs to preserve backwards compat. in this case its all internal so i would not recommend going with ....

test/e2e/warnings_test.go
```diff
@@ -72,27 +81,62 @@ spec:
   image: my-awesome-cron-image
 `
 
+	expectedOutputTemplate := `
+Target cluster 'https://127.0.0.1:60817' (nodes: minikube)
```
cppforlife: port and nodes would be different probably so dont think we can keep this info.

test/e2e/warnings_test.go
```diff
@@ -24,15 +30,18 @@ func TestWarningsFlag(t *testing.T) {
...
-		kapp.Run([]string{"delete", "-a", crName})
+		kapp.Run([]string{"delete", "-a", crName1})
+		kapp.Run([]string{"delete", "-a", crName2})
 		kapp.Run([]string{"delete", "-a", crdName})
 	}
 
 	cleanUp()
 	defer cleanUp()
 	customWarning := "example.com/v1alpha1 CronTab is deprecated; use example.com/v1 CronTab"
+	outputWarning := fmt.Sprintf("\n%sWarning:%s %s", yellowColor, resetColor, customWarning)
```
cppforlife: kapp has a way to disable colors (--color...). does it work with warnings?

test/e2e/warnings_test.go
```diff
@@ -72,27 +81,62 @@ spec:
   image: my-awesome-cron-image
 `
 
+	expectedOutputTemplate := `
```
cppforlife: lets just duplicate this template for each section, instead of doing bunch of replaces. assertion will become more readable imho.

pkg/kapp/resources/resources.go
100mik: Noted! Resolving this thread for now

test/e2e/warnings_test.go
100mik: We have a function `replaceTarget()` which removes the entire target string, we use this both for expected output and the output itself. We could exclude it in the expected output and call `replaceTarget()` only for the output itself

test/e2e/warnings_test.go
100mik: Noted! will do

test/e2e/warnings_test.go
praveenrewar: These warnings are generated by [client-go](https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/rest/warnings.go#L125), where in the unicode characters are used to add colour. I think we will only be able to disable them entirely and not just for the tests. We can take a look at the --color option to see how it's working.

test/e2e/warnings_test.go
100mik: Would not be necessary as we are not using the template anymore

test/e2e/warnings_test.go
100mik: I just tried it out. It does not seem to work for warnings as the warnings are not generated by us.

praveenrewar: We might be able to use the configure warnings method to disable the colour for warnings whenever the flag is set to false as they both are being used in the same package.

cppforlife: hmm... i vaguely remember also having support for NO_COLOR or something like that (https://github.com/vmware-tanzu/color/commit/ba5534bd301c4471f2a82eb42ad3ed966a6361d5)

cppforlife: i wonder if for warnings we shouldnt use color. at least for now.

100mik: We figured out a way to make the `--color` flag work. Wanna have a look at that?

100mik: Made warnings respect the `FORCE_COLOR` env variable as well

pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -138,12 +139,16 @@ func (f *DepsFactoryImpl) newWarningHandler() rest.WarningHandler {
 	}
 	options := rest.WarningWriterOptions{
 		Deduplicate: true,
-		Color:       f.WarningsColor,
+		Color:       f.WarningsColor && envVarsAllowColors(),
```
praveenrewar: Is it fine to use this environment variable here? 
I was thinking of disabling the colouring of warnings for now and later on implement a new method in ui library to display warnings, which will control the colouring as well. IMO that would be more neat and clean.

PR #300, opened 2021-09-01
Added retryable etcdserver errors to the retry list

### LINKED ISSUE #106: [apply retry] retry on etcdserver leader changed errors
This tends to happen when I deploy my full dev cluster form scratch which is a lot of resources

```
kapp: Error: waiting on reconcile ......: etcdserver: leader changed
```

It would be good if kapp detected this error and retried

100mik:


pkg/kapp/resources/resources.go
```diff
@@ -453,6 +453,14 @@ func (c *ResourcesImpl) isServerRescaleErr(err error) bool {
 	return false
 }
 
+// Retries retryable errors thrown by etcd server.
+// Comprehensive list of errors at : https://github.com/etcd-io/etcd/blob/main/server/etcdserver/errors.go
+// Addresses : https://github.com/vmware-tanzu/carvel-kapp/issues/106
+func (c *ResourcesImpl) isEtcdRetryableError(err error) bool {
+	return strings.Contains(err.Error(), "etcdserver:") &&
+		(strings.Contains(err.Error(), "timed out") || strings.Contains(err.Error(), "leader changed"))
```
cppforlife: should we do this via a regexp?

100mik: I think something as simple as `((etcdserver:).*(leader changed|timed out))` should work. Sounds like the better option. I will test it out and make the change

pkg/kapp/resources/resources.go
```diff
@@ -515,6 +515,10 @@ var (
 	//   Post https://cert-manager-webhook.cert-manager.svc:443/convert?timeout=30s:
 	//   x509: certificate signed by unknown authority (reason: )
 	conversionWebhookErrCheck = regexp.MustCompile("conversion webhook for (.+) failed:")
+
+	// Matches retryable etcdserver errors
+	// Comprehensive list of errors at : https://github.com/etcd-io/etcd/blob/main/server/etcdserver/errors.go
+	etcdserverRetryableErrCheck = regexp.MustCompile("etcdserver:(.+)(leader changed|timed out)")
```
praveenrewar, line 521: Shouldn't there be `(.+)` at the end as well?

praveenrewar: Resolving it as it returns true for a substring match as well.

PR #308, opened 2021-09-07
Updating the dependency to fix the security issue reported by trivy

rohitagg2020:
Updating the dependency to fix the security issue reported by trivy


100mik:
We seem to have bumped up versions for a number of dependencies instead of enforcing a certain version of `gogo/protobuff` as an indirect dependency. Would that have been the better approach? This seems to touch a lot of vendored dependencies.

rohitagg2020:
> We seem to have bumped up versions for a number of dependencies instead of enforcing a certain version of `gogo/protobuff` as an indirect dependency. Would that have been the better approach? This seems to touch a lot of vendored dependencies.

it does work but we have taken a decision to bump the direct dependency rather than doing indirect way

PR #309, opened 2021-09-08
Update error message for invalid change rule annotation values

### LINKED ISSUE #178: Output context about the offending annotation on a change rule error
I was deploying a few yaml resources and got the error:
```
kapp: Error: Change graph:
  Calculating required deps:
    Expected change rule annotation value to have 4 pieces but was 1
```
This wasn't very helpful for tracking down where in the yaml the issue was.

I would like to see some context about the offending annotation. Outputting the annotation that caused the error as part of the message is one option.

praveenrewar:
Update the error message which is shown when change rule annotation value has wrong number of arguments/pieces. The message should be more descriptive and provide a better idea of the format.

### DIFF
--- a/pkg/kapp/diffgraph/change_rule.go
+++ b/pkg/kapp/diffgraph/change_rule.go
@@ -38,7 +38,7 @@ func NewChangeRuleFromAnnString(ann string) (ChangeRule, error) {
 	pieces := strings.Split(ann, " ")
 	if len(pieces) != 4 {
 		return ChangeRule{}, fmt.Errorf(
-			"Expected change rule annotation value to have 4 pieces but was %d", len(pieces))
+			"Expected change rule annotation value to have format '(upsert|delete) (before|after) (upserting|deleting) (change-group)', but was '%s'", ann)
 	}
 
 	rule := ChangeRule{
--- a/pkg/kapp/diffgraph/change.go
+++ b/pkg/kapp/diffgraph/change.go
@@ -141,7 +141,7 @@ func (c *Change) AllRules() ([]ChangeRule, error) {
 		if k == changeRuleAnnKey || strings.HasPrefix(k, changeRuleAnnPrefixKey) {
 			rule, err := NewChangeRuleFromAnnString(v)
 			if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/kapp/diffgraph/change_rule.go
```diff
@@ -38,7 +38,7 @@ func NewChangeRuleFromAnnString(ann string) (ChangeRule, error) {
 	pieces := strings.Split(ann, " ")
 	if len(pieces) != 4 {
 		return ChangeRule{}, fmt.Errorf(
-			"Expected change rule annotation value to have 4 pieces but was %d", len(pieces))
+			"Change rule annotation value should be of the format: (upsert|delete) (before|after) (upserting|deleting) (groupName)")
```
rohitagg2020: May be instead of groupName if we have ChangeGroupName, will that be more explicit ?

praveenrewar: Actually, targetGroupName might be a better choice, but I am not sure which is more descriptive.

cppforlife: lets tweak it to

```
Expected change rule annotation value to have format: '(upsert|delete) (before|after) (upserting|deleting) (change-group)'
```

praveenrewar: Done

praveenrewar: @cppforlife  Error message now looks like this

Screenshot 2021-09-09 at 7 47 33 PM

renuy: Sentence looks incomplete. Suggestions:
Expected change rule annotation value to have format: '(upsert|delete) (before|after) (upserting|deleting) (change-group)', provided 'server'

praveenrewar: Updated the message to 
```
Expected change rule annotation value to have format: '(upsert|delete) (before|after) (upserting|deleting) (change-group)', but was 'server'
```

@renuy @rohitagg2020 @cppforlife

PR #310, opened 2021-09-09
Add change-group placeholders for granular ordering of resources

### LINKED ISSUE #297: Granular ordering of CRDs/namespace-scoped resources
Currently the ordering of kapp is implemented in such a way that all the CRs are dependent on all the CRDs and all the namespaced resources are dependent on all the namespaces and so on.

We should update the ordering graph in such a way that all the CRs are dependent on their respective CRDs, all the namespaced resources are dependent on their respective namespaces and so on.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritise our backlog. Use the "smiley face" up to the right of this comment to vote.

praveenrewar:
Add change-group placeholders to allow dynamic update of change-group and change-rule. 
Use the placeholders in default config to have granular ordering of CRDs and Namespaces.
The present set of allowed placeholders are:
- `{api-group}`  Resource apiGroup
- `{kind}` Resource Kind 
- `{name}` Resource Name
- `{namespace}` Resource Namespace
- `{crd-kind}` CRD Kind (`spec.names.kind`)
- `{crd-group}` CRD Group (`spec.groups`)

These placeholders can be used to dynamically update change-groups or target group inside a change-rule.


pkg/kapp/diffgraph/change.go
```diff
@@ -232,3 +239,53 @@ func (cs Changes) MatchesRule(rule ChangeRule, exceptChange *Change) ([]*Change,
...
+		case namespaceDefaultChangeGroup:
+			group.Name += "-" + change.Change.Resource().Name()
+		case crdDefaultChangeGroup:
+			group.Name += "-" + strings.ToLower(change.CRDKind())
+		}
+	}
+}
+
+func (rule *ChangeRule) UpdateDefault(change *Change) {
+	if change.Change.Op() == ActualChangeOpUpsert {
```
praveenrewar: Should we update delete ordering as well? This could lead to a bit of complexity as there are multiple change-rules associated with a single change-rule

pkg/kapp/diffgraph/change.go
```diff
@@ -232,3 +241,27 @@ func (cs Changes) MatchesRule(rule ChangeRule, exceptChange *Change) ([]*Change,
 
 	return result, nil
 }
+
+func parseChangeGroupPlaceholders(resource ctlres.Resource, changeGroup string) (parsed string, err error) {
```
praveenrewar: @cppforlife We might want to refactor this function if the wildcard approach seems good.

pkg/kapp/config/default.go
```diff
@@ -386,10 +386,18 @@ changeGroupBindings:
...
+- name: change-groups.kapp.k14s.io/crds-{crd-group}-{name}
+  resourceMatchers: &crdMatchers
+  - apiGroupKindMatcher: {kind: CustomResourceDefinition, apiGroup: apiextensions.k8s.io}
+
 - name: change-groups.kapp.k14s.io/namespaces
   resourceMatchers: &namespaceMatchers
   - apiGroupKindMatcher: {kind: Namespace, apiGroup: ""}
 
+- name: change-groups.kapp.k14s.io/namespaces-{name}
+  resourceMatchers: &namespaceMatchers
```
rohitagg2020, line 393: We can reuse namespaceMatcher from above instead of redeclaring again.

pkg/kapp/config/default.go
```diff
@@ -386,10 +386,18 @@ changeGroupBindings:
   resourceMatchers: &crdMatchers
   - apiGroupKindMatcher: {kind: CustomResourceDefinition, apiGroup: apiextensions.k8s.io}
 
+- name: change-groups.kapp.k14s.io/crds-{crd-group}-{name}
+  resourceMatchers: &crdMatchers
```
rohitagg2020: We can reuse crdMatcher from above instead of redeclaring again.

pkg/kapp/diffgraph/change.go
```diff
@@ -152,6 +157,10 @@ func (c *Change) AllRules() ([]ChangeRule, error) {
 
 		if (ctlres.AnyMatcher{rms}).Matches(res) {
 			for _, ruleStr := range ruleConfig.Rules {
+				ruleStr, err := parseChangeGroupPlaceholders(c.Change.Resource(), ruleStr)
```
rohitagg2020: parseChangeGroupPlaceholders will never return error as per implementation. If that function is not supposed to return error, then we can omit it.

pkg/kapp/config/default.go
praveenrewar: Makes sense, will do.

pkg/kapp/config/default.go
praveenrewar: Makes sense, will make the change.

pkg/kapp/diffgraph/change.go
praveenrewar: It's returning error in cases of CRDs, but we have to refactor this anyways, so we can discuss about it.

100mik: I think what Rohit was trying to say is that when `crd.Name()` or `crd.Group()` errors out we return `parsed, nil` when we should be returning the error itself (`parsed, err`) in lines 251 and 255. I will go ahead and make this change

pkg/kapp/config/default.go
100mik: Resolved in amended commit

pkg/kapp/diffgraph/change.go
100mik: Resolved in amended commit

pkg/kapp/config/default.go
100mik: Resolved in amended commit

pkg/kapp/config/default.go
```diff
@@ -386,10 +386,16 @@ changeGroupBindings:
   resourceMatchers: &crdMatchers
   - apiGroupKindMatcher: {kind: CustomResourceDefinition, apiGroup: apiextensions.k8s.io}
 
+- name: change-groups.kapp.k14s.io/crds-{crd-group}-{name}
```
praveenrewar:
```suggestion
- name: change-groups.kapp.k14s.io/crds-{crd-group}-{crd-kind}
```

100mik:
- Removed [unused change group](https://github.com/vmware-tanzu/carvel-kapp/pull/310/files#diff-93c184d7793d59702521d5cd6e870417573e8fc77490db32b978c36671f3734aL389) `change-groups.kapp.k14s.io/namespaces` from `default.go`
- Added a [check](https://github.com/vmware-tanzu/carvel-kapp/pull/310/files#diff-e68a56a3dde9732bfff340a37a8599828e21eddb6f4a490f267dff010674ec28R44) which causes kapp to error out if CRD specific placeholders are used with non-CRD resources
- Changed placeholder [names](https://github.com/vmware-tanzu/carvel-kapp/pull/310/files#diff-e68a56a3dde9732bfff340a37a8599828e21eddb6f4a490f267dff010674ec28R73) to be more legible.
- Place holders can now be used in annotations.

pkg/kapp/diffgraph/placeholder_parser.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (p PlaceholderParser) placeholders() (placeholders []string) {
+	for k := range p.values {
+		placeholders = append(placeholders, k)
+	}
+	return placeholders
+}
+
+func valueMapFromResource(resource ctlres.Resource) (values map[string]string, err error) {
```
cppforlife: nit: we typically do not use named returns

pkg/kapp/diffgraph/placeholder_parser.go
```diff
@@ -0,0 +1,83 @@
...
+	placeholderMatcher = regexp.MustCompile("{.+?}")
+)
+
+func (p PlaceholderParser) Parse(val string) (string, error) {
+	placeholders := placeholderMatcher.FindAllString(val, -1)
+
+	for _, placeholder := range placeholders {
+		value, found := p.values[placeholder]
+		if !found {
+			return val, fmt.Errorf(`Expected placeholder to be one of these: %s but was %s`, p.placeholders(), placeholder)
```
cppforlife: nit: tick -> double quotes

pkg/kapp/diffgraph/placeholder_parser.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		if value == "" {
+			return val, fmt.Errorf("Placeholder %s does not have a value for target resource, placeholders with the 'crd-' prefix can only be used with CRDs", placeholder)
```
cppforlife: nit: i would place `placeholders with the 'crd-' prefix can only be used with CRDs` as `(hint: placeholders with the 'crd-' prefix can only be used with CRDs)`

pkg/kapp/diffgraph/placeholder_parser.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+	ctlcrd "github.com/k14s/kapp/pkg/kapp/resourcesmisc"
+)
+
+type PlaceholderParser struct {
```
cppforlife: in a context of diffgraph PlaceholderParser is a bit too general of a name. how about:

```
PlaceholderParser                -> ChangeGroupName
NewPlaceholderParserWithResource -> NewChangeGroupNameForResource(name, res) => ChangeGroupName
PlaceholderParser.Parse(...)     -> ChangeGroupName.AsString() => (name, error)
```

(placeholder is more of an implementation detail)

i would also recommend shifting valueMapFromResource call into AsString (currently Parse), away from NewChangeGroupNameForResource as there are benefits of not doing any work in constructors (aka delayed execution). in our current usage the only benefit we get is that we can drop parseChangeGroupPlaceholders, and just use NewChangeGroupNameForResource(name, res).AsString().

pkg/kapp/diffgraph/placeholder_parser.go
```diff
@@ -0,0 +1,83 @@
...
+		crdGroup, err = crd.Group()
+		if err != nil {
+			return values, err
+		}
+	}
+
+	values = map[string]string{
+		"{name}":      resource.Name(),
+		"{crd-kind}":  crdKind,
+		"{crd-group}": crdGroup,
```
cppforlife: nit: move to the bottom of the list

pkg/kapp/diffgraph/placeholder_parser.go
```diff
@@ -0,0 +1,83 @@
...
+		"{namespace}": resource.Namespace(),
+		"{api-group}": resource.APIGroup(),
+		"{kind}":      resource.Kind(),
```
cppforlife: nit: move to the top of the list

pkg/kapp/diffgraph/placeholder_parser.go
```diff
@@ -0,0 +1,83 @@
...
+
+func NewPlaceholderParserWithResource(resource ctlres.Resource) (PlaceholderParser, error) {
+	values, err := valueMapFromResource(resource)
+	if err != nil {
+		return PlaceholderParser{}, fmt.Errorf("Could not obtain placeholder values from resource: %s", err)
+	}
+	return PlaceholderParser{values}, nil
+}
+
+func NewPlaceholderParserWithMap(values map[string]string) PlaceholderParser {
```
cppforlife: nit: unused

pkg/kapp/diffgraph/placeholder_parser.go
100mik: Refactored `placeholder_parser.go` to `change_group_name.go` keeping in mind the recommended changes

pkg/kapp/diffgraph/change_group_name.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+	ctlcrd "github.com/k14s/kapp/pkg/kapp/resourcesmisc"
+)
+
+type ChangeGroupName struct {
```
100mik, line 14: @cppforlife Just realised that we are also using the `AsString()` method to[ parse change rules](https://github.com/vmware-tanzu/carvel-kapp/pull/310/commits/2fc07f72b807def6bacc5d57f264b7cb7698ac8e#diff-72f6f14971e177e069bd13bca3d398b63f3896678b931fc1e57f873dedec2762R167) (the change group names contained in them) 
Should we consider a better name maybe?

pkg/kapp/diffgraph/change_graph_test.go
```diff
@@ -95,6 +95,10 @@ kind: CustomResourceDefinition
 apiVersion: apiextensions.k8s.io/v1
 metadata:
   name: app-config
+spec:
+  group: app-group
+  names:
+    kind: app-kind
```
cppforlife, line 101: overall looks pretty good. lets add a test in this file that asserts that CRDs and namespaces using are ordered correctly by the default config.

pkg/kapp/diffgraph/change_group_name.go
```diff
@@ -0,0 +1,78 @@
...
+
+	for _, placeholder := range placeholders {
+		value, found := values[placeholder]
+		if !found {
+			return c.name, fmt.Errorf("Expected placeholder to be one of these: %s but was %s", c.placeholders(values), placeholder)
+		}
+		if value == "" {
+			return c.name, fmt.Errorf("Placeholder %s does not have a value for target resource (hint: placeholders with the 'crd-' prefix can only be used with CRDs)", placeholder)
+		}
+		c.name = strings.Replace(c.name, placeholder, value, 1)
```
cppforlife: any reason to specificy exactly 1 replacement?

100mik: In line 31 we are getting all the placeholders and replacing them with respective values one by one. So if there are two `{name}` placeholders, a replace for it is attempted twice even if the first replace takes care of all occurrences.

No specific reason otherwise. Also, does not seem likely that the same placeholder is used twice though

pkg/kapp/diffgraph/change_group_name.go
```diff
@@ -0,0 +1,78 @@
...
+	values := map[string]string{
+		"{api-group}": c.resource.APIGroup(),
+		"{kind}":      c.resource.Kind(),
+		"{name}":      c.resource.Name(),
+		"{namespace}": c.resource.Namespace(),
+		"{crd-kind}":  crdKind,
+		"{crd-group}": crdGroup,
+	}
+
+	for _, placeholder := range placeholders {
```
100mik: Attempt a replace using regex and a replacement function ([possible solution](https://pkg.go.dev/regexp#Regexp.ReplaceAllFunc))

100mik: Resolved in commit 7f798e0

pkg/kapp/diffgraph/change_group_name.go
```diff
@@ -0,0 +1,72 @@
...
+	values := map[string]string{
+		"{api-group}": c.resource.APIGroup(),
+		"{kind}":      c.resource.Kind(),
+		"{name}":      c.resource.Name(),
+		"{namespace}": c.resource.Namespace(),
+		"{crd-kind}":  crdKind,
+		"{crd-group}": crdGroup,
+	}
+
+	replaced := placeholderMatcher.ReplaceAllStringFunc(c.name, func(placeholder string) string {
```
100mik, line 53: @cppforlife Looks good to you?

pkg/kapp/diffgraph/change_graph_test.go
praveenrewar: Done

pkg/kapp/diffgraph/change_graph_test.go
```diff
@@ -465,6 +469,73 @@ metadata:
 	}
 }
 
+func TestChangeGraphWithChangeGroupPlaceholderInConfDefaults(t *testing.T) {
```
praveenrewar: Still thinking of a good name

cppforlife: TestChangeGraphWithChangeGroupPlaceholderInConfDefaults -> TestChangeGraphWithNamespaceAndCRDs

pkg/kapp/diffgraph/change_graph_test.go
```diff
@@ -465,6 +469,73 @@ metadata:
...
+	configYAML := `
+apiVersion: v1
```
cppforlife, line 474: i think we need to have multiple namespaces and multiple CRDs to see that the relationship is granular. otherwise it's not different from today.

pkg/kapp/diffgraph/change_group_name.go
cppforlife: yup, that looks cleaner/simpler.

pkg/kapp/diffgraph/change_graph_test.go
praveenrewar: Makes sense, I will add more namespaces and CRDs

praveenrewar: Done

pkg/kapp/diffgraph/change_graph_test.go
praveenrewar: Done

pkg/kapp/diffgraph/change_graph_test.go
praveenrewar: @cppforlife Added more namespaces, namespaced resources, CRDs and CRs.

PR #311, opened 2021-09-14
Add 'versioned-explicit-ref' annotation

### LINKED ISSUE #286: Restart deployment/daemonset when referred resources' content is changing
Hi team, I met a senario when using kapp's versioned annotation.  Opened this issue to track the question.
The scenario is:
**config C refers a secret S in its data,  and this config C is referred by a Deployment D. Will update the content of secret S trigger restart of Deployment D?**

I get a solution that is referring the secret S directly in Depoyment D. But I think that may be not the best way.

Will kapp's versioned annotation supports above scenario? Or already supported? 
Thanks.

100mik:
Synthetic reference to a versioned resource.
For example: 
```yaml
kapp.k14s.io/versioned-explicit-ref: |
    apiVersion: v1
    kind: Secret
    name: root-password
```
The value of the key `name` is to be replaced by the latest versioned name of the referred resource
Pending:
- [x]  E2E tests

### DIFF
--- a/pkg/kapp/diff/explicit_versioned_ref.go
+++ b/pkg/kapp/diff/explicit_versioned_ref.go
@@ -1,0 +1,90 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package diff
+
+import (
+	"fmt"
+)
+
+type VersionedRefDesc struct {
+	Namespace string `json:"namespace"`
+	APIGroup  string `json:"apiGroup"`
+	Kind      string `json:"kind"`
+	Name      string `json:"name"`
+}
+
+type ExplicitVersionedRefAnn struct {
+	References     []VersionedRefDesc `json:"references"`
+	VersionedNames map[string]string
+}
+
+type ExplicitVersionedRef struct {
+	Resource   VersionedResource
+	Annotation ExplicitVersionedRefAnn
+}
+
+const (
+	explicitReferenceKey = "kapp.k14s.io/versioned-explicit-ref"
+)
+
+func NewExplicitVersionedRef(res VersionedResource, annotation ExplicitVersionedRefAnn) *ExplicitVersionedRef {
+	return &ExplicitVersionedRef{res, annotation}
+}
+
+// Returns true if the
+func (e *ExplicitVersionedRef) IsReferenced() (bool, error) {
+	references, err := e.refStringList()
+	if err != nil {
+		return false, err
+	}
+
+	referenceKey := e.resourceRefString()
+
+	for _, v := range references {
+		if v == referenceKey {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (e *ExplicitVersionedRef) refStringList() ([]string, error) {
+	list := []string{}
+	for _, v := range e.Annotation.References {
+		v, err := e.validateAndReplaceCoreGroup(v)
+		if err != nil {
+			return list, err
+		}
+
+		list = append(list, fmt.Sprintf("%s/%s/%s/%s", v.Namespace, v.APIGroup, v.Kind, v.Name))
+	}
+	return list, nil
+}
+
+func (e *ExplicitVersionedRef) validateAndReplaceCoreGroup(resourceDescription VersionedRefDesc) (VersionedRefDesc, error) {
+	// Replacing APIGroup value "core" with an empty string
+	// Edgecase for resources that are a part of "core"
+	if resourceDescription.APIGroup == "core" {
+		resourceDescription.APIGroup = ""
+	}
+
+	if resourceDescription.Kind == "" || resourceDescription.Name == "" {
+		return resourceDescription, fmt.Errorf("Explicit Reference Error: Name and Kind are required values in an explicit reference")
+	}
+	return resourceDescription, nil
+}
+
+func (e *ExplicitVersionedRef) resourceRefString() string {
+	return e.Resource.UniqVersionedKey().String()
+}
+
+/*
+Annotation with a list of references in JSON format:
+
+kapp.k14s.io/versioned-explicit-ref: '{ "references": [ { "namespace": <resource-namespace>, "apiGroup":  <resource-apiGroup>, "kind" : <resource-kind>, "name":  <resource-name> } ] }'
+
+"namespace" need not be assigned a value for cluster-scoped resources.
+"apiGroup" need not be assigned a value for resources belonging to "core" API group.
+*/
--- a/pkg/kapp/diff/explicit_versioned_ref.go
+++ b/pkg/kapp/diff/explicit_versioned_ref.go
@@ -32,14 +32,14 @@ func NewExplicitVersionedRef(res VersionedResource, annotation ExplicitVersioned
 	return &ExplicitVersionedRef{res, annotation}
 }
 
-// Returns true if the
+// Returns true if the resource is referenced by the annotation
 func (e *ExplicitVersionedRef) IsReferenced() (bool, error) {
-	references, err := e.refStringList()
+	references, err := e.references()
 	if err != nil {
 		return false, err
 	}
 
-	referenceKey := e.resourceRefString()
+	referenceKey := e.referenceKey()
 
 	for _, v := range references {
 		if v == referenceKey {
@@ -50,7 +50,7 @@ func (e *ExplicitVersionedRef) IsReferenced() (bool, error) {
 	return false, nil
 }
 
-func (e *ExplicitVersionedRef) refStringList() ([]string, error) {
+func (e *ExplicitVersionedRef) references() ([]string, error) {
 	list := []string{}
 	for _, v := range e.Annotation.References {
 		v, err := e.validateAndReplaceCoreGroup(v)
@@ -76,7 +76,7 @@ func (e *ExplicitVersionedRef) validateAndReplaceCoreGroup(resourceDescription V
 	return resourceDescription, nil
 }
 
-func (e *ExplicitVersionedRef) resourceRefString() string {
+func (e *ExplicitVersionedRef) referenceKey() string {
 	return e.Resource.UniqVersionedKey().String()
 }
 
--- a/pkg/kapp/diff/explicit_versioned_ref.go
+++ b/pkg/kapp/diff/explicit_versioned_ref.go
@@ -5,79 +5,69 @@ package diff
 
 import (
 	"fmt"
+	"reflect"
+
+	"gopkg.in/yaml.v2"
 )
 
 type VersionedRefDesc struct {
-	Namespace string `json:"namespace"`
-	APIGroup  string `json:"apiGroup"`
-	Kind      string `json:"kind"`
-	Name      string `json:"name"`
-}
-
-type ExplicitVersionedRefAnn struct {
-	References     []VersionedRefDesc `json:"references"`
-	VersionedNames map[string]string
+	Namespace  string `yaml:"namespace"`
+	APIVersion string `yaml:"apiVersion"`
+	Kind       string `yaml:"kind"`
+	Name       string `yaml:"name"`
 }
 
 type ExplicitVersionedRef struct {
 	Resource   VersionedResource
-	Annotation ExplicitVersionedRefAnn
+	Annotation string
 }
 
 const (
-	explicitReferenceKey = "kapp.k14s.io/versioned-explicit-ref"
+	explicitReferenceKey       = "kapp.k14s.io/versioned-explicit-ref"
+	explicitReferenceKeyPrefix = "kapp.k14s.io/versioned-explicit-ref."
 )
 
-func NewExplicitVersionedRef(res VersionedResource, annotation ExplicitVersionedRefAnn) *ExplicitVersionedRef {
+func NewExplicitVersionedRef(res VersionedResource, annotation string) *ExplicitVersionedRef {
 	return &ExplicitVersionedRef{res, annotation}
 }
 
 // Returns true if the resource is referenced by the annotation
 func (e *ExplicitVersionedRef) IsReferenced() (bool, error) {
-	references, err := e.references()
+	reference := VersionedRefDesc{}
+	err := yaml.Unmarshal([]byte(e.Annotation), &reference)
 	if err != nil {
-		return false, err
+		return false, fmt.Errorf("Error unmarshalling versioned reference: %s", err)
 	}
 
-	referenceKey := e.referenceKey()
-
-	for _, v := range references {
-		if v == referenceKey {
-			return true, nil
-		}
+	if reference.APIVersion == "" || reference.Kind == "" || reference.Name == "" {
+		return false, fmt.Errorf("Explicit reference error: apiVersion, kind and name are required values in an explicit versioned reference")
 	}
 
-	return false, nil
+	baseName, _ := e.Resource.BaseNameAndVersion()
+	versionedResourceDesc := VersionedRefDesc{
+		Namespace:  e.Resource.res.Namespace(),
+		APIVersion: e.Resource.res.APIVersion(),
+		Kind:       e.Resource.res.Kind(),
+		Name:       baseName,
+	}
+
+	return reflect.DeepEqual(versionedResourceDesc, reference), nil
 }
 
-func (e *ExplicitVersionedRef) references() ([]string, error) {
-	list := []string{}
-	for _, v := range e.Annotation.References {
-		v, err := e.validateAndReplaceCoreGroup(v)
-		if err != nil {
-			return list, err
-		}
-
-		list = append(list, fmt.Sprintf("%s/%s/%s/%s", v.Namespace, v.APIGroup, v.Kind, v.Name))
-	}
-	return list, nil
-}
-
-func (e *ExplicitVersionedRef) validateAndReplaceCoreGroup(resourceDescription VersionedRefDesc) (VersionedRefDesc, error) {
-	// Replacing APIGroup value "core" with an empty string
-	// Edgecase for resources that are a part of "core"
-	if resourceDescription.APIGroup == "core" {
-		resourceDescription.APIGroup = ""
+func (e *ExplicitVersionedRef) VersionedReference() (string, error) {
+	reference := VersionedRefDesc{}
+	err := yaml.Unmarshal([]byte(e.Annotation), &reference)
+	if err != nil {
+		return "", fmt.Errorf("Error unmarshalling versioned reference: %s", err)
 	}
 
-	if resourceDescription.Kind == "" || resourceDescription.Name == "" {
-		return resourceDescription, fmt.Errorf("Explicit Reference Error: Name and Kind are required values in an explicit reference")
+	reference.Name = e.Resource.res.Name()
+	versionedReference, err := yaml.Marshal(reference)
+	if err != nil {
+		return "", fmt.Errorf("Error marshalling versioned reference: %s", err)
 	}
-	return resourceDescription, nil
-}
 
-func (e *ExplicitVersionedRef) referenceKey() string {
-	return e.Resource.UniqVersionedKey().String()
+	return string(versionedReference), nil
 }
 
 /*
--- a/pkg/kapp/diff/explicit_versioned_ref.go
+++ b/pkg/kapp/diff/explicit_versioned_ref.go
@@ -5,21 +5,14 @@ package diff
 
 import (
 	"fmt"
-	"reflect"
 
+	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
 	"gopkg.in/yaml.v2"
 )
 
-type VersionedRefDesc struct {
-	Namespace  string `yaml:"namespace"`
-	APIVersion string `yaml:"apiVersion"`
-	Kind       string `yaml:"kind"`
-	Name       string `yaml:"name"`
-}
-
 type ExplicitVersionedRef struct {
-	Resource   VersionedResource
-	Annotation string
+	AnnotationKey string
+	Annotation    string
 }
 
 const (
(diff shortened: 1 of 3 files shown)


cppforlife:
questions to be resolved:

- apigroup vs apiversion
- namespacing (especially when resource doesnt specify explicit ns). is this similar to what we do with other refs?
- cluster scoped resources (how it works today and how should ann work with that?)

100mik:
We have to consider a resource that might wanna have multiple versioned resources referenced explicitly.

100mik:
Changed how explicit references are structured. Multiple references are now supported using the format specified [here](https://github.com/vmware-tanzu/carvel-kapp/pull/311/commits/09853d1cda2e5892154a140a548467f0854c0a1b#diff-3a4768871151150b99289f09a0d3b8a6be2926d600300178104cc94ac894a460R64). The format should also address our other requirements.

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -0,0 +1,70 @@
...
+	explicitReferenceKey = "kapp.k14s.io/versioned-explicit-ref"
+)
+
+func NewExplicitVersionedRef(res VersionedResource, annotation ExplicitVersionedRefAnn) *ExplicitVersionedRef {
+	return &ExplicitVersionedRef{res, annotation}
+}
+
+func (e *ExplicitVersionedRef) IsReferenced() bool {
+	refList := e.refStringList()
+	resRef := e.resourceRefString()
```
praveenrewar: Is my understanding correct that we are getting a list of references and then a reference key to check if it matches with any of the references? 

If yes, should we have something like this?
```suggestion
	references := e.references()
	referenceKey := e.resourceRefKey()
```

100mik: Yep that's right! That does seem to be more intuitive I'll make that change

100mik: Resolved!

praveenrewar: Should we also change the function names?

100mik: Makes sense to, will do!

100mik:
Will attempt appending existing annotation instead of adding labels. Do we also wanna avoid adding more annotations?
cc: @cppforlife

cppforlife:
> Do we also wanna avoid adding more annotations?

yup, as much as possible. (especially ones that we would have to add automatically)

100mik:
- The annotation now has a [map](https://github.com/vmware-tanzu/carvel-kapp/pull/311/commits/0475c8ab1fec7b4041a59e38d451c6000592a495#diff-3a4768871151150b99289f09a0d3b8a6be2926d600300178104cc94ac894a460R17) which maps unique keys to the latest versioned name.
- This map is [populated](https://github.com/vmware-tanzu/carvel-kapp/pull/311/commits/0475c8ab1fec7b4041a59e38d451c6000592a495#diff-1299fda59a62e741c24d715184df939466614418c142b1dbf81cf9fc8840590aR105) for a resource if it's reference is found in the list of user-defined resources.

For example: 
For annotation value,
```json
{
   "references":[
      {
         "namespace":"default",
         "kind":"ConfigMap",
         "name":"config-1"
      },
      {
         "namespace":"default",
         "kind":"ConfigMap",
         "name":"config-2"
      }
   ]
}
```
The updated value is:
```json
{
   "references":[
      {
         "namespace":"default",
         "apiGroup":"",
         "kind":"ConfigMap",
         "name":"config-1"
      },
      {
         "namespace":"default",
         "apiGroup":"",
         "kind":"ConfigMap",
         "name":"config-2"
      }
   ],
   "VersionedNames":{
      "default//ConfigMap/config-1":"config-1-ver-2",
      "default//ConfigMap/config-2":"config-2-ver-3"
   }
}
```
The versioned names are the updated ones added by kapp. Demo YAML on [gist](https://gist.github.com/100mik/be7f0b942aad46cb9b38addef1898ad6#file-explicit-reference-yaml).

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -0,0 +1,90 @@
...
+		if err != nil {
+			return list, err
+		}
+
+		list = append(list, fmt.Sprintf("%s/%s/%s/%s", v.Namespace, v.APIGroup, v.Kind, v.Name))
+	}
+	return list, nil
+}
+
+func (e *ExplicitVersionedRef) validateAndReplaceCoreGroup(resourceDescription VersionedRefDesc) (VersionedRefDesc, error) {
```
100mik: Is this too much overhead for an advanced feature?

praveenrewar: I feel that the validation part is bit too much, as anyone using kapp would know that these are the basic things required to identify a resource. Although, I think that having it is not a bad idea either.

test/e2e/versioned_explicit_reference_test.go
```diff
@@ -0,0 +1,107 @@
...
+	yaml2 := `
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-1
+  annotations:
+    kapp.k14s.io/versioned: ""
+data:
+  foo: bar1
```
praveenrewar: Should we use a totally different value instead of `bar1`, like `blahblahblah`, which is easy to identify? Right now this is the only difference in yaml1 and yaml2, so it's a bit difficult to spot the difference. Or maybe we can use replace, as it would be used only in a few places.

100mik: Not sure about a replace, but yep a significantly different value will certainly help. Will do!

pkg/kapp/diff/explicit_versioned_ref.go
100mik: Agreed. I guess I will leave as is for now

cppforlife:
lets refine this a bit more...

let's support `kapp.k14s.io/versioned-explicit-ref.X` annotation where X can be anything like number or some string. resource can have multiple of those annotations. value of that annotation would be `{"name": "...", "apiGr...}`. when we match  reference we would update `name` field with versioned name. this way this is very similar to regular k8s references:

```yaml
secretRef:
  name: 
  kind: 
  ...
```

another question to think about is, should we use json or yaml within that annotation? e.g.

```
annotations:
  kapp.k14s.io/versioned-explicit-ref.X: |
    name: foo
    namespace: blah
    kind: ConfigMap
```

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -0,0 +1,90 @@
...
+	VersionedNames map[string]string
+}
+
+type ExplicitVersionedRef struct {
+	Resource   VersionedResource
+	Annotation ExplicitVersionedRefAnn
+}
+
+const (
+	explicitReferenceKey = "kapp.k14s.io/versioned-explicit-ref"
```
rohitagg2020: Nit: Move it to Versioned Resource file as it is used in that file only.

100mik: The idea was to keep anything being used to implement this annotation in one file. I would say we should move this [function](https://github.com/vmware-tanzu/carvel-kapp/pull/311/files#diff-1299fda59a62e741c24d715184df939466614418c142b1dbf81cf9fc8840590aR118) to this file as well. Thoughts?

rohitagg2020: No, that function should stay there as it belong to VersionedResource struct.

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -0,0 +1,80 @@
...
+// Returns true if the resource is referenced by the annotation
+func (e *ExplicitVersionedRef) IsReferenced() (bool, error) {
+	reference := VersionedRefDesc{}
+	err := yaml.Unmarshal([]byte(e.Annotation), &reference)
+	if err != nil {
+		return false, fmt.Errorf("Error unmarshalling versioned reference: %s", err)
+	}
+
+	if reference.APIVersion == "" || reference.Kind == "" || reference.Name == "" {
+		return false, fmt.Errorf("Explicit reference error: apiVersion, kind and name are required values in an explicit versioned reference")
```
sethiyash: Nit: Maybe we can have something like this instead of mentioning `Explicit reference` two times - 
```suggestion
		return false, fmt.Errorf("Explicit versioned reference error: apiVersion, kind and name are required values")
```

100mik: Doesn't hurt to be more specific. Could certainly make this change!

pkg/kapp/diff/explicit_versioned_ref.go
100mik: It was added with this change so maybe should be a method of ExplicitVersionedRef struct instead

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -5,76 +5,50 @@ package diff
...
-	err := yaml.Unmarshal([]byte(e.Annotation), &reference)
+func (e *ExplicitVersionedRef) AsObjectRef() (map[string]interface{}, error) {
+	var objectRef map[string]interface{}
+	err := yaml.Unmarshal([]byte(e.Annotation), &objectRef)
 	if err != nil {
-		return false, fmt.Errorf("Error unmarshalling versioned reference: %s", err)
+		return nil, fmt.Errorf("Explicit reference error: could not create object ref for explicit versioned reference")
 	}
 
-	if reference.APIVersion == "" || reference.Kind == "" || reference.Name == "" {
```
100mik: Are there any pros of unmarshalling to a struct like :
```go
type VersionedRefDesc struct {
       Namespace  string `yaml:"namespace"`
       APIVersion string `yaml:"apiVersion"`
       Kind       string `yaml:"kind"`
       Name       string `yaml:"name"`
}
```
before unmarshalling it to a variable of type `map[string]interface{}` ?
I would prefer doing something like `reference.Name` instead of `reference["name"]`.
But I guess it won't matter. It is later unmarshalled to [coreV1.ObjectReference](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5400) which is unlikely to change.

pkg/kapp/diff/versioned_resource.go
```diff
@@ -91,6 +91,29 @@ func (d VersionedResource) updateAffected(rule ctlconf.TemplateRule, rs []ctlres
...
+				if err != nil {
+					return err
+				}
+
+				err = d.buildObjRefReplacementFunc(ctlconf.TemplateAffectedObjRef{})(objectRef)
+				if err != nil {
+					return fmt.Errorf("Error updating explicit versioned reference: %s", err)
+				}
+
+				annotationMod, err := explicitRef.AnnotationMod(objectRef)
```
cppforlife, line 109: skipping err check

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, fmt.Errorf("Explicit versioned reference error: apiVersion, kind and name are required values in an explicit versioned reference")
+	}
+
+	return objectRef, nil
+}
+
+func (e *ExplicitVersionedRef) AnnotationMod(objectRef map[string]interface{}) (ctlres.StringMapAppendMod, error) {
+	value, err := yaml.Marshal(objectRef)
+	if err != nil {
+		return ctlres.StringMapAppendMod{}, fmt.Errorf("Error marshalling explicit reference annotation: %s", err)
```
cppforlife: we are already building an error, so there is no need to say "Error" -- describe the context of the error e.g. 

```
fmt.Errorf("Marshalling explicit reference: %s", err)
```

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewExplicitVersionedRef(annotationKey string, annotation string) *ExplicitVersionedRef {
+	return &ExplicitVersionedRef{annotationKey, annotation}
+}
+
+func (e *ExplicitVersionedRef) AsObjectRef() (map[string]interface{}, error) {
+	var objectRef map[string]interface{}
+	err := yaml.Unmarshal([]byte(e.Annotation), &objectRef)
+	if err != nil {
+		return nil, fmt.Errorf("Explicit reference error: could not create object ref for explicit versioned reference")
```
cppforlife: `fmt.Errorf("Parsing versioned explicit reference from annotation '%s': %s", ...)`

pkg/kapp/diff/explicit_versioned_ref.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	if objectRef["apiVersion"] == "" || objectRef["kind"] == "" || objectRef["name"] == "" {
+		return nil, fmt.Errorf("Explicit versioned reference error: apiVersion, kind and name are required values in an explicit versioned reference")
+	}
```
cppforlife: `fmt.Errorf("Expected versioned explicit reference to specify non-empty apiVersion, kind and name keys")`

test/e2e/versioned_explicit_reference_test.go
```diff
@@ -0,0 +1,113 @@
...
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy initial", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
```
cppforlife: should check diff summary as well.

test/e2e/versioned_explicit_reference_test.go
```diff
@@ -0,0 +1,113 @@
...
+kind: ConfigMap
+metadata:
+  name: config-2
+  annotations:
+    kapp.k14s.io/versioned-explicit-ref: |
+      apiVersion: v1
+      kind: ConfigMap
+      name: config-1
+data:
+  foo: bar
```
cppforlife: probably worth adding some additional refs that do not match and may be an additional resource with more refs.

pkg/kapp/diff/explicit_versioned_ref.go
100mik: `fmt.Errorf("Expected versioned  reference to specify non-empty apiVersion, kind and name keys")` maybe? Typo?

pkg/kapp/diff/explicit_versioned_ref.go
100mik: I think I will add a test to keep these in check

test/e2e/versioned_explicit_reference_test.go
100mik: Will do. Also thinking about adding tests for the key validation since they are [hardcoded keys](https://github.com/vmware-tanzu/carvel-kapp/pull/311/files#diff-3a4768871151150b99289f09a0d3b8a6be2926d600300178104cc94ac894a460R34).

100mik: This does not seem to be possible today as `RunWithOpts` errors out in case of an error.

test/e2e/versioned_explicit_reference_test.go
100mik: Added cases with prefixed annotations and refs which do not get versioned. Checking diff summary now

cppforlife: im not seeing where we check diff for the initial deploy.

100mik: Oh have not checked for that yet. My bad missed it. Will make the change

100mik: I actually thought we wanted to see the text diff. Should we switch back to checking for rows instead of the actual text diff?

pkg/kapp/diff/versioned_resource.go
```diff
@@ -91,6 +91,34 @@ func (d VersionedResource) updateAffected(rule ctlconf.TemplateRule, rs []ctlres
...
+				explicitRef := NewExplicitVersionedRef(k, v)
+				objectRef, err := explicitRef.AsObjectRef()
+				if err != nil {
+					return err
+				}
+
+				// Passing empty TemplateAffectedObjRef as explicit references do not have a special name key
+				err = d.buildObjRefReplacementFunc(ctlconf.TemplateAffectedObjRef{})(objectRef)
+				if err != nil {
+					return fmt.Errorf("Error updating explicit versioned reference: %s", err)
```
cppforlife: `Error updating explicit versioned reference: %s` => `Updating versioned explicit reference: %s`

pkg/kapp/diff/versioned_resource.go
```diff
@@ -91,6 +91,34 @@ func (d VersionedResource) updateAffected(rule ctlconf.TemplateRule, rs []ctlres
 		}
 	}
 
+	for _, res := range rs {
+		for k, v := range res.Annotations() {
+			if k == explicitReferenceKey || strings.HasPrefix(k, explicitReferenceKeyPrefix) {
+				explicitRef := NewExplicitVersionedRef(k, v)
+				objectRef, err := explicitRef.AsObjectRef()
+				if err != nil {
+					return err
```
cppforlife: prolly worth adding `fmt.Errorf("Parsing versioned explicit ref on resource '%s': %s", res.Description(), err)`

pkg/kapp/diff/versioned_resource.go
```diff
@@ -91,6 +91,34 @@ func (d VersionedResource) updateAffected(rule ctlconf.TemplateRule, rs []ctlres
...
+				}
+
+				annotationMod, err := explicitRef.AnnotationMod(objectRef)
+				if err != nil {
+					return err
+				}
+
+				err = annotationMod.Apply(res)
+				if err != nil {
+					return err
```
cppforlife: `fmt.Errorf("Updating versioned explicit ref on resource '%s': %s", res.Description(), err)`

pkg/kapp/diff/versioned_resource.go
100mik: Oops missed this. Made this change

pkg/kapp/diff/versioned_resource.go
100mik: Certainly better. Will do!

pkg/kapp/diff/versioned_resource.go
100mik: Will do!

pkg/kapp/diff/versioned_resource.go
100mik: Putting in something similar at lines 106 and 111 as well

pkg/kapp/diff/versioned_resource.go
100mik: Making this something like `"Parsing versioned explicit ref on resource` as `buildObjRefReplacementFunc` only returns unmarshalling and marshalling errors. Even though it's unlikely to error out after getting parsed in line 100

pkg/kapp/diff/versioned_resource.go
```diff
@@ -91,6 +91,34 @@ func (d VersionedResource) updateAffected(rule ctlconf.TemplateRule, rs []ctlres
...
+				explicitRef := NewExplicitVersionedRef(k, v)
+				objectRef, err := explicitRef.AsObjectRef()
+				if err != nil {
+					return fmt.Errorf("Parsing versioned explicit ref on resource '%s': %s", res.Description(), err)
+				}
+
+				// Passing empty TemplateAffectedObjRef as explicit references do not have a special name key
+				err = d.buildObjRefReplacementFunc(ctlconf.TemplateAffectedObjRef{})(objectRef)
+				if err != nil {
+					return fmt.Errorf("Parsing versioned explicit ref on resource '%s': %s", res.Description(), err)
```
cppforlife: this isnt parsing...

pkg/kapp/diff/versioned_resource.go
```diff
@@ -91,6 +91,34 @@ func (d VersionedResource) updateAffected(rule ctlconf.TemplateRule, rs []ctlres
...
+				}
+
+				annotationMod, err := explicitRef.AnnotationMod(objectRef)
+				if err != nil {
+					return fmt.Errorf("Parsing versioned explicit ref on resource '%s': %s", res.Description(), err)
```
cppforlife: also not parsing...

pkg/kapp/diff/versioned_resource.go
100mik: The only error thrown by these 2 functions are Marshalling/Unmarshalling errors. Doesn't it make sense to wrap those with something like `Parsing error` ?

100mik: If we want to keep messages unique we could have 
```suggestion
					return fmt.Errorf("Creating explicit reference for resource  '%s': %s", res.Description(), err)
```

pkg/kapp/diff/versioned_resource.go
100mik: Same reason as above. Alternatively,
```suggestion
					return fmt.Errorf("Updating versioned explicit ref on resource '%s': %s", res.Description(), err)
```
(Cannot think of a better one at the moment)

cppforlife: preparing annotation modification or something like that. point being here that if we are going to add some context to the error, it should be unique (otherwise we ll be looking at the wrong piece of code if its all the same), and ideally provides a little bit more info to the end user -- not really useful in this specific case.

100mik: Caught your drift! Making the changes

PR #312, opened 2021-09-14
diff filter 

### LINKED ISSUE #182: Add filtering by label
I would like the ability to filter `kapp deploy` using label selectors

jenkins-x 3 will be switching to use kapp to apply resources to a cluster and as a jenkins-x 3 user I would like to use this in conjunction with okteto.

Okteto replaces deploymens with a developer containers to let you code on your app inside kubernetes.  It does this by rewriting the deployment resource.  It also adds a label to indicate that the deployment has been modified by okteto.

I wish to prevent kapp from overriding the okteto managed deployments which would be much easier to do by specifying a 'not' filter using a label selector.  

A possible workaround is to use kubectl to find all resources with the labels, and convert them into a resource filter using kinds and names.

sethiyash:
- new filtering system that would filter the differences between the cluster and local.

### DIFF
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -100,8 +100,8 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
-	// Deciding based on flag from resource filter and diff filter
-	if !shouldFullyDeleteApp || changesSummary.SkippedChanges {
+	// Deciding based on flag from diff filter
+	if changesSummary.SkippedChanges {
 		shouldFullyDeleteApp = false
 		o.ui.PrintLinef("App '%s' (namespace: %s) will not be fully deleted "+
 			"because some resources are excluded by filters",
@@ -203,6 +203,7 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 		appliedChanges := diffFilter.Apply(changes)
 
 		if len(changes) != len(appliedChanges) {
+			// setting it to true when changes are filtered by diffFilter
 			skippedChanges = true
 		}
 
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -182,6 +182,14 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 			return ctlcap.ClusterChangeSet{}, nil, false, err
 		}
 
+		// Apply diff-filter on the changes
+		diffFilter, err := o.DiffFlags.DiffFilter()
+		if err != nil {
+			return ctlcap.ClusterChangeSet{}, nil, false, err
+		}
+
+		filteredChanges := diffFilter.Apply(changes)
+
 		{ // Build cluster changes based on diff changes
 			msgsUI := cmdcore.NewDedupingMessagesUI(cmdcore.NewPlainMessagesUI(o.ui))
 
@@ -194,7 +202,7 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 				changeFactory, changeSetFactory, convergedResFactory, msgsUI)
 
 			clusterChangeSet = ctlcap.NewClusterChangeSet(
-				changes, o.ApplyFlags.ClusterChangeSetOpts, clusterChangeFactory,
+				filteredChanges, o.ApplyFlags.ClusterChangeSetOpts, clusterChangeFactory,
 				conf.ChangeGroupBindings(), conf.ChangeRuleBindings(), msgsUI, o.logger)
 		}
 	}
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -30,7 +30,7 @@ type DeleteOptions struct {
 	ResourceTypesFlags  ResourceTypesFlags
 }
 
-type changesFlag struct {
+type changeFlags struct {
 	hasNoChanges     bool
 	isFullyDeleteApp bool
 }
@@ -179,7 +179,7 @@ func (o *DeleteOptions) existingResources(app ctlapp.App,
 }
 
 func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Resource, conf ctlconf.Conf,
-	supportObjs FactorySupportObjs) (ctlcap.ClusterChangeSet, *ctldgraph.ChangeGraph, changesFlag, error) {
+	supportObjs FactorySupportObjs) (ctlcap.ClusterChangeSet, *ctldgraph.ChangeGraph, changeFlags, error) {
 
 	var (
 		clusterChangeSet ctlcap.ClusterChangeSet
@@ -192,12 +192,12 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 
 		changes, err := changeSetFactory.New(existingResources, nil).Calculate()
 		if err != nil {
-			return ctlcap.ClusterChangeSet{}, nil, changesFlag{}, err
+			return ctlcap.ClusterChangeSet{}, nil, changeFlags{}, err
 		}
 
 		diffFilter, err := o.DiffFlags.DiffFilter()
 		if err != nil {
-			return ctlcap.ClusterChangeSet{}, nil, changesFlag{}, err
+			return ctlcap.ClusterChangeSet{}, nil, changeFlags{}, err
 		}
 
 		appliedChanges := diffFilter.Apply(changes)
@@ -225,7 +225,7 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 
 	clusterChanges, clusterChangesGraph, err := clusterChangeSet.Calculate()
 	if err != nil {
-		return ctlcap.ClusterChangeSet{}, nil, changesFlag{}, err
+		return ctlcap.ClusterChangeSet{}, nil, changeFlags{}, err
 	}
 
 	{ // Present cluster changes in UI
@@ -235,7 +235,7 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 		changeSetView.Print(o.ui)
 	}
 
-	return clusterChangeSet, clusterChangesGraph, changesFlag{len(clusterChanges) == 0, isFullyDeleteApp}, nil
+	return clusterChangeSet, clusterChangesGraph, changeFlags{len(clusterChanges) == 0, isFullyDeleteApp}, nil
 }
 
 const (
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -30,6 +30,11 @@ type DeleteOptions struct {
 	ResourceTypesFlags  ResourceTypesFlags
 }
 
+type changesFlag struct {
+	hasNoChanges     bool
+	isFullyDeleteApp bool
+}
+
 func NewDeleteOptions(ui ui.UI, depsFactory cmdcore.DepsFactory, logger logger.Logger) *DeleteOptions {
 	return &DeleteOptions{ui: ui, depsFactory: depsFactory, logger: logger}
 }
@@ -76,7 +81,7 @@ func (o *DeleteOptions) Run() error {
 
 	failingAPIServicesPolicy.MarkRequiredGVs(usedGVs)
 
-	existingResources, fullyDeleteApp, err := o.existingResources(app, supportObjs)
+	existingResources, isFullyDeleteApp, err := o.existingResources(app, supportObjs)
 	if err != nil {
 		return err
 	}
@@ -86,15 +91,22 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
-	clusterChangeSet, clusterChangesGraph, hasNoChanges, isFullyDeleteApp, err :=
-		o.calculateAndPresentChanges(app, existingResources, conf, supportObjs, fullyDeleteApp)
+	clusterChangeSet, clusterChangesGraph, changesFlag, err :=
+		o.calculateAndPresentChanges(existingResources, conf, supportObjs)
 	if err != nil {
 		if o.DiffFlags.UI && clusterChangesGraph != nil {
 			return o.presentDiffUI(clusterChangesGraph)
 		}
 		return err
 	}
-	fullyDeleteApp = isFullyDeleteApp
+
+	// Deciding based on isFullyDeleteFlag flag from resource filter and diff filter
+	if !isFullyDeleteApp || !changesFlag.isFullyDeleteApp {
+		isFullyDeleteApp = false
+		o.ui.PrintLinef("App '%s' (namespace: %s) will not be fully deleted "+
+			"because some resources are excluded by filters",
+			app.Name(), o.AppFlags.NamespaceFlags.Name)
+	}
 
 	if o.DiffFlags.UI {
 		return o.presentDiffUI(clusterChangesGraph)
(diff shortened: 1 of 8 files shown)


pkg/kapp/cmd/tools/diff_flags.go
```diff
@@ -35,4 +36,6 @@ func (s *DiffFlags) SetWithPrefix(prefix string, cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&s.Mask, prefix+"mask", true, "Apply masking rules")
 
 	cmd.Flags().BoolVar(&s.AgainstLastApplied, prefix+"against-last-applied", true, "Show changes against last applied copy when possible")
+
+	cmd.Flags().StringVar(&s.Filter, prefix+"filter", "", "Filter the differences between new and existing resource (example: )")
```
rohitagg2020: Example is incomplete

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+	return result
+}
+
+func (f DiffFilter) Matches(newResource res.Resource, existingResource res.Resource) bool {
+	if f.BoolFilter != nil {
+		return f.BoolFilter.Matches(newResource, existingResource)
+	}
+	return false
```
rohitagg2020: If there is no filter, then it should return true

pkg/kapp/cmd/app/delete.go
```diff
@@ -182,6 +182,14 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 			return ctlcap.ClusterChangeSet{}, nil, false, err
 		}
 
+		// Apply diff-filter on the changes
```
cppforlife: nit: comment does not say anything that the code is not doing.

pkg/kapp/cmd/app/delete.go
```diff
@@ -182,6 +182,14 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
...
+		diffFilter, err := o.DiffFlags.DiffFilter()
+		if err != nil {
+			return ctlcap.ClusterChangeSet{}, nil, false, err
+		}
+
+		filteredChanges := diffFilter.Apply(changes)
```
cppforlife: nit: `changes = diffFilter.Apply(changes)`. by overriding same var, it's not possible for code below use the wrong var accidently.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -329,6 +329,14 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
 			return clusterChangeSet, nil, false, "", err
 		}
 
+		// Apply diff-filter on the changes
```
cppforlife: (same comments as in delete.go)

pkg/kapp/cmd/tools/diff_flags.go
cppforlife: `Set changes filter (example: {"and":[{"ops":["update"]},{"existingResource":{"kinds":["Deployment"]}]})`

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,233 @@
...
+	"encoding/json"
+	"fmt"
+
+	"github.com/k14s/kapp/pkg/kapp/matcher"
+	res "github.com/k14s/kapp/pkg/kapp/resources"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+type ChangeSetFilter struct {
+	df     DiffFilter
```
cppforlife: no need for this field

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,233 @@
...
+type DiffFilter struct {
+	Kinds          []string
+	Namespaces     []string
+	Names          []string
+	KindNames      []string
+	KindNamespaces []string
+	KindNsNames    []string
+	LabelSelector  []string
+	BoolFilter     *BoolFilter
+}
```
cppforlife: no need for this struct.

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,233 @@
...
+
+type BoolFilter struct {
```
cppforlife: BoolFilter -> ChangeSetFilterRoot

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,233 @@
...
+	And              []BoolFilter
+	Or               []BoolFilter
+	Not              *BoolFilter
+	NewResource      *DiffFilter
+	ExistingResource *DiffFilter
```
cppforlife: type for both of these should be *ctlres.ResourceFilter

pkg/kapp/diff/filter.go
cppforlife: lets also add `ops []string` to allow selecting changes by their operation.

pkg/kapp/cmd/tools/diff_flags.go
sethiyash: Done

sethiyash: Sure

pkg/kapp/diff/filter.go
sethiyash: removed

pkg/kapp/diff/filter.go
sethiyash: removed

pkg/kapp/diff/filter.go
sethiyash: Done

sethiyash: I have created a `type OpsFilter []ChangeOp` will it be fine ?

pkg/kapp/diff/filter.go
sethiyash: Done

pkg/kapp/cmd/app/delete.go
sethiyash: Yeah removed.

pkg/kapp/cmd/app/delete.go
sethiyash: Yeah sure started overriding same var

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,111 @@
...
+		changeSetFilter, err := NewChangeSetFilterFromString(s.Filter)
+		if err != nil {
+			return ChangeSetFilterRoot{}, err
+		}
+		filter = *changeSetFilter
+	}
+	return filter, nil
+}
+
+func NewChangeSetFilterFromString(data string) (*ChangeSetFilterRoot, error) {
```
rohitagg2020: Function says NewChangeSetFilter but it returns ChangeSetFilterRoot

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,111 @@
...
+	Not              *ChangeSetFilterRoot
+	Ops              OpsFilter
+	NewResource      *ctlres.ResourceFilter
+	ExistingResource *ctlres.ResourceFilter
+}
+
+func (s *ChangeSetFilter) DiffFilter() (ChangeSetFilterRoot, error) {
+	var filter ChangeSetFilterRoot
+	if len(s.Filter) > 0 {
+		changeSetFilter, err := NewChangeSetFilterFromString(s.Filter)
```
rohitagg2020: Nit: Variable name can be changeSetFilterRoot as NewChangeSetFilterFromString returns changeSetFilterRoot

pkg/kapp/diff/filter.go
```diff
@@ -0,0 +1,111 @@
...
+		return f.ExistingResource.Matches(change.ExistingResource())
+	}
+
+	if len(f.Ops) > 0 && change.Op() != "" {
+		return f.Ops.Matches(change)
+	}
+	return false
+}
+
+func (f ChangeSetFilterRoot) IsEmpty() bool {
```
rohitagg2020: Nit: This can be at line num: 63 so that all the ChangeSetFilterRoot functions are all together.

pkg/kapp/cmd/tools/diff_flags.go
```diff
@@ -35,4 +36,6 @@ func (s *DiffFlags) SetWithPrefix(prefix string, cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&s.Mask, prefix+"mask", true, "Apply masking rules")
 
 	cmd.Flags().BoolVar(&s.AgainstLastApplied, prefix+"against-last-applied", true, "Show changes against last applied copy when possible")
+
+	cmd.Flags().StringVar(&s.Filter, prefix+"filter", "", "Set changes filter (example: {\"and\":[{\"ops\":[\"update\"]},{\"existingResource\":{\"kinds\":[\"Deployment\"]}]})")
```
cppforlife: nit, instead of `"` you can use `, so that you dont have to escape "

pkg/kapp/diff/filter.go
cppforlife: i think we can drop IsEmpty function entirely and not short circuit above.

pkg/kapp/diff/filter.go
cppforlife: i would just flip this entire function

```
func (s *ChangeSetFilter) DiffFilter() (ChangeSetFilterRoot, error) {
	if len(s.Filter) == 0 {
           return ChangeSetFilterRoot{}, nil
        }
        return NewChangeSetFilterFromString(s.Filter)
}
```

pkg/kapp/diff/filter.go
rohitagg2020: @cppforlife: Reason we dont want to short circuit above is because we think that in most of the cases, there will be some filter provided? I am thinking from performance perspective.

pkg/kapp/diff/filter.go
sethiyash: Done

pkg/kapp/diff/filter.go
sethiyash: Done

pkg/kapp/cmd/tools/diff_flags.go
sethiyash: Done

pkg/kapp/cmd/app/delete.go
```diff
@@ -154,11 +155,8 @@ func (o *DeleteOptions) existingResources(app ctlapp.App,
 	fullyDeleteApp := true
 	applicableExistingResources := resourceFilter.Apply(existingResources)
 
-	if len(applicableExistingResources) != len(existingResources) {
-		fullyDeleteApp = false
-		o.ui.PrintLinef("App '%s' (namespace: %s) will not be fully deleted "+
-			"because some resources are excluded by filters",
-			app.Name(), o.AppFlags.NamespaceFlags.Name)
+	if !resourceFilter.IsEmpty() {
```
sethiyash, line 158: This condition is important as `len(applicableExistingResources)` and `len(existingResources)` can be equal in two cases - 
1. On applying filter every resource is matched 
2. filter flag is not there then also we match all the resources 
So, to make sure we will mark `fullyDeleteApp ` flag as true only when user provided filter flag and everything matched.

pkg/kapp/cmd/app/delete.go
```diff
@@ -30,6 +30,11 @@ type DeleteOptions struct {
 	ResourceTypesFlags  ResourceTypesFlags
 }
 
+type changesFlag struct {
```
rohitagg2020: nit: We can rename to changeFlags ?

rohitagg2020 (approved):

pkg/kapp/cmd/app/delete.go
```diff
@@ -30,6 +30,11 @@ type DeleteOptions struct {
 	ResourceTypesFlags  ResourceTypesFlags
 }
 
+type changeFlags struct {
```
cppforlife: changeFlags -> changesSummary

pkg/kapp/cmd/app/delete.go
```diff
@@ -30,6 +30,11 @@ type DeleteOptions struct {
...
+	hasNoChanges     bool
```
cppforlife: should be public since it's being accessed outside of the struct.

pkg/kapp/cmd/app/delete.go
```diff
@@ -30,6 +30,11 @@ type DeleteOptions struct {
...
+	isFullyDeleteApp bool
```
cppforlife: instead of tracking isFullyDeleteApp, it's probably better to track whether changes were skipped. `SkippedChanges bool`

pkg/kapp/cmd/app/delete.go
```diff
@@ -95,13 +100,21 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
+	// Deciding based on isFullyDeleteFlag flag from resource filter and diff filter
```
cppforlife: before this if, i would recommend to reset isFullyDeleteApp to something like

```
if changesSummary.SkippedChanges {
  isFullyDeleteApp = false
}
```

also i would recommend renaming it back to `fullyDeleteApp`, or to `shouldFullyDeleteApp`

pkg/kapp/cmd/app/delete.go
```diff
@@ -211,7 +235,7 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 		changeSetView.Print(o.ui)
 	}
 
-	return clusterChangeSet, clusterChangesGraph, (len(clusterChanges) == 0), nil
+	return clusterChangeSet, clusterChangesGraph, changeFlags{len(clusterChanges) == 0, isFullyDeleteApp}, nil
```
cppforlife: nit: when building a struct with fields of the same type, it's better to use field names to make sure we are setting correct fields.

pkg/kapp/cmd/app/delete.go
sethiyash: Ok sure. So `SkippedChanges` flag will be false by default and I can make it true when 
```
if len(changes) != len(appliedChanges) {
	skippedChanges = true
}
```

pkg/kapp/cmd/app/delete.go
sethiyash: I believe instead of reseting `isFullyDeleteApp` flag now we can change the condition at 104 to something like this - 
```
if !shouldFullyDeleteApp || changesSummary.SkippedChanges {
		shouldFullyDeleteApp = false
		o.ui.PrintLinef("App '%s' (namespace: %s) will not be fully deleted "+
			"because some resources are excluded by filters",
			app.Name(), o.AppFlags.NamespaceFlags.Name)
}
```
Here I renamed it to `shouldFullyDeleteApp`

pkg/kapp/cmd/app/delete.go
sethiyash: Done

pkg/kapp/cmd/app/delete.go
sethiyash: Done

pkg/kapp/cmd/app/delete.go
sethiyash: Done

pkg/kapp/cmd/app/delete.go
```diff
@@ -95,13 +100,21 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
+	// Deciding based on flag from diff filter
+	if changesSummary.SkippedChanges {
+		shouldFullyDeleteApp = false
+		o.ui.PrintLinef("App '%s' (namespace: %s) will not be fully deleted "+
```
cppforlife: i think this reads better

```
if changesSummary.SkippedChanges {
  shouldFullyDeleteApp = false 
}

if !shouldFullyDeleteApp {
  o.ui.PrintLinef...
```

PR #319, opened 2021-09-23
Add exists operation for resources with exists annotation

praveenrewar:
Add support for exists annotation.
Add exists operation for resources with exists annotation. 
The exists operation will check if the resource exists or not and retry if it doesn't.

### DIFF
--- a/pkg/kapp/clusterapply/cluster_change_set.go
+++ b/pkg/kapp/clusterapply/cluster_change_set.go
@@ -42,19 +42,16 @@ func (c ClusterChangeSet) Calculate() ([]*ClusterChange, *ctldgraph.ChangeGraph,
 		clusterChange := c.clusterChangeFactory.NewClusterChange(change)
 		wrappedClusterChanges = append(wrappedClusterChanges, wrappedClusterChange{clusterChange})
 	}
-
 	changesGraph, err := ctldgraph.NewChangeGraph(wrappedClusterChanges,
 		c.changeGroupBindings, c.changeRuleBindings, c.logger)
 	if err != nil {
 		// Return graph for inspection
 		return nil, changesGraph, err
 	}
-
 	changesGraph.AllMatching(func(change *ctldgraph.Change) bool {
 		c.markChangesToWait(change)
 		return false
 	})
-
 	// Prune out changes that are not involved with anything
 	changesGraph.RemoveMatching(func(change *ctldgraph.Change) bool {
 		clusterChange := change.Change.(wrappedClusterChange).ClusterChange
@@ -141,7 +138,6 @@ type wrappedClusterChange struct {
 
 func (c wrappedClusterChange) Op() ctldgraph.ActualChangeOp {
 	op := c.ApplyOp()
-
 	switch op {
 	case ClusterChangeApplyOpAdd, ClusterChangeApplyOpUpdate:
 		return ctldgraph.ActualChangeOpUpsert
@@ -151,6 +147,9 @@ func (c wrappedClusterChange) Op() ctldgraph.ActualChangeOp {
 
 	case ClusterChangeApplyOpNoop:
 		return ctldgraph.ActualChangeOpNoop
+
+	case ClusterChangeApplyOpExists:
+		return ctldgraph.ActualChangeOpExists
 
 	default:
 		panic(fmt.Sprintf("Unknown change apply operation: %s", op))
--- a/pkg/kapp/clusterapply/cluster_change_set.go
+++ b/pkg/kapp/clusterapply/cluster_change_set.go
@@ -153,7 +153,7 @@ func (c wrappedClusterChange) Op() ctldgraph.ActualChangeOp {
 		return ctldgraph.ActualChangeOpNoop
 
 	case ClusterChangeApplyOpExists:
-		return ctldgraph.ActualChangeOpExists
+		return ctldgraph.ActualChangeOpUpsert
 
 	default:
 		panic(fmt.Sprintf("Unknown change apply operation: %s", op))
--- a/pkg/kapp/clusterapply/cluster_change_set.go
+++ b/pkg/kapp/clusterapply/cluster_change_set.go
@@ -42,16 +42,19 @@ func (c ClusterChangeSet) Calculate() ([]*ClusterChange, *ctldgraph.ChangeGraph,
 		clusterChange := c.clusterChangeFactory.NewClusterChange(change)
 		wrappedClusterChanges = append(wrappedClusterChanges, wrappedClusterChange{clusterChange})
 	}
+
 	changesGraph, err := ctldgraph.NewChangeGraph(wrappedClusterChanges,
 		c.changeGroupBindings, c.changeRuleBindings, c.logger)
 	if err != nil {
 		// Return graph for inspection
 		return nil, changesGraph, err
 	}
+
 	changesGraph.AllMatching(func(change *ctldgraph.Change) bool {
 		c.markChangesToWait(change)
 		return false
 	})
+
 	// Prune out changes that are not involved with anything
 	changesGraph.RemoveMatching(func(change *ctldgraph.Change) bool {
 		clusterChange := change.Change.(wrappedClusterChange).ClusterChange
@@ -138,6 +141,7 @@ type wrappedClusterChange struct {
 
 func (c wrappedClusterChange) Op() ctldgraph.ActualChangeOp {
 	op := c.ApplyOp()
+
 	switch op {
 	case ClusterChangeApplyOpAdd, ClusterChangeApplyOpUpdate:
 		return ctldgraph.ActualChangeOpUpsert
--- a/pkg/kapp/clusterapply/exists_change.go
+++ b/pkg/kapp/clusterapply/exists_change.go
@@ -33,13 +33,13 @@ func (e ExistsStrategy) Apply() error {
 		if err != nil {
 			return err
 		}
-		return &ExistsChangeError{}
+		return ExistsChangeError{}
 	}
 	return nil
 }
 
 type ExistsChangeError struct{}
 
-func (e *ExistsChangeError) Error() string {
+func (e ExistsChangeError) Error() string {
 	return fmt.Sprint("External resource doesn't exists")
 }
--- a/pkg/kapp/clusterapply/exists_change.go
+++ b/pkg/kapp/clusterapply/exists_change.go
@@ -1,0 +1,36 @@
+// Copyright 2021 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package clusterapply
+
+import (
+	"errors"
+
+	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
+	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+)
+
+type ExistsChange struct {
+	change              ctldiff.Change
+	identifiedResources ctlres.IdentifiedResources
+}
+
+func (c ExistsChange) ApplyStrategy() (ApplyStrategy, error) {
+	res := c.change.NewResource()
+	return WaitStrategy{res, c}, nil
+}
+
+type WaitStrategy struct {
+	res ctlres.Resource
+	d   ExistsChange
+}
+
+func (c WaitStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
+
+func (c WaitStrategy) Apply() error {
+	exists, _ := c.d.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
+	if exists {
+		return nil
+	}
+	return errors.New("Resource doesn't exists")
+}
--- a/pkg/kapp/clusterapply/exists_change.go
+++ b/pkg/kapp/clusterapply/exists_change.go
@@ -32,5 +32,5 @@ func (c WaitStrategy) Apply() error {
 	if exists {
 		return nil
 	}
-	return errors.New("Resource doesn't exists")
+	return errors.New("Placeholder resource doesn't exists")
 }
--- a/pkg/kapp/clusterapply/exists_change.go
+++ b/pkg/kapp/clusterapply/exists_change.go
@@ -4,7 +4,7 @@
 package clusterapply
 
 import (
-	"errors"
+	"fmt"
 
 	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
@@ -17,23 +17,29 @@ type ExistsChange struct {
 
 func (c ExistsChange) ApplyStrategy() (ApplyStrategy, error) {
 	res := c.change.NewResource()
-	return WaitStrategy{res, c}, nil
+	return ExistsStrategy{res, c.identifiedResources}, nil
 }
 
-type WaitStrategy struct {
-	res ctlres.Resource
-	e   ExistsChange
+type ExistsStrategy struct {
+	res                 ctlres.Resource
+	identifiedResources ctlres.IdentifiedResources
 }
 
-func (c WaitStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
+func (e ExistsStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
 
-func (c WaitStrategy) Apply() error {
-	exists, err := c.e.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
+func (e ExistsStrategy) Apply() error {
+	exists, err := e.identifiedResources.Exists(e.res, ctlres.ExistsOpts{})
 	if !exists {
 		if err != nil {
 			return err
 		}
-		return errors.New("Placeholder resource doesn't exists")
+		return &ExistsChangeError{}
 	}
 	return nil
+}
+
+type ExistsChangeError struct{}
+
+func (e *ExistsChangeError) Error() string {
+	return fmt.Sprint("External resource doesn't exists")
 }
--- a/pkg/kapp/clusterapply/exists_change.go
+++ b/pkg/kapp/clusterapply/exists_change.go
@@ -22,13 +22,13 @@ func (c ExistsChange) ApplyStrategy() (ApplyStrategy, error) {
 
 type WaitStrategy struct {
 	res ctlres.Resource
-	d   ExistsChange
+	e   ExistsChange
 }
 
 func (c WaitStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
 
 func (c WaitStrategy) Apply() error {
-	exists, _ := c.d.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
+	exists, _ := c.e.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
 	if exists {
 		return nil
 	}
--- a/pkg/kapp/clusterapply/exists_change.go
+++ b/pkg/kapp/clusterapply/exists_change.go
@@ -28,9 +28,12 @@ type WaitStrategy struct {
 func (c WaitStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
 
 func (c WaitStrategy) Apply() error {
-	exists, _ := c.e.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
-	if exists {
-		return nil
+	exists, err := c.e.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
+	if !exists {
+		if err != nil {
+			return err
+		}
+		return errors.New("Placeholder resource doesn't exists")
 	}
-	return errors.New("Placeholder resource doesn't exists")
+	return nil
 }
--- a/pkg/kapp/clusterapply/exists_change.go
+++ b/pkg/kapp/clusterapply/exists_change.go
@@ -28,7 +28,7 @@ type ExistsStrategy struct {
 func (e ExistsStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
 
 func (e ExistsStrategy) Apply() error {
-	exists, err := e.identifiedResources.Exists(e.res, ctlres.ExistsOpts{})
+	_, exists, err := e.identifiedResources.Exists(e.res, ctlres.ExistsOpts{})
 	if !exists {
 		if err != nil {
 			return err
--- a/pkg/kapp/diff/change.go
+++ b/pkg/kapp/diff/change.go
@@ -82,7 +82,7 @@ func (d *ChangeImpl) AppliedResource() ctlres.Resource  { return d.appliedRes }
 
 func (d *ChangeImpl) Op() ChangeOp {
 	if d.existingRes == nil {
-		if isAdoptedResource(d.newRes) {
+		if hasPlaceholderAnnotation(d.newRes) {
 			return ChangeOpExists
 		}
 		return ChangeOpAdd
@@ -93,7 +93,7 @@ func (d *ChangeImpl) Op() ChangeOp {
 	}
 
 	if d.ConfigurableTextDiff().Full().HasChanges() {
-		if isAdoptedResource(d.newRes) {
+		if hasPlaceholderAnnotation(d.newRes) {
 			return ChangeOpExists
 		}
 		return ChangeOpUpdate
@@ -160,7 +160,7 @@ func (d *ChangeImpl) calculateOpsDiff() OpsDiff {
 	return OpsDiff(patch.Diff{Left: existingObj, Right: newObj}.Calculate())
 }
 
-func isAdoptedResource(res ctlres.Resource) bool {
+func hasPlaceholderAnnotation(res ctlres.Resource) bool {
 	_, hasPlaceholderAnnotation := res.Annotations()[placeHolderAnnKey]
 	return hasPlaceholderAnnotation
 }
--- a/pkg/kapp/diff/change.go
+++ b/pkg/kapp/diff/change.go
@@ -20,7 +20,7 @@ const (
 )
 
 const (
-	placeHolderAnnKey = "kapp.k14s.io/placeholder" //value is ignored
+	ExternalResourceAnnKey = "kapp.k14s.io/external-resource" //value is ignored
 )
 
 type Change interface {
@@ -82,7 +82,7 @@ func (d *ChangeImpl) AppliedResource() ctlres.Resource  { return d.appliedRes }
 
 func (d *ChangeImpl) Op() ChangeOp {
 	if d.existingRes == nil {
-		if hasPlaceholderAnnotation(d.newRes) {
+		if hasExternalResourceAnnotation(d.newRes) {
 			return ChangeOpExists
 		}
 		return ChangeOpAdd
@@ -93,7 +93,7 @@ func (d *ChangeImpl) Op() ChangeOp {
 	}
 
 	if d.ConfigurableTextDiff().Full().HasChanges() {
-		if hasPlaceholderAnnotation(d.newRes) {
+		if hasExternalResourceAnnotation(d.newRes) {
 			return ChangeOpExists
 		}
 		return ChangeOpUpdate
@@ -160,7 +160,7 @@ func (d *ChangeImpl) calculateOpsDiff() OpsDiff {
 	return OpsDiff(patch.Diff{Left: existingObj, Right: newObj}.Calculate())
 }
 
-func hasPlaceholderAnnotation(res ctlres.Resource) bool {
-	_, hasPlaceholderAnnotation := res.Annotations()[placeHolderAnnKey]
-	return hasPlaceholderAnnotation
+func hasExternalResourceAnnotation(res ctlres.Resource) bool {
+	_, hasExternalResourceAnnotation := res.Annotations()[ExternalResourceAnnKey]
+	return hasExternalResourceAnnotation
 }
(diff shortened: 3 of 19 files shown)


pkg/kapp/clusterapply/cluster_change_set.go
```diff
@@ -152,6 +152,9 @@ func (c wrappedClusterChange) Op() ctldgraph.ActualChangeOp {
 	case ClusterChangeApplyOpNoop:
 		return ctldgraph.ActualChangeOpNoop
 
+	case ClusterChangeApplyOpExists:
+		return ctldgraph.ActualChangeOpExists
```
cppforlife: should just return ActualChangeOpUpsert here.

pkg/kapp/clusterapply/exists_change.go
```diff
@@ -0,0 +1,36 @@
...
+
+type WaitStrategy struct {
+	res ctlres.Resource
+	d   ExistsChange
+}
+
+func (c WaitStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
+
+func (c WaitStrategy) Apply() error {
+	exists, _ := c.d.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
```
cppforlife: if we need access to identifiedResources, that's what we should initialize object with instead of passing down ExistsChange

pkg/kapp/diffgraph/change.go
```diff
@@ -30,6 +30,7 @@ const (
 	ActualChangeOpUpsert ActualChangeOp = "upsert"
 	ActualChangeOpDelete ActualChangeOp = "delete"
 	ActualChangeOpNoop   ActualChangeOp = "noop"
+	ActualChangeOpExists ActualChangeOp = "exists"
```
cppforlife: no need for this.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,131 @@
...
+			t.Fatalf("Expected output to be >>%s<<, but got >>%s<<\n", expectedOutput, out)
+		}
+	})
+}
+
+func deployExternalResource(env Env, clientset *kubernetes.Clientset, name string) {
+	time.Sleep(2 * time.Second)
+	metaData := metav1.ObjectMeta{Name: name}
+	data := map[string]string{}
+	data["key1"] = "value1"
```
cppforlife: inline into ConfigMap.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,131 @@
...
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		if expectedOutput != out {
+			t.Fatalf("Expected output to be >>%s<<, but got >>%s<<\n", expectedOutput, out)
+		}
+	})
+}
+
+func deployExternalResource(env Env, clientset *kubernetes.Clientset, name string) {
+	time.Sleep(2 * time.Second)
```
cppforlife: lets sleep outside of this function, in the test inside a go routine.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,131 @@
...
+	metaData := metav1.ObjectMeta{Name: name}
+	data := map[string]string{}
+	data["key1"] = "value1"
+	configMap := v1.ConfigMap{
+		ObjectMeta: metaData,
+		Data:       data,
+	}
+	_, err := clientset.CoreV1().ConfigMaps(env.Namespace).Create(context.TODO(), &configMap, metav1.CreateOptions{})
+	if err != nil {
+		panic(err)
```
cppforlife: use t.Fatalf to fail, or better yet, use require.NoError()

test/e2e/exists_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func deleteExternalResource(env Env, clientset *kubernetes.Clientset, name string) {
+	err := clientset.CoreV1().ConfigMaps(env.Namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if err != nil {
+		if err.Error() != fmt.Sprintf(`configmaps "%s" not found`, name) {
```
cppforlife: there is errors.IsNotFound() to check for not found error.

wouldnt we want to fail on all errors?

pkg/kapp/diff/change.go
```diff
@@ -16,6 +16,11 @@ const (
 	ChangeOpDelete ChangeOp = "delete"
 	ChangeOpUpdate ChangeOp = "update"
 	ChangeOpKeep   ChangeOp = "keep" // unchanged
+	ChangeOpExists ChangeOp = "exists"
+)
+
+const (
+	placeHolderAnnKey = "kapp.k14s.io/placeholder" //value is ignored
```
cppforlife: kapp.k14s.io/placeholder -> kapp.k14s.io/external-resource

pkg/kapp/clusterapply/exists_change.go
```diff
@@ -0,0 +1,36 @@
...
+	change              ctldiff.Change
+	identifiedResources ctlres.IdentifiedResources
+}
+
+func (c ExistsChange) ApplyStrategy() (ApplyStrategy, error) {
+	res := c.change.NewResource()
+	return WaitStrategy{res, c}, nil
+}
+
+type WaitStrategy struct {
```
cppforlife: name is not really representative of what's happening.

pkg/kapp/clusterapply/exists_change.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (c WaitStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
+
+func (c WaitStrategy) Apply() error {
+	exists, _ := c.d.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
+	if exists {
+		return nil
+	}
+	return errors.New("Resource doesn't exists")
```
cppforlife: we need to find a better way to check on this error. asserting on text isnt great.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,131 @@
...
+	configOverrides := &clientcmd.ConfigOverrides{}
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(loadingRules, configOverrides)
+	config, err := kubeConfig.ClientConfig()
+	if err != nil {
+		t.Fatalf("Error getting kube config, %v", err)
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		t.Fatalf("Error creating client, %v", err)
+	}
```
praveenrewar: Thinking of moving this to another function as we are doing the same thing in `warnings_test.go` as well.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,131 @@
...
+<replaced>: ok: reconcile configmap/external (v1) namespace: kapp-test
+<replaced>: ---- applying complete [3/3 done] ----
+<replaced>: ---- waiting complete [3/3 done] ----
+
+Succeeded`
+
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		if expectedOutput != out {
+			t.Fatalf("Expected output to be >>%s<<, but got >>%s<<\n", expectedOutput, out)
```
praveenrewar: Should we also assert on the inspect output, delete output and the data of the external config map?

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Should we update this error message to something like 
```suggestion
	return errors.New("Placeholder resource doesn't exists")
```

pkg/kapp/resources/resources.go
```diff
@@ -557,6 +557,8 @@ func IsResourceChangeBlockedErr(err error) bool {
 		return true
 	case strings.Contains(errMsg, "Internal error occurred: failed calling webhook"):
 		return true
+	case strings.Contains(errMsg, "Resource doesn't exists"):
```
100mik: `isResourceChangeBlockedErr()` seems to exclusively handle cases where the change is blocked by admission web hooks?
Maybe we could have a separate function like `isWaitingForPlaceholderErr()` which handles this specific case?

praveenrewar: Makes sense. Will do.

test/e2e/exists_test.go
praveenrewar: Done.

test/e2e/exists_test.go
praveenrewar: Added an assert on inspect.

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Done

pkg/kapp/resources/resources.go
praveenrewar: Done.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,137 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+func TestExistsOpWithPlaceholderAnn(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	clientset, err := getKubernetesClientset()
```
100mik: Should we deploy the external resource using `kapp` with a different app name? I think that will make it simpler. Thoughts?

test/e2e/exists_test.go
```diff
@@ -0,0 +1,137 @@
...
+	logger.Section("inspecting app", func() {
+		out, _ := kapp.RunWithOpts([]string{"inspect", "-a", name},
+			RunOpts{StdinReader: strings.NewReader(app)})
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+
+		expectedOutput := `
+Resources in app 'app'
+
+Namespace  Name     Kind       Owner  Conds.  Rs  Ri  Age  $
+(cluster)  kapp-ns  Namespace  kapp   -       ok  -   2s  $
```
100mik: We could use rows instead but I think it makes sense to keep it consistent as we previously matched the output as well.

pkg/kapp/clusterapply/exists_change.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (c ExistsChange) ApplyStrategy() (ApplyStrategy, error) {
+	res := c.change.NewResource()
+	return WaitStrategy{res, c}, nil
+}
+
+type WaitStrategy struct {
+	res ctlres.Resource
+	d   ExistsChange
```
100mik: Why not name it `change` ?

```suggestion
	change   ExistsChange
```

test/e2e/exists_test.go
praveenrewar: Deploying it with kapp might cause ownership issues. Even if it doesn't cause ownership issues, it would still be shown as a resource whose owner is kapp and I think that the owner should be cluster.

pkg/kapp/clusterapply/exists_change.go
praveenrewar: We already have a `change` variable in `ExistsChange`. But it makes moree sense to use `e` than using `d`. I will update it.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2021 VMware, Inc.
```
sethiyash: will it be better to rename this file as `exists_op_test.go` ?

`exists_test.go` name I believe is not  communicating that we are trying to test exists operation here.

test/e2e/exists_test.go
```diff
@@ -0,0 +1,137 @@
...
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		if expectedOutput != out {
+			t.Fatalf("Expected output to be >>%s<<, but got >>%s<<\n", expectedOutput, out)
+		}
+	})
+
+	logger.Section("inspecting app", func() {
+		out, _ := kapp.RunWithOpts([]string{"inspect", "-a", name},
+			RunOpts{StdinReader: strings.NewReader(app)})
```
sethiyash:
```suggestion
			RunOpts{})
```
nit: you can ignore passing app yaml here as it is `inspect` command.

test/e2e/exists_test.go
praveenrewar: Makes sense. Will do.

test/e2e/exists_test.go
praveenrewar: Yep, forgot to remove it. Good catch :)

test/e2e/exists_test.go
100mik: That does make sense

100mik (approved):

pkg/kapp/clusterapply/exists_change.go
```diff
@@ -0,0 +1,36 @@
...
+
+type WaitStrategy struct {
+	res ctlres.Resource
+	e   ExistsChange
+}
+
+func (c WaitStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
+
+func (c WaitStrategy) Apply() error {
+	exists, _ := c.e.identifiedResources.Exists(c.res, ctlres.ExistsOpts{})
```
rohitagg2020: I think we should catch the error and only if the error contains "Resource doesnt exist", then we send errors.New("Placeholder Resource doesn't exist"). Otherwise, we should send back the original error.

praveenrewar: In most of the scenarios [here](https://github.com/vmware-tanzu/carvel-kapp/blob/5ce3839d0046d649f4bf915ba5d6fb9ed7f29168/pkg/kapp/resources/resources.go#L372), we are not returning any error. So, should I check for `nil` instead?

pkg/kapp/diff/change.go
```diff
@@ -148,3 +159,8 @@ func (d *ChangeImpl) calculateOpsDiff() OpsDiff {
 
 	return OpsDiff(patch.Diff{Left: existingObj, Right: newObj}.Calculate())
 }
+
+func isAdoptedResource(res ctlres.Resource) bool {
```
renuy: why is this called isAdoptedResource?

pkg/kapp/resources/resources.go
```diff
@@ -564,6 +564,16 @@ func IsResourceChangeBlockedErr(err error) bool {
 	}
 }
 
+func IsPlaceholderResourceExistenceErr(err error) bool {
+	errMsg := err.Error()
+	switch {
+	case strings.Contains(errMsg, "Placeholder resource doesn't exists"):
```
renuy: since this is used twice. is there a case to extract this and make an error's constant and use the constant here?

sethiyash (approved):

pkg/kapp/diff/change.go
praveenrewar: I was thinking of using `isExternalResource`, but this scenario doesn't apply to all the external resources. Maybe we can use something like `isPlaceholderResource`. Thoughts?

pkg/kapp/resources/resources.go
praveenrewar: I just thought that this error won't be used anywhere else apart from these 2 scenarios, so didn't use a constant. But I guess there's no harm in doing that. Will make the change.

pkg/kapp/diff/change.go
praveenrewar: Or we can follow the existing terminology and call it `hasPlaceholderAnnotation`?

pkg/kapp/clusterapply/exists_change.go
rohitagg2020: IMO, it shouldnt rely on what underlying function is considering as error today.. If underlying function tomorrow starts sending more error messages, then it becomes regression issue.

pkg/kapp/resources/resources.go
100mik: I would suggest something like `IsPlaceholderError()` in the same file as the error itself. And maybe that can be consumed by this function? Similar to how api-machinery exposes functions like `IsStatusInternalError()`. A constant certainly is better than maintaining two versions here I would say

pkg/kapp/clusterapply/exists_change.go
100mik: Something like 
```golang
if err != nil {
    return err
}
return errors.New("Placeholder resource doesn't exists")
```
Will cover all scenarios in my opinion. Thoughts?

praveenrewar: @rohitagg2020 The underlying functions send nil when the resource doesn't exists and value of exists as false. So, if we are checking for an error message containing "Resource doesn't exists", then we might not be retrying in any scenario.

praveenrewar: @100mik Yep, that's what I am thinking.

pkg/kapp/resources/resources.go
praveenrewar: I had thought of creating a function called `IsPlaceolderError()` and consuming it directly while checking for retries. 
Currently we are importing `resources` package inside `clusterapply`, so we won't be able to import `clusterapply` inside `resources`.

pkg/kapp/diff/change.go
rohitagg2020: hasPlaceholderAnnotation LGTM

pkg/kapp/resources/resources.go
100mik: Ah a circular dependency. Did not think of that. Makes sense

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Done.

test/e2e/exists_test.go
praveenrewar: > there is errors.IsNotFound() to check for not found error.

I was looking for something similar. Thanks :)
> wouldnt we want to fail on all errors?

This is to handle cleanUp(). The resource wouldn’t be present initially but we would still be trying to delete it.

pkg/kapp/diff/change.go
praveenrewar: I will try to think of a better annotation.

pkg/kapp/clusterapply/exists_change.go
praveenrewar: I am thinking of using a custom error, then we won't have to check the error message, we can just retry based on error type. Does that sound good?

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Should we just use `ExistsStrategy`?

pkg/kapp/clusterapply/cluster_change_set.go
praveenrewar: Done

pkg/kapp/diffgraph/change.go
praveenrewar: Removed

test/e2e/exists_test.go
praveenrewar: Done.

test/e2e/exists_test.go
praveenrewar: Used a channel to pass the error from goroutine and the user t.Fatalf to fail.

pkg/kapp/diff/change.go
praveenrewar: Used ` kapp.k14s.io/external-resource` for now. I will update it later on (if needed)

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Updated with `ExistsStrategy`

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Used a custom error for now.

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Done.

pkg/kapp/clusterapply/exists_change.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (e ExistsStrategy) Op() ClusterChangeApplyStrategyOp { return "" }
+
+func (e ExistsStrategy) Apply() error {
+	exists, err := e.identifiedResources.Exists(e.res, ctlres.ExistsOpts{})
+	if !exists {
+		if err != nil {
+			return err
+		}
+		return &ExistsChangeError{}
```
cppforlife: any reason why we need to make it a pointer?

pkg/kapp/diff/change.go
```diff
@@ -16,6 +16,11 @@ const (
 	ChangeOpDelete ChangeOp = "delete"
 	ChangeOpUpdate ChangeOp = "update"
 	ChangeOpKeep   ChangeOp = "keep" // unchanged
+	ChangeOpExists ChangeOp = "exists"
+)
+
+const (
+	ExternalResourceAnnKey = "kapp.k14s.io/external-resource" //value is ignored
```
cppforlife: nit: space `//`

pkg/kapp/diff/change.go
```diff
@@ -77,6 +82,9 @@ func (d *ChangeImpl) AppliedResource() ctlres.Resource  { return d.appliedRes }
 
 func (d *ChangeImpl) Op() ChangeOp {
```
cppforlife, line 85: can this method be rearranged to only have one call to hasExternalResourceAnnotation?

test/e2e/exists_op_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func deleteExternalResource(clientset *kubernetes.Clientset, name string, namespace string) error {
+	err := clientset.CoreV1().ConfigMaps(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+	}
+	return err
```
cppforlife: can we just use our existing Kubectl class?

cppforlife: take a look at https://github.com/vmware-tanzu/carvel-kapp/blob/develop/test/e2e/cluster_resource.go. should this functionality be on something like that?

pkg/kapp/clusterapply/exists_change.go
praveenrewar: I had initially planned to add the resource description in the error as well, but then didn't do it as the description is already available in the logs. Will remove the pointer.

pkg/kapp/diff/change.go
praveenrewar: I will try to do that without compromising readability.

test/e2e/exists_op_test.go
praveenrewar: I will take a look at it :)

praveenrewar: Since we are just creating an empty config map and deleting it, using our existing Kubectl class makes more sense. I will make the changes.

pkg/kapp/clusterapply/exists_change.go
praveenrewar: Done

pkg/kapp/diff/change.go
praveenrewar: Done

pkg/kapp/diff/change.go
praveenrewar: Done

test/e2e/exists_op_test.go
praveenrewar: Done

pkg/kapp/resources/resources.go
```diff
@@ -557,6 +557,8 @@ func IsResourceChangeBlockedErr(err error) bool {
 		return true
 	case strings.Contains(errMsg, "Internal error occurred: failed calling webhook"):
 		return true
+	case strings.Contains(errMsg, `cluster: admission webhook "validation.gatekeeper.sh" denied the request`):
```
praveenrewar: Should we make it more generic by using regex for `"validation.gatekeeper.sh"`.
Something like this `cluster: admission webhook (.+) denied the request`

praveenrewar: TBH, I am not really happy with either of them and I am open for suggestions.

praveenrewar: Error message that we are checking for
```
kapp: Error: Applying create k8srequiredlabels/ns-must-have-gk (constraints.gatekeeper.sh/v1beta1) cluster:
  Creating resource k8srequiredlabels/ns-must-have-gk (constraints.gatekeeper.sh/v1beta1) cluster: admission webhook "validation.gatekeeper.sh" denied the request:
    Constraint kind K8sRequiredLabels is not recognized (reason: Constraint kind K8sRequiredLabels is not recognized)
```

pkg/kapp/diff/change.go
```diff
@@ -148,3 +159,8 @@ func (d *ChangeImpl) calculateOpsDiff() OpsDiff {
 
 	return OpsDiff(patch.Diff{Left: existingObj, Right: newObj}.Calculate())
 }
+
+func hasExternalResourceAnnotation(res ctlres.Resource) bool {
```
cppforlife: nit: can this be inlined (now that it's being used only in one place)?

praveenrewar: Sure, will make it inline.

praveenrewar:
Is it normal for PRs to show all the commits from develop after rebasing or did I mess up somewhere 😕

pkg/kapp/diff/change.go
praveenrewar: Done

pkg/kapp/resources/resources.go
praveenrewar: Removed retries for now.

cppforlife:
> Is it normal for PRs to show all the commits from develop after rebasing or did I mess up somewhere

rebase didnt quite work im guessing. try again? (git fetch origin && git rebase origin/develop)

praveenrewar:
> rebase didnt quite work im guessing. try again?

Something did go wrong, maybe I forgot to resolve one of the conflicts after rebasing. I will try to reset and then rebase.

Update: I forgot to force push after rebasing and did pull and push which had caused the issue.

pkg/kapp/diff/change.go
```diff
@@ -16,6 +16,11 @@ const (
 	ChangeOpDelete ChangeOp = "delete"
 	ChangeOpUpdate ChangeOp = "update"
 	ChangeOpKeep   ChangeOp = "keep" // unchanged
+	ChangeOpExists ChangeOp = "exists"
+)
+
+const (
+	ExternalResourceAnnKey = "kapp.k14s.io/external-resource" // Value is ignored
```
cppforlife: should we call this something like `kapp.k14s.io/unmanaged: ""` (or is external-resource better name?) so that we could in future also support `does-not-exist` change (aka ensure resource is deleted)?

praveenrewar: Both `unmanaged` and `external-resource` seem good to me, although I think we do have a certain flow defined for external resources and having the annotation external-resource could be a bit misleading as the behaviour of resources with this annotation would be different than other external resources. So, I think we can go ahead with `unmanaged`.

praveenrewar: > so that we could in future also support does-not-exist change (aka ensure resource is deleted)

Interesting thought, but I am curious as to how we would decide which change (`exists` or `does-not-exists`) to have based on this annotation.

praveenrewar: Used `kapp.k14s.io/unmanaged: ""` for now.

praveenrewar:
@cppforlife  Should we add `ClusterChangeApplyOpExists` to the list of visibleApplyOps [here](https://github.com/vmware-tanzu/carvel-kapp/blob/63eacfb05e8a0cdf7fc8ef1380245d4aa5adf3a7/pkg/kapp/clusterapply/changes_view.go#L231)?
(I think it we should, otherwise the number of resources in the table and summary don't match.) 
Adding screenshots for reference.

Screenshot 2021-11-10 at 3 20 11 PM
Screenshot 2021-11-10 at 3 21 11 PM

However, we might wanna hide `0 exists` when there are no resources with exists operation.

Update: Planning to create a separate PR for this as we would need to update a lot of test cases for it.

pkg/kapp/clusterapply/cluster_change_set.go
praveenrewar: Just thinking if we should keep this to show `(exists)` besides unmanaged resources in diff-ui-alpha, or `(upsert)` is fine?
cc @cppforlife

pkg/kapp/diff/change.go
```diff
@@ -148,3 +159,8 @@ func (d *ChangeImpl) calculateOpsDiff() OpsDiff {
 
 	return OpsDiff(patch.Diff{Left: existingObj, Right: newObj}.Calculate())
 }
+
+func hasExistsAnnotation(res ctlres.Resource) bool {
```
cppforlife: should be on `func (d *ChangeImpl)`

praveenrewar: Done

PR #327, opened 2021-10-03
add --dangerous-scope-to-fallback-allowed-namespaces

cppforlife:
this is useful to force kapp only look within a fallback-allowed-namespace -- and in case there is only 1 ns (currently that's the case), it eliminates a lot list requests when running under a service account that has access to list all namespaces, but does not have access to list resources within those namespaces.

### DIFF
--- a/pkg/kapp/resources/resources.go
+++ b/pkg/kapp/resources/resources.go
@@ -524,8 +524,10 @@ func (c *ResourcesImpl) assumedAllowedNamespaces() ([]string, error) {
 				return c.fallbackAllowedNamespaces, nil
 			}
 		}
-		return nil, fmt.Errorf("fetching all namespaces: %s", err)
+		return nil, fmt.Errorf("Fetching all namespaces: %s", err)
 	}
+
+	c.logger.Info("Falled back to checking each namespace separately (much slower)")
 
 	var nsNames []string
 
--- a/pkg/kapp/resources/resources.go
+++ b/pkg/kapp/resources/resources.go
@@ -53,11 +53,11 @@ type ExistsOpts struct {
 }
 
 type ResourcesImpl struct {
-	resourceTypes             ResourceTypes
-	coreClient                kubernetes.Interface
-	dynamicClient             dynamic.Interface
-	mutedDynamicClient        dynamic.Interface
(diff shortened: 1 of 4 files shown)


pkg/kapp/resources/resources.go
```diff
@@ -520,13 +539,15 @@ func (c *ResourcesImpl) assumedAllowedNamespaces() ([]string, error) {
...
-				return c.fallbackAllowedNamespaces, nil
+			if len(c.opts.FallbackAllowedNamespaces) > 0 {
+				return c.opts.FallbackAllowedNamespaces, nil
 			}
 		}
-		return nil, fmt.Errorf("fetching all namespaces: %s", err)
+		return nil, fmt.Errorf("Fetching all namespaces: %s", err)
 	}
 
+	c.logger.Info("Falled back to checking each namespace separately (much slower)")
```
100mik: nit, I think it should be 
```suggestion
	c.logger.Info("Fell back to checking each namespace separately (much slower)")
```
(minor grammar nit) or maybe `Falling back to.....`

cppforlife: used Falling back to.....

cppforlife:
included in 0.41.0

renuy:
available as part of v0.41.0

redbaron:
why it is not a default? I need to do more tests, but looks like `kapp delete --namespace abc --app 'label:x=y'`    is not actually scoped to a namespace `abc`.

praveenrewar:
Hi @redbaron! This becomes the default behaviour when you have scoped permisions (no permission to list namespaces).

redbaron:
If I have a cluster wide permissions, I'd still expect `kapp delete --namespace abc` to act _only_ on abc namespace, not anywhere else.

praveenrewar:
The `--namespace` flag is actually used to store the metadata configmap, the resources present in the app could be present in other namespaces as well. Right now we only use the namespace provided in `namespace` flag as a fallback, we are working on including all the resource namespaces as fallback namespaces. But even then there are chances that all the resource namespaces are not included in some scenarios (failure to save/update the metadata configmap), that's why this is a fallback approach rather than the direct one.

PR #337, opened 2021-10-06
allow long change group names

cppforlife:


renuy:
@enj , this fix is part of release v0.42.0

PR #345, opened 2021-10-07
Add Resource to the Exists function signature

### LINKED ISSUE #344: While checking for resource existence with "Exist" function, exist func should also return the resource.
## Value Addition
Today, in the `resources.go` file, both the `Resources` interface and `ResourcesImpl` struct, has the following method definition for Exists method `Exists(Resource, ExistsOpts) (bool, error)`. It doesn't return actual Resource object. Thus, if the caller of the Exists method has to act on the Resource Obj, he has to do another `Get()` call to get the latest resource. By adding the resource obj among the objects returned by Exists func, we can gain some performance.

## Code Changes:

- Navigate to the resources.go file.
- Change the `Resources` Interface `Exists(Resource, ExistsOpts) (bool, error)`  to `Exists(Resource, ExistsOpts) (Resource, bool, error)`. 
- There are two implementation of this interface:
    - `FakeResources` in `identified_resources_list_test.go` file
    As there is no method body for this, just changing the method signature will be enough here.
    - `ResourcesImpl` in `resources.go` file
    Here, we need to change both method signature as well as method body. In method body, we need to create the Resource object similar to like we do in `Get` method of resources.go file. Keep in mind that Retry functionality has to be retained.
- This method is consumed directly in `identified_resources.go` file. We have to change the method signature of this function to return Resource Obj as well.
   https://github.com/vmware-tanzu/carvel-kapp/blob/b5d6a0aa7c6bbef9857b472ca70b8a8c38a70d1a/pkg/kapp/resources/identified_resources.go#L101
- Now, we have to change the places from where `Exists` method of `identified_resources.go` is called. Those places are:
   - In add_or_update_change file
   https://github.com/vmware-tanzu/carvel-kapp/blob/b5d6a0aa7c6bbef9857b472ca70b8a8c38a70d1a/pkg/kapp/clusterapply/add_or_update_change.go#L102
   - In delete_change file
   https://github.com/vmware-tanzu/carvel-kapp/blob/b5d6a0aa7c6bbef9857b472ca70b8a8c38a70d1a/pkg/kapp/clusterapply/delete_change.go#L63
   - In labeled_resources.go file
   https://github.com/vmware-tanzu/carvel-kapp/blob/b5d6a0aa7c6bbef9857b472ca70b8a8c38a70d1a/pkg/kapp/resources/labeled_resources.go#L222 
   In labeled_resources.go file, immediately after this, we check if it exist and then make another call to get the Resource. Now, we can remove that and put this returned Resource Obj over the channel `resCh`
   https://github.com/vmware-tanzu/carvel-kapp/blob/b5d6a0aa7c6bbef9857b472ca70b8a8c38a70d1a/pkg/kapp/resources/labeled_resources.go#L228-L234

## Acceptance Criteria
As these methods are consumed internally, acceptance criteria for this will be to:
-  Adopt the changed method signature at all the places it is called from( as mentioned above).
-  No compilation error. To compile, you can navigate to the root directory i.e. `carvel-kapp` and run this script `hack/build.sh`.

shubham14bajpai:
Fixes: #344 

Signed-off-by: shubham <shubham14bajpai@gmail.com>

### DIFF
--- a/pkg/kapp/resources/identified_resources_list_test.go
+++ b/pkg/kapp/resources/identified_resources_list_test.go
@@ -56,9 +56,11 @@ metadata:
 
 	return []ctlres.Resource{antreaRes, deploymentRes}, nil
 }
-func (r *FakeResources) Delete(ctlres.Resource) error                            { return nil }
-func (r *FakeResources) Exists(ctlres.Resource, ctlres.ExistsOpts) (bool, error) { return true, nil }
-func (r *FakeResources) Get(ctlres.Resource) (ctlres.Resource, error)            { return nil, nil }
+func (r *FakeResources) Delete(ctlres.Resource) error { return nil }
+func (r *FakeResources) Exists(ctlres.Resource, ctlres.ExistsOpts) (ctlres.Resource, bool, error) {
+	return nil, true, nil
+}
+func (r *FakeResources) Get(ctlres.Resource) (ctlres.Resource, error) { return nil, nil }
 func (r *FakeResources) Patch(ctlres.Resource, types.PatchType, []byte) (ctlres.Resource, error) {
 	return nil, nil
 }
--- a/pkg/kapp/resources/resources.go
+++ b/pkg/kapp/resources/resources.go
@@ -41,7 +41,7 @@ const (
 type Resources interface {
 	All([]ResourceType, AllOpts) ([]Resource, error)
 	Delete(Resource) error
-	Exists(Resource, ExistsOpts) (bool, error)
+	Exists(Resource, ExistsOpts) (Resource, bool, error)
 	Get(Resource) (Resource, error)
 	Patch(Resource, types.PatchType, []byte) (Resource, error)
 	Update(Resource) (Resource, error)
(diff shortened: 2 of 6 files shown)


100mik:
hey @shubham14bajpai, could you sign the CLA?

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -56,9 +56,11 @@ metadata:
 
 	return []ctlres.Resource{antreaRes, deploymentRes}, nil
 }
-func (r *FakeResources) Delete(ctlres.Resource) error                            { return nil }
-func (r *FakeResources) Exists(ctlres.Resource, ctlres.ExistsOpts) (bool, error) { return true, nil }
-func (r *FakeResources) Get(ctlres.Resource) (ctlres.Resource, error)            { return nil, nil }
+func (r *FakeResources) Delete(ctlres.Resource) error { return nil }
```
100mik, line 59: Looks like, `go fmt` was not run for this part of the code. Running `/hack/build.sh` should do that. Maybe you could [set up the dev environment ](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/docs/dev.md)and run `/hack/test-all.sh` once as well before committing changes?

pkg/kapp/resources/resources.go
```diff
@@ -388,23 +388,24 @@ func (c *ResourcesImpl) Get(resource Resource) (Resource, error) {
...
 		if _, ok := err.(ResourceTypesUnknownTypeErr); ok {
-			return false, nil
+			return nil, false, nil
 		}
-		return false, err
+		return nil, false, err
 	}
 
 	var found bool
+	var resObj *ResourceImpl
```
100mik: nit 
`var resObj * ResourceIml`           ==>           `var clusterRes Resource`

Using the `Resource` type instead of `ResourceImpl` makes more sense imo?

pkg/kapp/resources/identified_resources_list_test.go
100mik: Tests seem to be passing! Just ensure that go fmt is run once

pkg/kapp/resources/resources.go
shubham14bajpai: Thanks. Yes that makes more sense.

pkg/kapp/resources/identified_resources_list_test.go
shubham14bajpai: I tried that, all the tests passed but still no change in the file. I even tried running `go fmt` on that particular file.

pkg/kapp/resources/resources.go
rohitagg2020: Yes, we should rely on interface than implementation

pkg/kapp/resources/identified_resources_list_test.go
100mik: I think we are good then! Just making sure

cppforlife:
@shubham14bajpai thank you for your contribution!

PR #351, opened 2021-10-11
Add verify-no-dirty-files check

100mik:
To mimic `kapp-controller`
Ensures that `go mod vendor; go mod tidy` does not lead to any uncommitted changes

### DIFF
--- a/.github/workflows/test-gh.yml
+++ b/.github/workflows/test-gh.yml
@@ -30,6 +30,8 @@ jobs:
         mkdir -p /tmp/bin
         export PATH=/tmp/bin:$PATH
 
+        ./hack/verify-no-dirty-files.sh
+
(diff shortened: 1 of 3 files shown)


praveenrewar (approved):
LGTM

.github/workflows/test-gh.yml
```diff
@@ -30,6 +30,8 @@ jobs:
         mkdir -p /tmp/bin
         export PATH=/tmp/bin:$PATH
 
+        ./hack/verify-no-dirty-files.sh
```
cppforlife, line 33: what is the reasoning for checking for dirty files after a check out? why would they be dirty at this point? i would have imagined we want to do this after building binaries.

100mik: The script itself runs `./hack/build.sh`
This allows the script to work fine by itself too

Issue #349, opened 2021-10-08
Update tests to use testify/require instead of value comparisons and Fatal


praveenrewar:
**Describe the problem/challenge you have**
Currently unit tests and e2e tests are using value comparisons and Fatal which is not very intuitive.

**Describe the solution you'd like**
We should update the tests to use testify/require which will make them more readable and the output would be very descriptive and concise in case the tests are failing.

**Anything else you would like to add:**
This has already been done for diff package, and the [PR](https://github.com/vmware-tanzu/carvel-kapp/pull/347) can be found here.
Update the tests for the following packages
- [x] pkg/kapp/diffgraph and pkg/kapp/matcher - @khan-ajamal
- [x] pkg/kapp/resources - @khan-ajamal
- [x] pkg/kapp/resourcemisc - @ShajithaMohammed
- [x] test/e2e - @ShajithaMohammed 

We are also happy to receive and review Pull Requests if you want to help working on this issue.

khan-ajamal:
Hi @praveenrewar i would like to take this up.

praveenrewar:
@khan-ajamal Sure, thanks for contributing!

rohitagg2020:
@khan-ajamal Please let us know how much you have progressed on the same.

100mik:
Thanks for the PR @khan-ajamal ! We are still accepting PRs for the remaining packages:
 - pkg/kapp/resources
 - pkg/kapp/resourcemisc
 - test/e2e

Feel free to comment on this thread if you would like to contribute to these! The changes expected are similar to ones in the PRs:
 - #354 
 - #347

khan-ajamal:
Hi, @100mik I would love to take on the remaining tasks, will send PRs for them soon.

100mik:
That will be wonderful @khan-ajamal ! Putting your name against `pkg/kapp/resources`. We can tick them off as and when the PRs roll in.
Feel free to make a single PR or separate PRs for each package as per your convenience.
Cheers!

ShajithaMohammed:
Hi @praveenrewar @100mik I would like to take up remaining packages - pkg/kapp/resourcemisc and test/e2e

100mik:
Sure! I will put your name against those.

100mik:
@ShajithaMohammed let's put together a PR for the `pkg/kappresourcemisc` package and we will move onto `test/e2e`!

100mik:
[...]

PR #354, opened 2021-10-12
Update tests to use testify/require

khan-ajamal:
Update all tests in `pkg/kapp/diffgraph` and `pkg/kapp/matcher` to use testify/require instead value comparisons and Fatal() as it's more clean and provides error message in a more intuitive way

Issue #349

### DIFF
--- a/pkg/kapp/diffgraph/change_graph_cf_for_k8s_test.go
+++ b/pkg/kapp/diffgraph/change_graph_cf_for_k8s_test.go
@@ -12,23 +12,18 @@ import (
 	ctlconf "github.com/k14s/kapp/pkg/kapp/config"
 	ctldgraph "github.com/k14s/kapp/pkg/kapp/diffgraph"
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+	"github.com/stretchr/testify/require"
 )
 
 func TestChangeGraphCFForK8sUpsert(t *testing.T) {
 	configYAML, err := ioutil.ReadFile("assets/cf-for-k8s.yml")
-	if err != nil {
-		t.Fatalf("Reading cf-for-k8s asset: %s", err)
-	}
+	require.NoErrorf(t, err, "Reading cf-for-k8s asset")
 
 	configRs, err := ctlres.NewFileResource(ctlres.NewBytesSource([]byte(configYAML))).Resources()
-	if err != nil {
-		t.Fatalf("Error parsing resources: %s", err)
-	}
+	require.NoErrorf(t, err, "Parsing resources")
 
 	rs, conf, err := ctlconf.NewConfFromResourcesWithDefaults(configRs)
-	if err != nil {
-		t.Fatalf("Error parsing conf defaults: %s", err)
-	}
+	require.NoErrorf(t, err, "Parsing conf defaults")
 
 	opts := buildGraphOpts{
 		resources:           rs,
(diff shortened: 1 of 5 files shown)


pkg/kapp/diffgraph/change_graph_cf_for_k8s_test.go
```diff
@@ -12,23 +12,18 @@ import (
...
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+	"github.com/stretchr/testify/require"
 )
 
 func TestChangeGraphCFForK8sUpsert(t *testing.T) {
 	configYAML, err := ioutil.ReadFile("assets/cf-for-k8s.yml")
-	if err != nil {
-		t.Fatalf("Reading cf-for-k8s asset: %s", err)
-	}
+	require.NoErrorf(t, err, "reading cf-for-k8s asset")
```
praveenrewar: Should we use an uppercase as the starting character for an error message? 
We usually do that for layering purposes, but it makes sense to keep the behaviour consistent.
Note: This applies to other usage of `require.NoErrorf()` as well.

pkg/kapp/diffgraph/change_graph_cf_for_k8s_test.go
```diff
@@ -40,37 +35,25 @@ func TestChangeGraphCFForK8sUpsert(t *testing.T) {
...
-	}
+	require.Less(t, time.Now().Sub(t1), time.Duration(1*time.Second), "graph build took too long")
 
 	output := strings.TrimSpace(graph.PrintLinearizedStr())
 	expectedOutput := strings.TrimSpace(cfForK8sExpectedOutputUpsert)
 
-	if output != expectedOutput {
-		t.Fatalf("Expected output to be >>>%s<<< but was >>>%s<<<", expectedOutput, output)
-	}
+	require.Equal(t, output, expectedOutput)
```
praveenrewar: I think `require.Equal()` expects `expected` output first and then the `actual` output. Changing the order might cause ambiguity in the error message if the tests fail.
Note: This would apply to other usage of `require.Equal` as well.

pkg/kapp/diffgraph/change_graph_test.go
```diff
@@ -351,12 +329,9 @@ metadata:
...
 	_, err := buildChangeGraph(circularDep1YAML, ctldgraph.ActualChangeOpUpsert, t)
-	if err == nil {
-		t.Fatalf("Expected graph to fail building")
-	}
-	if err.Error() != "Detected cycle while ordering changes: [job/job1 () cluster] -> [job/job2 () cluster] -> [job/job1 () cluster] (found repeated: job/job1 () cluster)" {
-		t.Fatalf("Expected to detect cycle: %s", err)
-	}
+	require.Error(t, err, "expected graph to fail building")
+
+	require.EqualError(t, err, "Detected cycle while ordering changes: [job/job1 () cluster] -> [job/job2 () cluster] -> [job/job1 () cluster] (found repeated: job/job1 () cluster)", "expected to detect cycle")
```
praveenrewar: Should we declare the expected error first or use a function to retrieve it to make it more readable?
Something like
```suggestion
	expectedErr := "Detected cycle while ordering changes: [job/job1 () cluster] -> [job/job2 () cluster] -> [job/job1 () cluster] (found repeated: job/job1 () cluster)"
	require.EqualError(t, err, expectedErr, "Expected to detect cycle")
```

pkg/kapp/matcher/string_test.go
```diff
@@ -47,8 +48,5 @@ type stringMatcherExample struct {
 
 func (e stringMatcherExample) Check(t *testing.T) {
 	result := matcher.NewStringMatcher(e.Expected).Matches(e.Actual)
-	if result != e.Result {
-		t.Fatalf("Did not match result: expected=%s actual=%s wanted=%t got=%t",
-			e.Expected, e.Actual, e.Result, result)
-	}
+	require.Equal(t, result, e.Result, "Did not match result: expected=%s actual=%s wanted=%t got=%t", e.Expected, e.Actual, e.Result, result)
```
praveenrewar: I think we can ignore the `e.Result` value and `result` value from this message as that will already be covered by testify/require. We can just use `e.Expected` and `e.Actual` values.

pkg/kapp/diffgraph/change_graph_cf_for_k8s_test.go
khan-ajamal: @praveenrewar changed for all `require.NoErrorf()`

pkg/kapp/diffgraph/change_graph_cf_for_k8s_test.go
khan-ajamal: Good catch @praveenrewar . Thanks

pkg/kapp/diffgraph/change_graph_cf_for_k8s_test.go
```diff
@@ -82,20 +65,14 @@ func TestChangeGraphCFForK8sDelete(t *testing.T) {
...
-	}
+	require.Less(t, time.Now().Sub(t1), time.Duration(1*time.Second), "Graph build took too long")
 
 	output := strings.TrimSpace(graph.PrintLinearizedStr())
 	expectedOutput := strings.TrimSpace(cfForK8sExpectedOutputDelete)
 
-	if output != expectedOutput {
-		t.Fatalf("Expected output to be >>>%s<<< but was >>>%s<<<", expectedOutput, output)
-	}
+	require.Equal(t, output, expectedOutput)
```
praveenrewar: Just one last instance of incorrect order of expected and actual output is remaining.

pkg/kapp/diffgraph/change_graph_test.go
```diff
@@ -351,12 +329,9 @@ metadata:
...
 	_, err := buildChangeGraph(circularDep1YAML, ctldgraph.ActualChangeOpUpsert, t)
-	if err == nil {
-		t.Fatalf("Expected graph to fail building")
-	}
-	if err.Error() != "Detected cycle while ordering changes: [job/job1 () cluster] -> [job/job2 () cluster] -> [job/job1 () cluster] (found repeated: job/job1 () cluster)" {
-		t.Fatalf("Expected to detect cycle: %s", err)
-	}
+	require.Error(t, err, "Expected graph to fail building")
+	expectedErr := "Detected cycle while ordering changes: [job/job1 () cluster] -> [job/job2 () cluster] -> [job/job1 () cluster] (found repeated: job/job1 () cluster)"
+	require.EqualError(t, err, expectedErr, "Expected to detect cycle")
```
praveenrewar: Should we use new lines in between for better readability?

pkg/kapp/diffgraph/change_group_test.go
```diff
@@ -29,7 +29,7 @@ func TestNewChangeGroupFromAnnString(t *testing.T) {
 	for _, name := range names {
 		cg, err := ctldgraph.NewChangeGroupFromAnnString(name)
 		require.NoError(t, err)
-		require.Equal(t, cg.Name, name)
+		require.Equal(t, name, cg.Name)
```
praveenrewar, line 32: Good catch.

praveenrewar (approved):

100mik (approved):
LGTM

cppforlife:
thank you!

PR #355, opened 2021-10-13
Update resources tests to use testify/require

khan-ajamal:
Update all tests in `pkg/kapp/resources` to use `testify/require` instead of value comparisons and `Fatal()` as it's cleaner and provides error messages in a more intuitive way

Issue #349

### DIFF
--- a/pkg/kapp/resources/resource_test.go
+++ b/pkg/kapp/resources/resource_test.go
@@ -4,10 +4,10 @@
 package resources_test
 
 import (
-	"strings"
 	"testing"
 
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+	"github.com/stretchr/testify/require"
 )
 
 func TestCompactBytesLength(t *testing.T) {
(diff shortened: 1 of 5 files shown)


pkg/kapp/resources/resource_test.go
```diff
@@ -40,29 +40,18 @@ metadata:
...
 	resFromCompact, err := ctlres.NewResourceFromBytes([]byte(compactBs))
-	if err != nil {
-		t.Fatalf("Expected to parse compact bytes: %s", err)
-	}
+	require.NoError(t, err, "Expected to parse compact bytes")
 
-	if !resFromFull.Equal(resFromCompact) {
-		t.Fatalf("Expected resources to match: '%s' vs '%s'", fullBs, compactBs)
-	}
+	require.True(t, resFromFull.Equal(resFromCompact), "Expected resources to match")
```
praveenrewar: Should we keep the error message same, i.e. include the fullBs and compactBs comparison, as `require.True()` would only check if the result is true or not and won't print the values.

khan-ajamal: @praveenrewar sure. Added the context required

praveenrewar (approved):

PR #368, opened 2021-11-11
Fix: Remove input flag to trivy-action

neil-hickey:
The current trivy workflow does not work as the `input` flag is specified with `.` results in an error. 

```
2021-11-11T00:07:45.203Z	FATAL	flag provided but not defined: -input
Incorrect Usage: flag provided but not defined: -input
```

The input  flag is not actually needed, and the trivy-action defaults to the current directory anyways.

https://github.com/vmware-tanzu/carvel-kapp/blob/develop/.github/workflows/trivy-scan.yml

Example failure runs:

https://github.com/vmware-tanzu/carvel-kapp/actions/workflows/trivy-scan.yml

### DIFF
--- a/.github/workflows/trivy-scan.yml
+++ b/.github/workflows/trivy-scan.yml
@@ -29,7 +29,6 @@ jobs:
         uses: aquasecurity/trivy-action@master
         with:
           scan-type: 'fs'
-          input: '.'
           ignore-unfixed: true
           format: 'template'
           template: '@/contrib/sarif.tpl'


rohitagg2020:
Hi Neil,

In earlier versions, it was able to scan both the `go.mod` and `kapp` binary file. But now, with latest release, it is scanning only `go.sum` file. Also your observation that, input flag is not supported now is bang on.

Our primary focus is to run the scan on `kapp` binary. And for that, we are planning to switch to the`rootfs`.

neil-hickey:
> Hi Neil,
> 
> In earlier versions, it was able to scan both the `go.mod` and `kapp` binary file. But now, with latest release, it is scanning only `go.sum` file. Also your observation that, input flag is not supported now is bang on.
> 
> Our primary focus is to run the scan on `kapp` binary. And for that, we are planning to switch to the`rootfs`.

hey @rohitagg2020 that makes sense! If you all are working on fixing it / modifying to use the new type I will close out this PR and leave it in your hands 👏

PR #372, opened 2021-11-18
Fixing container logs issue.

### LINKED ISSUE #359: restarted container logs are not picked up in logs -f
**What steps did you take:**
i believe that when container is restarts (pod stays the same) logs are not retrieved as part of `kapp logs -f ...`. i think currently log tailing stops after container exits from the current run.

**Environment:**

- kapp version (use `kapp --version`): latest
- OS (e.g. from `/etc/os-release`): os x
- Kubernetes version (use `kubectl version`): 1.22

---
Vote on this request

rohitagg2020:
Fixing container logs issue, if '-f' option is given, then once the container restarts, it should start outputing the logs again. Issue #359


pkg/kapp/logs/pod_container_log.go
```diff
@@ -44,6 +44,30 @@ func NewPodContainerLog(
...
+	for {
+		linePrefix := ""
+		if len(l.opts.LinePrefix) > 0 {
+			linePrefix = l.opts.LinePrefix + " | "
+		}
+		err := l.StartTail(ui, cancelCh)
+
+		if err == io.EOF {
+			if l.opts.Follow {
+				ui.BeginLinef("Container Restarted\n")
```
cppforlife: we follow same format here e.g. `ui.BeginLinef("%s# container restarting '%s' logs\n", linePrefix, l.tag)`

pkg/kapp/logs/pod_container_log.go
```diff
@@ -44,6 +44,30 @@ func NewPodContainerLog(
...
+				time.Sleep(500 * time.Millisecond)
```
cppforlife: in case container doesnt restart in 500ms, what happens?

pkg/kapp/logs/pod_container_log.go
rohitagg2020: Adopted

pkg/kapp/logs/pod_container_log.go
rohitagg2020: As container got restarted, we wait for 500ms before making a call to the `StartTail` function to obtain the stream. If it starts, then we get the stream and start outputting the data. However, if it doesn't start, then stream will be `nil` and `StartTail` will return nil. But because we are in endless `for` loop, we will continue to make calls to `StartTail` and once the container will be up, we will start streaming the logs eventually.

pkg/kapp/logs/pod_container_log.go
```diff
@@ -140,6 +163,16 @@ func (l PodContainerLog) readyToGetLogs() bool {
 	if err != nil {
 		return false
 	}
+	containerStatuses := pod.Status.ContainerStatuses
+	for _, containerStatus := range containerStatuses {
+		//We should focus only on the status of this particular container.
+		if l.container != containerStatus.Name {
+			continue
+		}
+		if containerStatus.State.Running != nil {
```
rohitagg2020, line 173: Please provide your comments on this condition `containerStatus.State.Running != nil`. We need to make a decision on when to start streaming the logs. If we dont check for any condition, then it stream the older logs until the container is up leading to bad user experience and repeating the last few lines of previous container.

pkg/kapp/logs/pod_container_log.go
```diff
@@ -44,6 +44,30 @@ func NewPodContainerLog(
 }
 
 func (l PodContainerLog) Tail(ui ui.UI, cancelCh chan struct{}) error {
+	for {
+		linePrefix := ""
+		if len(l.opts.LinePrefix) > 0 {
+			linePrefix = l.opts.LinePrefix + " | "
+		}
```
praveenrewar: I think we can keep this outside of the for loop.

pkg/kapp/logs/pod_container_log.go
```diff
@@ -44,6 +44,30 @@ func NewPodContainerLog(
...
+		err := l.StartTail(ui, cancelCh)
+
+		if err == io.EOF {
+			if l.opts.Follow {
+				ui.BeginLinef("%s# container restarting '%s' logs\n", linePrefix, l.tag)
+				time.Sleep(500 * time.Millisecond)
+				continue
+			} else {
```
praveenrewar, line 60: We don't need an `else` after `continue`.

pkg/kapp/logs/pod_container_log.go
```diff
@@ -76,8 +100,7 @@ func (l PodContainerLog) Tail(ui ui.UI, cancelCh chan struct{}) error {
 		line, err := reader.ReadBytes('\n')
 		if err != nil {
 			if err == io.EOF {
-				ui.BeginLinef("%s# ending tailing '%s' logs\n", linePrefix, l.tag)
-				return nil
+				return err
```
praveenrewar: I think we can remove this if condition now as we are returning the err at the end as well.

pkg/kapp/logs/pod_container_log.go
```diff
@@ -140,6 +163,16 @@ func (l PodContainerLog) readyToGetLogs() bool {
 	if err != nil {
 		return false
 	}
+	containerStatuses := pod.Status.ContainerStatuses
+	for _, containerStatus := range containerStatuses {
+		//We should focus only on the status of this particular container.
+		if l.container != containerStatus.Name {
+			continue
+		}
```
praveenrewar: Is this an old comment that you forgot to remove? As it says we should focus only on this particular container but we are just doing a continue in the if condition.

pkg/kapp/logs/pod_container_log.go
praveenrewar: I am not fully aware how and when this gets populated, but based on my understanding so far, this seems fine to me.

pkg/kapp/logs/pod_container_log.go
```diff
@@ -140,6 +163,16 @@ func (l PodContainerLog) readyToGetLogs() bool {
...
+		if containerStatus.State.Running != nil {
+			return true
+		}
+	}
+	return false
```
praveenrewar: nit: add newline before return statement and remove the one after it.

pkg/kapp/logs/pod_container_log.go
100mik: I believe it can be better phrased as "Filter out statuses from containers other than the one being tailed"
I believe a pod might have statuses from multiple containers and this means that we are only looking at ones from the container being observed

praveenrewar: Yeah, I did think that @rohitagg2020 started with if `l.container == containerStatus.Name` and then changed it to `if l.container != containerStatus.Name`, but forgot to update the comment later on. 

> "Filter out statuses from containers other than the one being tailed"

Makes sense. Although I would recommend to remove the comment as the if condition seems self explanatory to me.

pkg/kapp/logs/pod_container_log.go
rohitagg2020: `else` condition is for `l.opts.Follow`. continue is because I don't want to return from the function. Scenario is: if somebody has given `-f` option, then we should continuously try to get the stream

praveenrewar: I apologise for not being clear earlier, but what I meant was that, this
```go
for {
    if l.opts.Follow {
        //do something
        continue
    } else {
        //do something else
        return
    }
}
```
would behave same as this
 ```go
for {
    if l.opts.Follow {
        //do something
        continue
    }
    //do something else
    return
}
```
as `continue` at the end of `if` would have the same effect as that of an `else` statement.

**Comment by praveenrewar** [comment]
It doesn't matter much as we are going to use `Squash and merge`, but I think it would be better to use rebase instead of merge to get the latest changes from develop.

File: pkg/kapp/logs/pod_container_log.go [hunk 9]
```diff
@@ -140,6 +161,15 @@ func (l PodContainerLog) readyToGetLogs() bool {
 	if err != nil {
 		return false
 	}
+	containerStatuses := pod.Status.ContainerStatuses
```
cppforlife, line 167: what about init containers? this gets run against initContainers + plain containers: https://github.com/vmware-tanzu/carvel-kapp/blob/e0bf1c9d485c5c59a23b4cfe75915782e7fb4b42/pkg/kapp/logs/pod_log.go#L70

pkg/kapp/logs/pod_container_log.go
rohitagg2020: Adopted

pkg/kapp/logs/pod_container_log.go
rohitagg2020: I think it is more clear to me as it is right now

pkg/kapp/logs/pod_container_log.go
rohitagg2020: Adopted

pkg/kapp/logs/pod_container_log.go
rohitagg2020: No, as of today, it doesnt get run against `init containers`. I will add that as well.

rohitagg2020: Adopted

pkg/kapp/logs/pod_container_log.go
```diff
@@ -95,11 +119,10 @@ func (l PodContainerLog) Tail(ui ui.UI, cancelCh chan struct{}) error {
 }
 
 func (l PodContainerLog) obtainStream(ui ui.UI, linePrefix string, cancelCh chan struct{}) (io.ReadCloser, error) {
-	tryCount := 0
-
+	isWaitingMsgPrintedOnce := false
```
sethiyash: nit:
```
var (
	isWaitingMsgPrintedOnce bool
	err                     error
)
```

sethiyash: didn't quite understand the purpose of declaring `err` here as we are just returning it from line 148 only and not getting used anywhere else if so can we return `errors.New("")` from there as well ?

rohitagg2020: Did it so that in the future if we want to use the actual error inside the parent function, then no need to make a change at line: 148.

rohitagg2020: Adopted nit.

pkg/kapp/logs/pod_container_log.go
```diff
@@ -44,6 +44,34 @@ func NewPodContainerLog(
...
 
 func (l PodContainerLog) Tail(ui ui.UI, cancelCh chan struct{}) error {
+	linePrefix := ""
+	if len(l.opts.LinePrefix) > 0 {
+		linePrefix = l.opts.LinePrefix + " | "
+	}
+
+	for {
+		err := l.StartTail(ui, cancelCh)
+
```
cppforlife: nit: newline

pkg/kapp/logs/pod_container_log.go
```diff
@@ -44,6 +44,34 @@ func NewPodContainerLog(
...
+		if err == io.EOF {
+
```
cppforlife: nit: newline

pkg/kapp/logs/pod_container_log.go
```diff
@@ -44,6 +44,34 @@ func NewPodContainerLog(
...
+			if l.opts.Follow {
+				ui.BeginLinef("%s# container stopped '%s' logs\n", linePrefix, l.tag)
+				// Multiplying by a factor of 2 as some times for initContainers, it fetches the older stream.
+				time.Sleep(2 * 500 * time.Millisecond)
```
cppforlife: 1 * time.Second?

pkg/kapp/logs/pod_container_log.go
rohitagg2020: Adopted

pkg/kapp/logs/pod_container_log.go
rohitagg2020: Adopted

pkg/kapp/logs/pod_container_log.go
rohitagg2020: Adopted

PR #386, opened 2021-12-08
Don't panic on error in rebaseRule provided by user

redbaron:
User can provide rebaseRule with `{allIndexes: true}` as last element , which is not supported and should generate error message.

Other panics in this file left as is, because when these are more like asserts, indicating bug in the code when triggered.

### DIFF
--- a/pkg/kapp/resources/mod_field_copy.go
+++ b/pkg/kapp/resources/mod_field_copy.go
@@ -74,7 +74,7 @@ func (t FieldCopyMod) apply(obj interface{}, path Path, fullPath Path, srcs map[
 
 		case part.ArrayIndex != nil:
 			if isLast {
-				panic("Expected last part of the path to be map key")
+				return false, fmt.Errorf("Expected last part of the path to be map key")
 			}
 
 			switch {


pkg/kapp/resources/mod_field_copy.go
```diff
@@ -74,7 +75,7 @@ func (t FieldCopyMod) apply(obj interface{}, path Path, fullPath Path, srcs map[
 
 		case part.ArrayIndex != nil:
 			if isLast {
-				panic("Expected last part of the path to be map key")
+				return false, errors.New("Expected last part of the path to be map key")
```
100mik: We generally use `fmt` for creating our errors. Could you make this change?
```suggestion
				return false, fmt.Errorf("Expected last part of the path to be map key")
```

100mik (changes requested):
Looks reasonable! Thanks for the PR. Just a minor change to make sure it's consistent with the rest of the codebase.

pkg/kapp/resources/mod_field_copy.go
redbaron: @100mik, even if error message has no placeholders?

100mik: Yep! the idea is to stick to one import for the purpose of structuring strings _and_ errors

redbaron: Done! I didn't use github suggestion because there'd be unused import.

100mik (approved):
LGTM! 
Thanks for the contributions @redbaron 🚀

pkg/kapp/resources/mod_field_copy.go
100mik: Thank you 🥳

PR #389, opened 2021-12-09
Labeling kapp App and filtering App list on label and age (#380)

### LINKED ISSUE #380: support labels for app-groups
We like to annotate certain apps in order to filter them later for cleanup.

I see that kapp deploy --labels ... is supported, but that flag is not available under kapp app-group deploy.

Background: 
We like to run a cronjob that removes test deployments that are older than a week, keeping production deployments.
Devs are able to create preview deployments for acceptance test, but if they forget to clean them up they keep lingering and filling up the cluster. We cannot simply remove the namespace, we need to filter on apps.
So something like an annotation or label I assume could do the trick to mark production type apps, but this is not supported with app-groups

adding labels for app-group deployments like it is available for normal deployment

Assuming label is applied, what are your steps in finding them (by label and age) and deleting them?

sethiyash:
- Added functionality to set labels with app-group command.
- listing kapp apps with filtering on labels `kapp ls --filter-labels x=y`
-  listing kapp apps with filtering on age `kapp ls --filter-age 5m+`
- ~~showing all kapp app labels with `kapp ls --column label`~~


pkg/kapp/cmd/app/list.go
```diff
@@ -58,11 +61,21 @@ func (o *ListOptions) Run() error {
 		return err
 	}
 
-	items, err := supportObjs.Apps.List(nil)
+	appFilter, err := o.AppFilterFlags.AppFilter()
```
100mik, line 64: Missing error check!

test/e2e/list_test.go
```diff
@@ -0,0 +1,108 @@
...
+		kapp.Run([]string{"delete", "-a", name})
+		kapp.Run([]string{"delete", "-a", name2})
+	}
+
+	cleanUp()
+	defer cleanUp()
+	logger.Section("App listing and filter label", func() {
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name,
+			"--labels", "x=y"}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+		expectedOutput1 := `
```
100mik: We should be checking JSON output here

test/e2e/list_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+Name           Namespaces  Lcs   Lca  
+test-app-list  kapp-test   true  0s  
+
+Lcs: Last Change Successful
+Lca: Last Change Age
+
+1 apps
+
+Succeeded
```
100mik: I think for other test cases having tabular output we should stick to checking JSON format as well

test/e2e/list_test.go
```diff
@@ -0,0 +1,108 @@
...
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	yaml1 := `
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: redis-primary
+  labels:
+    x: "y"
```
praveenrewar: I think this is not required as we are adding the labels to whole app while deploying it.

test/e2e/list_test.go
```diff
@@ -0,0 +1,108 @@
...
+    app: redis
+    tier: backend
+`
+	yaml2 := `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: redis-config
+  labels:
+    x: "z"
```
praveenrewar: Shouldn't be required (reason: same as above)

test/e2e/list_test.go
praveenrewar: I think it's not necessary to test this output as we are just adding the label to the app here and therefore it's just a generic deployment summary which is tested in many places.

test/e2e/list_test.go
```diff
@@ -0,0 +1,108 @@
...
+kapp-test  redis-primary  Service  -       -    create  -       reconcile  -   -  
+
+Op:      1 create, 0 delete, 0 update, 0 noop, 0 exists
+Wait to: 1 reconcile, 0 delete, 0 noop
+`
+		require.Contains(t, out, expectedOutput1, "Did not find expected diff output")
+
+		out2, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name2,
+			"--labels", "a=b"}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml2)})
+		expectedOutput2 := `
```
praveenrewar: Might not be necessary (reason: same as above)

test/e2e/list_test.go
```diff
@@ -0,0 +1,108 @@
...
+test-app-list-2  kapp-test   true  0s  
+
+Lcs: Last Change Successful
+Lca: Last Change Age
+
+2 apps
+`
+		require.Contains(t, listedApps, expectedAppsList, "Did not find expected diff output")
+
+		filteredApps, _ := kapp.RunWithOpts([]string{"ls", "--filter-labels", "x=y"}, RunOpts{Interactive: true})
```
praveenrewar: Should we also test filter-age flag?

pkg/kapp/app/app_filter.go
```diff
@@ -0,0 +1,39 @@
...
+		if f.Matches(app) {
+			result = append(result, app)
+		}
+	}
+	return result
+}
+
+func (f AppFilter) Matches(app App) bool {
+	if f.CreatedAtBeforeTime != nil {
+		if lc, _ := app.LastChange(); lc.Meta().StartedAt.After(*f.CreatedAtBeforeTime) {
```
praveenrewar: Although inline check seems fine but just wondering if we should call `LastChange()` before the if condition and do an error check as well.

pkg/kapp/cmd/app/list.go
sethiyash: Yeah. Thanks Done

test/e2e/list_test.go
sethiyash: Yeah make sense removed.

test/e2e/list_test.go
sethiyash: Removed.

test/e2e/list_test.go
sethiyash: removed.

test/e2e/list_test.go
sethiyash: Removed.

test/e2e/list_test.go
sethiyash: Done

test/e2e/list_test.go
sethiyash: Done

pkg/kapp/app/app_filter.go
sethiyash: Yeah added error check.

pkg/kapp/app/app_filter.go
```diff
@@ -0,0 +1,45 @@
...
+	Labels []string
+}
+
+func (f AppFilter) Apply(apps []App) ([]App, error) {
+	var result []App
+
+	for _, app := range apps {
+		lastChange, err := app.LastChange()
+		if err != nil {
+			return []App{}, nil
```
rohitagg2020: Seems like we are eating the error. We should `log` it atleast. Ignore the comment if it is already being logged in `LastChange` method.

sethiyash: My bad we are not logging in `LastChange` started returning err in case of non nil.

test/e2e/list_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+		listedAppsWithAge, _ := kapp.RunWithOpts([]string{"ls", "--filter-age", "2m+", "--json"}, RunOpts{Interactive: true})
+
+		response := uitest.JSONUIFromBytes(t, []byte(listedAppsWithAge))
+
+		if len(response.Tables) > 0 {
+			require.Empty(t, response.Tables[0].Rows, "Expected table rows to empty")
+		}
+
+		time.Sleep(2 * time.Second)
```
praveenrewar: Let's add the `time.Sleep(2 * time.Second)` before deploying the second app, so that only the first app shows up while using the `2s+` filter.

test/e2e/list_test.go
```diff
@@ -0,0 +1,100 @@
...
+			"namespaces":             "kapp-test",
+		}, {
+			"last_change_successful": "true",
+			"name":                   "test-app-2",
+			"namespaces":             "kapp-test",
+		}}
+
+		resp := uitest.JSONUIFromBytes(t, []byte(listedApps))
+
+		validateAppListChanges(t, resp.Tables, expectedAppsList, listedApps)
```
praveenrewar: Instead of deleting the `last_change_age`, I think we can use something like [replaceAge](https://github.com/vmware-tanzu/carvel-kapp/blob/6b191d2550397dc1007f5bcb2ef536d0358fdd3f/test/e2e/diff_test.go#L389) to update the age of the app and then use `require.Equal()`

test/e2e/list_test.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"strings"
+	"testing"
+	"time"
+
+	uitest "github.com/cppforlife/go-cli-ui/ui/test"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAppListing(t *testing.T) {
```
praveenrewar: Should we call it `TestAppFilter` instead? (If so, we might wanna update the file name as well)

test/e2e/list_test.go
```diff
@@ -0,0 +1,100 @@
...
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	yaml1 := `
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: redis-primary
+  labels:
+    x: "y"
```
praveenrewar: I thought we were planning to remove this as it's not required?

test/e2e/list_test.go
```diff
@@ -0,0 +1,100 @@
...
+    app: redis
+    tier: backend
+`
+	yaml2 := `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: redis-config
+  labels:
+    x: "z"
```
praveenrewar: I thought we were planning to remove this as it's not required?

pkg/kapp/app/app_filter.go
```diff
@@ -0,0 +1,45 @@
...
+	Labels []string
+}
+
+func (f AppFilter) Apply(apps []App) ([]App, error) {
+	var result []App
+
+	for _, app := range apps {
+		lastChange, err := app.LastChange()
+		if err != nil {
+			return []App{}, err
```
rohitagg2020: nit: Should we return `nil, err` instead of `[]App{}, err`.

test/e2e/list_test.go
sethiyash: Yeah Done. Also moved  replaceAge in common.go file for the reusability.

test/e2e/list_test.go
sethiyash: Done. was not giving consistent behaviour as second app is getting 5s+ older sometimes so checking for 10s+ to just get the first app consistently (it can depend on the machine we are running on)

test/e2e/list_test.go
sethiyash: Yeah missed that. Removed now.

test/e2e/list_test.go
sethiyash: Yeah missed that. Removed now.

test/e2e/list_test.go
sethiyash: Yeah make sense. made the changes.

pkg/kapp/app/app_filter.go
sethiyash: Done.

praveenrewar (approved):
LGTM

test/e2e/app_filter_test.go
```diff
@@ -0,0 +1,85 @@
...
+		kapp.Run([]string{"delete", "-a", name2})
+	}
+
+	cleanUp()
+	defer cleanUp()
+	logger.Section("App listing and filter label", func() {
+		_, _ = kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name,
+			"--labels", "x=y"}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+
+		time.Sleep(2 * time.Second)
```
rohitagg2020, line 55: Q: Will this pass everytime? I think yesterday we were talking about this .

sethiyash:
Hi @universam1

As I can see you are monitoring this PR closely would love to hear your thoughts on this. 

While doing like `kapp ls --filter-age 24h-` currently we are filtering Kapp Apps based on the creation time. 

Curious to know if you would like to see something like - `kapp ls --filter-last-change-age 12h+`  this will filter apps based on last change age.  As while doing `kapp ls` we show last change age instead of actual age.

universam1:
> Curious to know if you would like to see something like - `kapp ls --filter-last-change-age 12h+` this will filter apps based on last change age. As while doing `kapp ls` we show last change age instead of actual age.

Thanks foremost for reaching out!
I guess `last-change-age` could be useful for our cleanup purpose as well. It might be even more handy selector to discover stale deployments.
We'd basically like to be able to mark a certain app to be ephemeral or a test deployment, and then discover its staleness.
So I believe that should work fine @sethiyash 
Thanks!

pkg/kapp/app/app_filter.go
```diff
@@ -0,0 +1,40 @@
...
+			return false
+		}
+	}
+
+	if f.CreatedAtAfterTime != nil {
+		if app.CreationTimestamp().Before(*f.CreatedAtAfterTime) {
+			return false
+		}
+	}
+	return true
```
praveenrewar: nit: add newline

pkg/kapp/app/recorded_app.go
```diff
@@ -23,8 +23,10 @@ const (
 )
 
 type RecordedApp struct {
-	name   string
-	nsName string
+	name              string
+	nsName            string
+	creationTimestamp time.Time
+	labels            map[string]string
```
praveenrewar: We are trying to add more values to RecordedApp in another [PR](https://github.com/vmware-tanzu/carvel-kapp/pull/403/files#diff-c4cadf071a66c0e934e58c35eb0a9116fee4c744abd14621dc1eeebe7f5e9fb8R31-R32).
I feel like this might get out of hand soon. Should we do some grouping on these resources, something like RecordedAppMeta? (Absolutely open to other suggestions)

sethiyash: Yeah make sense we can have RecordedAppMeta as a separate struct.

pkg/kapp/app/app_filter.go
sethiyash: Done

pkg/kapp/app/recorded_app.go
sethiyash: Done

test/e2e/common.go
```diff
@@ -24,3 +25,28 @@ func validateChanges(t *testing.T, respTable []ui.JSONUITableResp, expected []ma
 	require.Equalf(t, notesWaitTo, respTable[0].Notes[1], "Expected to see correct summary, but did not: '%s'", output)
 
 }
+
+func replaceAge(result []map[string]string) []map[string]string {
```
rohitagg2020, line 29: If this function is used only in the test cases, then it shouldn't be part of the main code.

test/e2e/common.go
```diff
@@ -24,3 +25,28 @@ func validateChanges(t *testing.T, respTable []ui.JSONUITableResp, expected []ma
...
+	for i, row := range result {
+		if len(row["age"]) > 0 {
+			row["age"] = "<replaced>"
+		}
+		result[i] = row
+	}
+	return result
+}
+
+func replaceLastChangeAge(result []map[string]string) []map[string]string {
```
rohitagg2020, line 39: If this function is used only in the test cases, then it shouldn't be part of the main code.

rohitagg2020:
Added Comments.

pkg/kapp/app/recorded_app.go
sethiyash: We can put this on hold as lot of refactoring is required and get everyones view on this and then let's take a call  ?

test/e2e/app_filter_test.go
sethiyash: Yeah I have checked this are getting passed.

pkg/kapp/app/recorded_app.go
rohitagg2020: I agree, If we want to do it, let's open another issue and work on it.

pkg/kapp/app/recorded_app.go
```diff
@@ -8,23 +8,26 @@ import (
...
 
-	"github.com/k14s/kapp/pkg/kapp/logger"
-	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes"
+
```
cppforlife: nit: extra newline

pkg/kapp/app/recorded_app.go
```diff
@@ -217,7 +222,12 @@ func (a *RecordedApp) labeledApp() (*LabeledApp, error) {
 		return nil, err
 	}
 
-	sel := labels.Set(meta.Labels()).AsSelector()
+	appLabels := meta.Labels()
+	for key, val := range a.labels {
+		appLabels[key] = val
+	}
+
+	sel := labels.Set(appLabels).AsSelector()
```
cppforlife: we should not be adding labels used to label app to LabeledApp

sethiyash: ok yeah, make sense to me. But now I'm not getting if I don't set them how I can get additional labels other than `kapp.k14s.io/app=<something>` with the help of label selector [here](https://github.com/vmware-tanzu/carvel-kapp/blob/edeb9dd9a923e15187f88999e983cb5762a14faa/pkg/kapp/cmd/app/list.go#L100)

Will it be a good idea to have a separate `getLabels` method for App interface and then concatenate this into what we are getting from `sel.String()` ?

sethiyash: Make sense. As this functionality was added additionally will it be a good idea to raise a separate issue for showing all the kapp app labels while doing `kapp ls --column name,label` and let's move this out of the scope of this PR ?

I'm removing this change as of now.

pkg/kapp/app/recorded_app.go
sethiyash: Done

pkg/kapp/app/recorded_app.go
```diff
@@ -39,8 +41,10 @@ var _ App = &RecordedApp{}
 func (a *RecordedApp) Name() string      { return a.name }
 func (a *RecordedApp) Namespace() string { return a.nsName }
 
+func (a *RecordedApp) CreationTimestamp() *time.Time { return &a.creationTimestamp }
+
 func (a *RecordedApp) Description() string {
-	return fmt.Sprintf("app '%s' namespace: %s", a.name, a.nsName)
+	return fmt.Sprintf("app '%s' namespace: %s", a.Name(), a.Namespace())
```
cppforlife: (i seemed to have lost a comment i thought i made...) lets use a.name and a.namespace within this class. i dont see a reason why we need to use public methods in this PR (will make it harder for other folks to rebase if they changed this file in active PRs). additionally generally within classes we do not use public methods to reference necessary info -- if you starting to do so, if you find yourself using public methods in other public methods, this might be an indication that these are 2 different classes shoved into one.

test/e2e/app_filter_test.go
```diff
@@ -0,0 +1,86 @@
...
```
cppforlife, line 55: lets live a comment why we need this sleep.

test/e2e/app_filter_test.go
```diff
@@ -0,0 +1,86 @@
...
+			"last_change_successful": "true",
+			"name":                   "test-app-1",
+			"namespaces":             "kapp-test",
+		}}
+
+		resp := uitest.JSONUIFromBytes(t, []byte(listedApps))
+
+		require.Equalf(t, expectedAppsList, replaceLastChangeAge(resp.Tables[0].Rows), "Expected to match")
+
+		filteredApps, _ := kapp.RunWithOpts([]string{"ls", "--filter-labels", "x=y", "--json"}, RunOpts{Interactive: true})
```
cppforlife: lets switch this to a=b so that we get a different result from previous list result.

pkg/kapp/app/recorded_app.go
sethiyash: My motivation behind this is as I have got comments saying that can we have something like `RecordedAppMeta` struct inside `RecordedApp` struct to segregate the meta fields so I feel like we should start using methods so that if we go that way we just have to change the methods in future. 

But as we are not adding something like `RecordedAppMeta` and rebase can be an issue I can revert this.

pkg/kapp/app/recorded_app.go
cppforlife: what uses labels?

pkg/kapp/app/app_filter.go
```diff
@@ -0,0 +1,41 @@
...
+
+package app
+
+import "time"
+
+type AppFilter struct {
+	CreatedAtBeforeTime *time.Time
+	CreatedAtAfterTime  *time.Time
+
+	Labels []string
```
cppforlife: what uses this?

pkg/kapp/app/app_filter.go
```diff
@@ -0,0 +1,41 @@
...
+	for _, app := range apps {
+		if f.Matches(app) {
+			result = append(result, app)
+		}
+	}
+	return result, nil
+}
+
+func (f AppFilter) Matches(app App) bool {
+
```
cppforlife: nit: extra newline

pkg/kapp/app/interfaces.go
```diff
@@ -6,11 +6,13 @@ package app
 import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"time"
 )
 
 type App interface {
 	Name() string
 	Namespace() string
+	CreationTimestamp() *time.Time
```
cppforlife: should this really be a pointer?

pkg/kapp/app/app_filter.go
sethiyash: Getting used to store labels got from `--filter-labels` flag

pkg/kapp/app/recorded_app.go
sethiyash: For setting `app.ObjectMeta.Labels` value in `labels` for recordedApp `recordedApp := &RecordedApp{app.Name, app.Namespace, app.ObjectMeta.Labels, a.coreClient...`

pkg/kapp/app/interfaces.go
sethiyash: Was initially made it pointer to return nil from `func (a *LabeledApp) CreationTimestamp() *time.Time { return nil }` but we can return the time zero value here started doing that.

test/e2e/app_filter_test.go
sethiyash: Added.

test/e2e/app_filter_test.go
sethiyash: Done

pkg/kapp/app/app_filter.go
```diff
@@ -0,0 +1,40 @@
...
+type AppFilter struct {
+	CreatedAtBeforeTime *time.Time
+	CreatedAtAfterTime  *time.Time
+
+	Labels []string
+}
+
+func (f AppFilter) Apply(apps []App) ([]App, error) {
+	var result []App
+
```
100mik, line 15: Do we wanna add a check where we verify whether or not After time is less than Before time if both are specified or just show the user an empty table?

100mik (commented):
Good to go other than this!

pkg/kapp/cmd/app/list.go
```diff
@@ -58,7 +61,23 @@ func (o *ListOptions) Run() error {
...
+	if err != nil {
+		return err
+	}
+
+	labelFlags := &LabelFlags{Labels: appFilter.Labels}
```
rohitagg2020: nit: Can we rename the variable to `filterLabelFlags` or `filterLabels`

pkg/kapp/cmd/app/list.go
```diff
@@ -58,7 +61,23 @@ func (o *ListOptions) Run() error {
...
+	additionalLabels, err := labelFlags.AsMap()
+	if err != nil {
+		return err
+	}
+
+	items, err := supportObjs.Apps.List(additionalLabels)
```
rohitagg2020: nit: Can we rename it to `filteredApps` or something like that.

test/e2e/app_filter_test.go
```diff
@@ -0,0 +1,87 @@
...
+kind: Service
+metadata:
+  name: redis-primary
+spec:
+  ports:
+  - port: 6380
+    targetPort: 6380
+  selector:
+    app: redis
+    tier: backend
```
rohitagg2020: nit: We can remove `tier`

rohitagg2020:
LGTM

pkg/kapp/app/app_filter.go
sethiyash: Here we don't have something like `--before` and `--after` flags we just take `--filter-age` flag from user and this both `CreatedAtBeforeTime`  `CreatedAtAfterTime` we are creating internally for filtering by age.

pkg/kapp/cmd/app/list.go
sethiyash: Adopted.

test/e2e/app_filter_test.go
sethiyash: Adopted.

pkg/kapp/app/app_filter.go
100mik: Oh I see, makes sense 🚀

100mik (approved):
LGTM!

Issue #388, opened 2021-12-08
Support additional apply mechanism -- server side apply


redbaron:
**Describe the problem/challenge you have**
I'd like make use of server side apply feature, bypassing rebaseRule entirely.

**Describe the solution you'd like**

An optional support for server-side apply, enabled by the `--server-side` flag. 

**Anything else you would like to add:**

When server-side apply is in use rebaseRules are entirely bypassed. Diff is generated using dry run and diffing result with kubernetes objects currently in cluster to get a most accurate preview of changes. 

Should server side apply generate a conflict, it can be resolved in kapp favour  with `--force` flag.

`kapp.k14s.io/original` annotation remains populated to support subsequent kapp invocations without `--server-side` flag.  Default rebaseRules probably need to be updated to keep cluster-managed `metadata.managedFields` path.

Related issues: https://github.com/vmware-tanzu/carvel-kapp/issues/43
 

Server side apply doc: https://kubernetes.io/docs/reference/using-api/server-side-apply/

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

100mik:
Interesting suggestion! We were working on having a `--dry-run` feature which would probably complement the [server side validation](https://github.com/kubernetes/enhancements/issues/2885) which just made it to alpha I believe. We did run into a few edge cases like where users were deploying CRDs and CRs using the same manifest.

Will definitely explore the Server Side apply feature to see how we can leverage it. 🥂
Worth noting that we do generate diffs using the resource on the cluster when there are changes 🤔 
Using the `--diff-changes` (or `-c`) flag will help you see the exact changes being applied before you confirm them.

cppforlife:
> > it looks like server side apply and it's conflict resolution framework is better approach
>
> kapp tries to provide a guarantee that what gets applied results in same outcome regardless of what happened on the server. in server side apply world, for example, keys that are set by something else on the server side will never be unset by future apply that do not touch that section. this definitely "fights" against general philosophy of kapp. but lets take that convo to this other thread you started.

(copying my comment from https://github.com/vmware-tanzu/carvel-kapp/issues/385#issuecomment-989290669)

redbaron:
[...]

PR #392, opened 2021-12-14
Support for server side apply

redbaron:
First take on server side apply feature  (https://github.com/vmware-tanzu/carvel-kapp/issues/388) to collect early feedback.

Some highlights:
- user provided manifest + some kapp generated annotations is what we are submitting to K8S via PATCH request. 
- diff calculation is at the core of `kapp` code paths, so I had to approximate it by diffing existing instances against dry run SSA result. This required calls to K8S much earlier in the code than it was previously, which required some plumbing
 - because we have to make dry run call, there is a new  `NewChangeSSA` change factory
 - E2E test runner now supports `KAPP_E2E_SSA=1` env var value. When set, `--ssa-enabled` flag is added to kapp invocation allowing same tests to be run in SSA mode. Tests with rebasing are skipped in SSA mode.
 - To make debugging e2e cases easier I created embeddable runner, which runs kapps `*corbra.Command` in the testing process. It is not a full replacement for a `kapp` command execution, but greatly helped developing this feature as most interesting scenarios come up in `e2e` and not   unit testing.

Unimplemented:
- [x] SSA conflict force flag. It is present, but not used anywhere.
- [x] conflicting CLI flags are ignored in favour of SSA feature. I think there should be  a clear error message if conflicting flags are specificed, like `--diff-against-last-applied`.

Yet untested (unless there is an existing E2E test covering it):
- [ ] how changes to namespace are handled
- [ ] how changes to apiVersion are handled (apps/v1beta1 -> apps/v1)

Features deliberately left outside of this PR:
- Ownership transfer as described in [docs](https://kubernetes.io/docs/reference/using-api/server-side-apply/#transferring-ownership).  There might be situations where user might want to give up ownership of certain fields after initial creation.  `kapp` could assist with it.
- Fine grained force conflict resolution. Right now it is all or nothing. There is no fine grained force option in the kubernetes, so it has to be implemented as 2 separate operations: non-SSA patch for conflicting fields + SSA for the rest. I'll create an issue with more detailed proposal once SSA work is merged


100mik:
Thanks for the draft PR @redbaron !
Will definitely go through the changes to get an idea of what's going on.

pkg/kapp/diff/change_factory.go
```diff
@@ -4,26 +4,76 @@
...
+func (f ChangeFactory) NewChangeAgainstLastApplied(ctx context.Context, existingRes, newRes ctlres.Resource) (Change, error) {
 	// Retain original copy of existing resource and use it
 	// for rebasing last applied resource and new resource.
 	existingResForRebasing := existingRes
+	var err error
 
 	if existingRes != nil {
+		// Strip rebasing "base" object of kapp history so that it never shows up in the diff, regardless
+		// of rebase rules used
+		existingResForRebasing, err = f.NewResourceWithHistory(existingRes).HistorylessResource()
```
redbaron, line 83: I believe there was a sleeper bug here, which unfolded like following:
- By the end of this function `existingRes` was unconditionally stripped of history with `HistorylessResource()`
- `newRes` was unconditionally stripped of history with `HistorylessResource()`
- `existingResForRebasing` has history kept, because it comes from the cluster as is
- When building `rebasedNewRes` using `existingResForRebasing + newRes`, history was removed due to default rebase rules included in the kapp transplanting history-less annotations from `newRes` 
- If we redeploy same object as previous deployment, end result has no spurious diff, because both objects have history stripped

Should provided config had different rebase rules, `rebasedNewRes` is going to have history annotations from `existingResForRebasing`, but `existingRes` wont, generating spurious diff.

Similar situation arises when `metadata.managedFields` is populated by the cluster. This field is considered history and is stripped in `HistorylessResource()` call, but default rebase rules have no mentions of it.

Instead of updating default rebase rule, I opted for fixing it at the source. Here I am stripping history from all objects, ensuring it is not going to show up in any diffs regardless of kapp config provided. Technically it is a change in behaviour, hence this comment to highlight it.

pkg/kapp/cmd/app/apply_flags.go
```diff
@@ -62,6 +64,10 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 	cmd.Flags().IntVar(&s.WaitingChangesOpts.Concurrency, prefix+"wait-concurrency",
 		5, "Maximum number of concurrent wait operations")
 
+	cmd.Flags().BoolVar(&s.ServerSideApply, prefix+"server-side", false, "If true, apply runs in the server instead of the client")
+	cmd.Flags().StringVar(&s.FieldManagerName, prefix+"field-manager", "kapp-server-side-apply", "Name of the manager used to track field ownership")
+	cmd.Flags().BoolVar(&s.ServerSideForceConflict, prefix+"force-conflicts", false, "If true, server-side apply will force the changes against conflicts.")
```
cppforlife: lets prefix these flags with server-side-apply (or may be just ssa) e.g.

--server-side-apply
--server-side-apply-field-manager
--server-side-apply-force-conflicts

or 

--ssa
--ssa-field-manager
--ssa-force-conflicts

cppforlife: also are there cases when somebody would want to force conflicts for one but not the other resource? what would be workflow there, or is it pretty much always --ssa-force-conflicts=true (and if so should that be default?)

pkg/kapp/cmd/app/deploy.go
```diff
@@ -51,7 +52,12 @@ func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 		Use:     "deploy",
 		Aliases: []string{"d", "dep"},
 		Short:   "Deploy app",
-		RunE:    func(_ *cobra.Command, _ []string) error { return o.Run() },
+		RunE: func(_ *cobra.Command, _ []string) error {
+			if o.ApplyFlags.ServerSideApply {
+				o.DiffFlags.ChangeSetOpts.Mode = ctldiff.ServerSideApplyChangeSetMode
+			}
```
cppforlife: not a fan of doing it in RunE, lets move this into Run()... (also going to see if there is a better place for this)

pkg/kapp/cmd/app/deploy.go
```diff
@@ -315,19 +324,17 @@ func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 	return resourceFilter.Apply(existingResources), o.existingPodResources(existingResources), nil
 }
 
-func (o *DeployOptions) calculateAndPresentChanges(existingResources,
-	newResources []ctlres.Resource, conf ctlconf.Conf, supportObjs FactorySupportObjs) (
-	ctlcap.ClusterChangeSet, *ctldgraph.ChangeGraph, bool, string, error) {
+func (o *DeployOptions) calculateAndPresentChanges(ctx context.Context, existingResources, newResources []ctlres.Resource, conf ctlconf.Conf, supportObjs FactorySupportObjs) (ctlcap.ClusterChangeSet, *ctldgraph.ChangeGraph, bool, string, error) {
```
cppforlife: nit: undo formatting change -- easier to see when it's on multiple lines

pkg/kapp/cmd/tools/diff_flags.go
```diff
@@ -35,7 +35,11 @@ func (s *DiffFlags) SetWithPrefix(prefix string, cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&s.LineNumbers, prefix+"line-numbers", true, "Show line numbers")
 	cmd.Flags().BoolVar(&s.Mask, prefix+"mask", true, "Apply masking rules")
 
-	cmd.Flags().BoolVar(&s.AgainstLastApplied, prefix+"against-last-applied", true, "Show changes against last applied copy when possible")
+	if *cmd.Flags().Bool(prefix+"against-last-applied", true, "Show changes against last applied copy when possible. Ignored when server-side apply is used.") {
```
cppforlife: nit: lets put (`Ignored when server-side apply is used.` into parenthesis). we typically dont have multiple sentences in descriptions.

pkg/kapp/diff/change_factory.go
```diff
@@ -4,26 +4,76 @@
...
 func NewChangeFactory(rebaseMods []ctlres.ResourceModWithMultiple,
-	diffAgainstLastAppliedFieldExclusionMods []ctlres.FieldRemoveMod) ChangeFactory {
+	diffAgainstLastAppliedFieldExclusionMods []ctlres.FieldRemoveMod, resources ctlres.Resources) ChangeFactory {
+
+	return ChangeFactory{rebaseMods, diffAgainstLastAppliedFieldExclusionMods, resources}
+}
+
+func (f ChangeFactory) NewChangeSSA(ctx context.Context, existingRes, newRes ctlres.Resource) (Change, error) {
+	if existingRes != nil {
+		historylessExistingRes, err := f.NewResourceWithHistory(existingRes).HistorylessResource()
```
cppforlife: should we even bother with "history" when doing SSA?

pkg/kapp/diff/change_factory.go
cppforlife: will think through it, thanks for call out 👍

test/e2e/diff_test.go
```diff
@@ -120,8 +120,9 @@ data:
 	})
 
 	logger.Section("deploy no change", func() {
-		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--json"},
+		out, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--json"},
 			RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+		require.NoError(t, err)
```
cppforlife: no need to check on err -- by default kapp object should do it for you unless AllowError is set to true. (this just makes it less noisy to write tests)

pkg/kapp/diff/change_set.go
```diff
@@ -23,13 +32,24 @@ type ChangeSet struct {
 func NewChangeSet(existingRs, newRs []ctlres.Resource,
 	opts ChangeSetOpts, changeFactory ChangeFactory) *ChangeSet {
 
+	if opts.Mode == "" {
+		opts.Mode = ExactChangeSetMode
+	}
```
cppforlife, line 37: seems like we should be able to make tighter assertion that mode isnt empty instead of just defaulting.

cppforlife (commented):
looking pretty good so far.

pkg/kapp/cmd/app/apply_flags.go
redbaron: I kept flag names and defaults as they are in `kubectl apply`, which many people are familiar with.

pkg/kapp/cmd/app/deploy.go
redbaron: I plan to add flag validation step and return error, rather than silently overwriting values provided by user, so it should be a RunE I believe because it can return error.

test/e2e/diff_test.go
redbaron: I'll do same error handling in RunEmbedded then

pkg/kapp/diff/change_set.go
redbaron: It is to preserve original behaviour where empty `ChangeSetOpts{}` used  exact change factory function.  Empty `ChangeSetOpts{}` is used in tests only.

test/e2e/diff_test.go
```diff
@@ -87,7 +87,7 @@ data:
 			"conditions":      "",
 			"kind":            "ConfigMap",
 			"name":            "redis-config",
-			"namespace":       "kapp-test",
+			"namespace":       env.Namespace,
```
praveenrewar, line 90: 💯

pkg/kapp/cmd/app/deploy.go
redbaron: done

pkg/kapp/cmd/tools/diff_flags.go
redbaron: Done. Also changed wording to indicate conflict.

pkg/kapp/diff/change_factory.go
redbaron: You are right. Because diff is between existing and dry run apply result, history remains unchanged between these two, so no need to strip it.

pkg/kapp/cmd/app/deploy.go
100mik: I believe we could move these validations to the `Run()`function  as it returns an error and let the command initialisation have
```go
RunE:    func(_ *cobra.Command, _ []string) error { return o.Run() }
```

pkg/kapp/resources/resources.go
```diff
@@ -293,7 +302,7 @@ func (c *ResourcesImpl) Update(resource Resource) (Resource, error) {
 	return NewResourceUnstructured(*updatedUn, resType), nil
 }
 
-func (c *ResourcesImpl) Patch(resource Resource, patchType types.PatchType, data []byte) (Resource, error) {
+func (c *ResourcesImpl) Patch(resource Resource, patchType types.PatchType, data []byte, dryRun bool) (Resource, error) {
```
100mik: I wonder if we  should add something like a struct `resource.PatchOpts` and pass that to the function instead.
Something like 
```go
type PatchOpts struct {
    dryRun bool
}
```
This allows us to add options in the future easily without changing the function definition. 🤔

redbaron: Done

test/e2e/config_test.go
```diff
@@ -313,6 +313,8 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: test-cm
+  annotations:
+    ann1: val1
```
redbaron, line 319: I've added this to show that not all annotations stripped when object is passed to rebase rules

test/e2e/config_test.go
```diff
@@ -351,15 +353,18 @@ data:
 			"values": asYAML(t, map[string]interface{}{
 				"existing": func() interface{} {
 					raw := cm.Raw()
+					removeHistory(raw)
```
redbaron, line 358: Consequence of https://github.com/vmware-tanzu/carvel-kapp/pull/392/files#r768970039  is that rebase rules don't see history in objects passed to them. I've adjusted test to reflect it.

redbaron:
Hm, all tests are passing locally on Docker Desktop managed 1.21 kubernetes cluster, but  `TestSSAUpdateSimple`  fails in CI.

pkg/kapp/cmd/app/apply_flags.go
redbaron: Done

pkg/kapp/cmd/app/deploy.go
redbaron: Moved to PreRunE

test/e2e/diff_test.go
redbaron: Done

100mik:
Looks like it is just the new test. Let me take a swing at it as well 🤔

cppforlife:
(will also circle back in beginning of Jan, happy holidays!)

redbaron:
> your company's legal contact did not review your signed contributor license agreement within the 14 day limit.

I don't understand, which company? Which address your bot contacted? I am contributing from my personal account.

praveenrewar:
> Hm, all tests are passing locally on Docker Desktop managed 1.21 kubernetes cluster, but TestSSAUpdateSimple fails in CI.

The tests are running on k8s v1.16 and v1.22, and I think the behaviour is a bit different on those.
For v1.22 you might wanna remove the `spec.selector` field in `kubectlChange` (I haven't looked at it in detail so I am not sure how that will effect the e2e test).

redbaron:
OK, it reached the stage where I'd like to propose it for a merge.

I added option to run E2E tests with SSA enabled and  test suite now passes with SSA on my 1.21 K8S,  except for tests with rebasing. I'll go through diff and add notes on  parts worth paying attention to.

`TestIgnoreFailingAPIServices` fails for me locally even on master, because error message text differs from what test expects.

There is a missing test for `--ssa-force` option, I'll add it in due course, but otherwise I consider initial implementation complete.
I'll also update  PR message to reflect current state shortly.

/cc @cppforlife

cppforlife:
cool cool. ill do a deep dive today/tomorrow on this @redbaron. other folks will also take a look...

cppforlife:
(still reading through the code... hoping to be done tomorrow)

pkg/kapp/cmd/app/deploy_flag_help_sections.go
```diff
@@ -17,6 +17,10 @@ var (
 		Title:       "Diff Flags:",
 		PrefixMatch: "diff",
 	}
+	SSAFlagGroup = cobrautil.FlagHelpSection{
+		Title:       "Server side apply Flags:",
```
cppforlife: nit: "Server side apply Flags:" => "Server Side Apply Flags:"

test/e2e/update_server_side_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	tmpFile := newTmpFile(string(yamlExpected), t)
+	defer os.Remove(tmpFile.Name())
+
+	expectedObj := corev1.Service{}
+
+	// Patch dry run returns merged object with all patch-file fields present
+	err = kubectl.RunWithOptsIntoJSON([]string{"patch", "svc", "redis-primary", "--patch-file", tmpFile.Name(), "--dry-run=client"},
+		RunOpts{IntoNs: true}, &expectedObj)
+	require.NoError(t, err)
```
cppforlife, line 126: seems like a complicated way of determining expected state. is there something simpler that can be done (aka more static/hard-coded into the test)?

test/e2e/update_server_side_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	require.Equal(t, expectedObj, inClusterObj)
+
```
cppforlife: nit: remove new line

test/e2e/kapp.go
```diff
@@ -96,10 +101,97 @@ func (k Kapp) RunWithOpts(args []string, opts RunOpts) (string, error) {
 	return stdoutStr, err
 }
 
+func (k Kapp) RunEmbedded(args []string, opts RunOpts) (string, error) {
+	if k.SSAEnabled {
+		args = enableSSA(args)
+	}
+	var stdoutBuf bytes.Buffer
+	//var stdout io.Writer = bufio.NewWriter(&stdoutBuf)
```
cppforlife: nit: remove

test/e2e/kapp.go
```diff
@@ -96,10 +101,97 @@ func (k Kapp) RunWithOpts(args []string, opts RunOpts) (string, error) {
...
+	err := command.Execute()
+	confUI.Flush()
+
+	if err != nil {
+		require.Truef(k.t, opts.AllowError, "Failed to successfully execute '%s': %v", k.cmdDesc(args, opts), err)
+	}
+	return stdoutBuf.String(), err
+}
+
+func enableSSA(args []string) []string {
```
cppforlife, line 153: nit: place these methods on `func (Kapp)` to scope them to this class (nobody else should be using them).

pkg/kapp/cmd/tools/ssa/ssa_flags.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+type SSAFlags struct {
+	SSAEnable        bool
+	SSAForceConflict bool
+	FieldManagerName string
+}
+
+func (s *SSAFlags) Set(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&s.SSAEnable, "ssa-enable", false, "Use server side apply")
```
cppforlife: ssa-enable => ssa (this is in line with other flags like --wait=true|false).

can still keep struct field as SSAEnable if you want -- i would have went for something like since it's all under SSAFlags.

```
type SSAFlags struct {
	Enabled        bool
	ForceConflicts bool
	FieldManagerName string
}
```

pkg/kapp/resources/resources.go
```diff
@@ -6,6 +6,7 @@ package resources
 import (
 	"context"
 	"fmt"
+	"github.com/k14s/kapp/pkg/kapp/cmd/tools/ssa"
```
cppforlife: resources pkg should not be importing things from cmd/* (commands depend on lower level packages, im guessing that's why you had to add ssa directory in tools.).

instead we should define either our own SSAOpts or have those fields on ResourcesImplOpts struct. SSAFlags struct in cmd/* can either embed this struct or just dup the fields.

pkg/kapp/resources/resources.go
```diff
@@ -305,9 +319,18 @@ func (c *ResourcesImpl) Patch(resource Resource, patchType types.PatchType, data
...
+
+	if opts.DryRun {
+		dryRunOpt = []string{"All"}
+	}
 
 	err = util.Retry2(time.Second, 5*time.Second, c.isGeneralRetryableErr, func() error {
-		patchedUn, err = resClient.Patch(context.TODO(), resource.Name(), patchType, data, metav1.PatchOptions{})
+		patchedUn, err = resClient.Patch(context.TODO(), resource.Name(), patchType, data, metav1.PatchOptions{
+			FieldManager: c.opts.SSAFlags.FieldManagerName,
+			Force:        c.opts.SSAFlags.ForceParamValue(patchType),
```
cppforlife: lets localize forceParamValue logic to this class since it's only used here.

pkg/kapp/cmd/app/factory.go
```diff
@@ -56,16 +59,17 @@ func FactoryClients(depsFactory cmdcore.DepsFactory, nsFlags cmdcore.NamespaceFl
 		CoreClient:          coreClient,
 		ResourceTypes:       resTypes,
 		IdentifiedResources: identifiedResources,
+		Resources:           resources,
```
cppforlife: looks like this line is dead code

pkg/kapp/cmd/app/factory.go
```diff
@@ -15,11 +16,12 @@ type FactorySupportObjs struct {
 	CoreClient          kubernetes.Interface
 	ResourceTypes       *ctlres.ResourceTypesImpl
 	IdentifiedResources ctlres.IdentifiedResources
+	Resources           ctlres.Resources
```
cppforlife: looks like this line is dead code

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -107,6 +107,6 @@ func (c DeleteOrphanStrategy) Apply() error {
 		return err
 	}
 
-	_, err = c.d.identifiedResources.Patch(c.res, types.JSONPatchType, patchJSON)
+	_, err = c.d.identifiedResources.Patch(c.res, types.JSONPatchType, patchJSON, ctlres.PatchOpts{DryRun: false})
```
cppforlife: no need to specify `DryRun: false`

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -218,36 +222,14 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
...
-		}
+	// Record last applied change on the latest version of a resource
+	recordHistoryPatch, err := savedResWithHistory.RecordLastAppliedResource(applyChange)
+	if err != nil {
+		return fmt.Errorf("Recording last applied resource: %s", err)
+	}
 
-		return true, nil
-	})
+	_, err = c.identifiedResources.Patch(savedRes, types.MergePatchType, recordHistoryPatch, ctlres.PatchOpts{DryRun: false})
```
cppforlife: i guess this is safe-ish to do because metadata.annotations will always be present because kapp adds identity annotation.

(it's worth noting that "patch" verb is required in permissions in addition to update.)

cppforlife: `DryRun: false` is not necessary

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -263,6 +245,26 @@ func (c AddPlainStrategy) Apply() error {
...
+`), ctlres.PatchOpts{DryRun: false})
+		if err != nil {
+			// TODO: potentially patch can fail if '"op": "test"' fails, which can happen if another
+			// controller changes managedFields. We
+			return err
+		}
+
+	}
+
 	return c.aou.recordAppliedResource(createdRes)
```
cppforlife, line 268: there is a gap of time between create succeeding and patch succeeding where other changes may be applied to the resource. recordAppliedResource on the result of patch which means that we would capture such changes as "correlated" to initial submission. im not sure if this matters at all when ssa is enabled, but definitely seems suspicious if we keep it. at the least this deserves a comment, explaining why its ok for ssa.

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -263,6 +245,26 @@ func (c AddPlainStrategy) Apply() error {
...
+	if c.aou.opts.ServerSideApply {
+		createdRes, err = c.aou.identifiedResources.Patch(createdRes, types.JSONPatchType, []byte(`
+[
+	{ "op": "test", "path": "/metadata/managedFields/0/manager", "value": "`+c.aou.opts.FieldManagerName+`" },
+	{ "op": "replace", "path": "/metadata/managedFields/0/operation", "value": "Apply" }
+]
+`), ctlres.PatchOpts{DryRun: false})
+		if err != nil {
+			// TODO: potentially patch can fail if '"op": "test"' fails, which can happen if another
+			// controller changes managedFields. We
```
cppforlife, line 262: comment got cut off...

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -295,7 +311,17 @@ type UpdatePlainStrategy struct {
 func (c UpdatePlainStrategy) Op() ClusterChangeApplyStrategyOp { return updateStrategyPlainAnnValue }
 
 func (c UpdatePlainStrategy) Apply() error {
-	updatedRes, err := c.aou.identifiedResources.Update(c.newRes)
+	var updatedRes ctlres.Resource
+	var err error
+
+	if c.aou.opts.ServerSideApply {
+		updatedRes, err = ctlres.WithIdentityAnnotation(c.newRes, func(r ctlres.Resource) (ctlres.Resource, error) {
+			resBytes, _ := r.AsYAMLBytes()
```
cppforlife: lets throw in err check. its unlikely to kick in but just for safety.

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -295,7 +311,17 @@ type UpdatePlainStrategy struct {
...
+			return c.aou.identifiedResources.Patch(r, types.ApplyPatchType, resBytes, ctlres.PatchOpts{DryRun: false})
+		})
+	} else {
+		updatedRes, err = c.aou.identifiedResources.Update(c.newRes)
```
cppforlife, line 326: should we be tryToResolveUpdateConflict doing when SSA is on?

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -295,7 +311,17 @@ type UpdatePlainStrategy struct {
 func (c UpdatePlainStrategy) Op() ClusterChangeApplyStrategyOp { return updateStrategyPlainAnnValue }
 
 func (c UpdatePlainStrategy) Apply() error {
-	updatedRes, err := c.aou.identifiedResources.Update(c.newRes)
+	var updatedRes ctlres.Resource
+	var err error
+
+	if c.aou.opts.ServerSideApply {
+		updatedRes, err = ctlres.WithIdentityAnnotation(c.newRes, func(r ctlres.Resource) (ctlres.Resource, error) {
```
cppforlife, line 318: it seems that we have two types of Patch operations: one where we are working with a resource, and one where we are working with some "minimal/raw" patch. instead of exposing WithIdentityAnnotation outside of IdentifiedResources resources class, lets keep existing Patch method to work with Resource class (instead of []byte) and change it so that it can add annotation on the fly. lets also add PatchRaw which does take []byte. this should result in keeping WithIdentityAnnotation as private.

pkg/kapp/diff/change_factory.go
cppforlife: to double check -- changes in NewChangeAgainstLastApplied are not really affecting SSA, right? (i might have to ask to separate them out into a diff PR).

pkg/kapp/diff/change_set.go
```diff
@@ -4,14 +4,23 @@
...
+
+const (
+	ExactChangeSetMode              = "Exact"
+	AgainstLastAppliedChangeSetMode = "AgainstLastApplied"
+	ServerSideApplyChangeSetMode    = "ServerSideApply"
+)
+
 type ChangeSetOpts struct {
-	AgainstLastApplied bool
+	Mode ChangeSetMode
```
cppforlife, line 23: after reading through flag "adjusting" above, i think we should just turns this into:

```
type ChangeSetOpts struct {
        ServerSideApply bool
	AgainstLastApplied bool
}
```

anywhere where we are doing a switch based on mode, we should just do:

```
switch {
case opts.ServerSideApply: ...
case opts.AgainstLastApplied: ...
default: ...
}
```

the reasoning for this, it seems that mode enum introduces quite a bit of gymanistics necessary to compute what it should be.

pkg/kapp/diff/resource_with_history.go
```diff
@@ -218,5 +211,11 @@ func (resourceWithoutHistory) removeAppliedResAnnKeysMods() []ctlres.ResourceMod
 			ResourceMatcher: ctlres.AllMatcher{},
 			Path:            ctlres.NewPathFromStrings([]string{"metadata", "annotations", debugAppliedResDiffFullAnnKey}),
 		},
+		// metadata.managedFields technically is not kapp history, but we want to strip it in same situations we are
+		// stripping kapp history, it is server side apply history after all
+		ctlres.FieldRemoveMod{
```
cppforlife, line 216: without this, was this breaking anything? (if so, just for ssa, or for non-ssa operation?)

pkg/kapp/resources/mod_string_map_append.go
```diff
@@ -29,6 +31,17 @@ func (t StringMapAppendMod) Apply(res Resource) error {
...
 }
 
+func (t StringMapAppendMod) AsPatchBytes() []byte {
+	obj := map[string]interface{}{}
+	kvs := make(map[string]interface{}, len(t.KVs))
+	for k, v := range t.KVs {
+		kvs[k] = v
+	}
+	unstructured.SetNestedField(obj, kvs, t.Path.AsStrings()...)
+	b, _ := json.Marshal(obj)
```
cppforlife: lets add `if err != nil { panic("Internal inconsistency") }`

cppforlife (commented):
looking pretty good. ive left some nit-picks, and a few "designy" changes to hopefully simplify a little bit of code.

redbaron:
Thanks for a quick review, I'll make requested changes shortly

pkg/kapp/cmd/app/deploy_flag_help_sections.go
redbaron: done

pkg/kapp/cmd/tools/ssa/ssa_flags.go
redbaron: done

pkg/kapp/resources/resources.go
redbaron: done by duping fields

pkg/kapp/resources/resources.go
redbaron: done.  I didn't test if Force flag on non Apply patch types breaks anything, so doing a safest option and passing it only on apply . It made logic a bit more convoluted, because we can't just pass `false` and every other case and has to pass `nil` instead.

pkg/kapp/clusterapply/add_or_update_change.go
redbaron: Even if `metadata.annotations` is not present MergePatch will create it. It would be a problem if we used JsonPatch.

pkg/kapp/clusterapply/add_or_update_change.go
redbaron: You are right. `tryToResolveUpdateConflict` recalculates Change from a new existing resource. In case of SSA only conflict we can have is when we are conflicting with another fieldManager, no amount of new diffing going to help here.

pkg/kapp/diff/resource_with_history.go
redbaron: Yes, with SSA enabled it was reporting diff on `managedFields` where it shouldn't

cppforlife:
@redbaron lmk, when to take a look at the PR again.

ebma16:
Hey guys,

Could you give me a brief update on the topic? Are there still plans to support server-side-apply or will there be no further work on this topic for the time being? 
I only ask because in my project I need to update some Kubernetes resources partially shared with Carvel as well as with other tools (e.g. Operator SDK) and the other tools support SSA.

praveenrewar:
Hi @ebma16! Unfortunately the maintainers don't have the bandwidth to work on this at the moment, and I am not sure when we will be able to prioritise it. We can definitely prioritise a review if you would be interested in submitting a PR or continuing on this one.

kapp [rebase rules](https://carvel.dev/kapp/docs/v0.62.x/config/#rebaserules) should still work well with other tools that support SSA. Could you share the specific problem that you are facing?

ebma16:
Hi @praveenrewar,
thank's for the update and your quick response! 

At the moment I have no problems with the kapp rebase rules. We are still in an early phase of the project and I just did a small test for the interaction between the Java-Operator-SDK (using SSA) and the kapp rebase rules (without SSA). So far I could not find any problems and everything worked. I was just curious about the current status when I read this thread ;)

I will continue with this constellation (kapp without SSA and other tools using SSA). If I notice any problems, I will definitely let you know and continue this thread!

praveenrewar:
Sure, keep us updated. Happy to chat on [carvel](https://kubernetes.slack.com/archives/CH8KCCKA5) slack channel too if you need help with anything.

PR #393, opened 2021-12-16
indicating if deletion step is waiting on finalizers being removed (#…

### LINKED ISSUE #324: Indicate if deletion step is waiting on finalizers being removed
Currently kapp does not show users if deletion step is waiting on finalizers. This should be showed in the progress log.

---
Vote on this request

sethiyash:
- Start indicating on progress log when deletion step is waiting for finalizers being removed.

### DIFF
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -58,7 +58,7 @@ func (c DeleteChange) IsDoneApplying() (ctlresm.DoneApplyState, []string, error)
 
 	if res.IsDeleting() && len(res.Finalizers()) > 0 {
 		return ctlresm.DoneApplyState{Done: false, Message: "waiting on finalizers being removed"},
-			[]string{fmt.Sprintf("Resource deletion step is waiting on following finalizers being removed: %s",
+			[]string{uiWaitMsgPrefix + fmt.Sprintf("Waiting on finalizers: %s",
 				strings.Join(res.Finalizers(), ","))}, nil
 	}
 
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -69,9 +69,7 @@ func (c DeleteChange) IsDoneApplying() (ctlresm.DoneApplyState, []string, error)
 		return ctlresm.DoneApplyState{Done: true, Successful: true}, nil, nil
 	}
 
-	state := ctlresm.NewDeleting(existingRes).IsDoneApplying()
-
-	return state, []string{uiWaitMsgPrefix + state.Message}, nil
+	return ctlresm.DoneApplyState{Done: false, Successful: true}, descMessage(existingRes), nil
 }
 
 type DeletePlainStrategy struct {
@@ -115,4 +113,12 @@ func (c DeleteOrphanStrategy) Apply() error {
 
 	_, err = c.d.identifiedResources.Patch(c.res, types.JSONPatchType, patchJSON)
 	return err
+}
+
+func descMessage(res ctlres.Resource) []string {
+	if res.IsDeleting() {
+		return []string{uiWaitMsgPrefix +
+			ctlresm.NewDeleting(res).IsDoneApplying().Message}
+	}
+	return []string{}
 }
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -8,11 +8,10 @@ import (
 	"fmt"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/types"
-
 	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
 	ctlresm "github.com/k14s/kapp/pkg/kapp/resourcesmisc"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
@@ -112,10 +111,6 @@ func (c DeleteOrphanStrategy) Apply() error {
 	return err
 }
 
-func (c DeleteChange) buildDescMsg(res ctlres.Resource, doneApplying bool) []string {
-	if !doneApplying && len(res.Finalizers()) > 0 {
-		return []string{uiWaitMsgPrefix + fmt.Sprintf("Waiting on finalizers: %s",
-			strings.Join(res.Finalizers(), ","))}
-	}
-	return []string{}
+func (c DeleteChange) buildDescMsg(res ctlres.Resource, isDoneApplying bool) []string {
+	return ctlresm.NewDeleting(res).BuildDescMsg(isDoneApplying)
 }
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -112,5 +112,8 @@ func (c DeleteOrphanStrategy) Apply() error {
 }
 
 func (c DeleteChange) buildDescMsg(res ctlres.Resource, isDoneApplying bool) []string {
-	return ctlresm.NewDeleting(res).BuildDescMsg(isDoneApplying)
+	if res.IsDeleting() {
+		return ctlresm.NewDeleting(res).BuildDescMsg(isDoneApplying)
+	}
+	return []string{}
 }
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -112,11 +112,8 @@ func (c DeleteOrphanStrategy) Apply() error {
 }
 
 func buildDescMsg(res ctlres.Resource, isDoneApplying bool) []string {
-	if !isDoneApplying && len(res.Finalizers()) > 0 {
-		if deletingObj := ctlresm.NewDeleting(res); deletingObj != nil {
+	if deletingObj := ctlresm.NewDeleting(res); deletingObj != nil && !isDoneApplying{
 			return []string{uiWaitMsgPrefix + deletingObj.IsDoneApplying().Message}
-		}
-		return []string{}
 	}
 	return []string{}
 }
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -8,10 +8,11 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/types"
+
 	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
 	ctlresm "github.com/k14s/kapp/pkg/kapp/resourcesmisc"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


100mik (approved):
LGTM!

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -56,6 +56,11 @@ func (c DeleteChange) IsDoneApplying() (ctlresm.DoneApplyState, []string, error)
 		return ctlresm.DoneApplyState{Done: true, Successful: true, Message: "Resource orphaned"}, nil, nil
 	}
 
+	if res.IsDeleting() && len(res.Finalizers()) > 0 {
+		return ctlresm.DoneApplyState{Done: false, Message: "waiting on finalizers being removed"},
+			[]string{"Resource deletion step is waiting on finalizers being removed"}, nil
```
redbaron: It would be useful if list of remaining finalizers was printed

sethiyash: Sure, That's a great input. Will make the required changes to start printing them.

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -56,6 +56,11 @@ func (c DeleteChange) IsDoneApplying() (ctlresm.DoneApplyState, []string, error)
 		return ctlresm.DoneApplyState{Done: true, Successful: true, Message: "Resource orphaned"}, nil, nil
 	}
 
+	if res.IsDeleting() && len(res.Finalizers()) > 0 {
```
rohitagg2020: Can we add some test case for it?

sethiyash: Rohit looks like we don't store progress log somewhere. Until the finalizer is not removed the progress log get stuck in between and once we apply a patch to remove finalizer then I think there is no way to get the progress log once the resource deletion is complete.

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -56,6 +56,12 @@ func (c DeleteChange) IsDoneApplying() (ctlresm.DoneApplyState, []string, error)
...
+		return ctlresm.DoneApplyState{Done: false, Message: "waiting on finalizers being removed"},
+			[]string{fmt.Sprintf("Resource deletion step is waiting on following finalizers being removed: %s",
+				strings.Join(res.Finalizers(), ","))}, nil
```
cppforlife: can you include example output of how this looks in this thread? we might want to tweak wording a little bit.

sethiyash: Yaml File - 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: mymap
  finalizers:
    - kubernetes
```
Example Output while trying to delete - 
```
Changes

Namespace  Name   Kind       Conds.  Age  Op      Op st.  Wait to  Rs       Ri  
default    mymap  ConfigMap  -       7m   delete  -       delete   ongoing  Deleting  

Op:      0 create, 1 delete, 0 update, 0 noop, 0 exists
Wait to: 0 reconcile, 1 delete, 0 noop

Continue? [yN]: y

4:25:22PM: ---- applying 1 changes [0/1 done] ----
04:25:22PM: info: Resources: TODO resource 'configmap/mymap (v1) namespace: default' is already deleting
4:25:22PM: delete configmap/mymap (v1) namespace: default
4:25:22PM: ---- waiting on 1 changes [0/1 done] ----
4:25:22PM: ongoing: delete configmap/mymap (v1) namespace: default
4:25:22PM: Resource deletion step is waiting on following finalizers being removed: kubernetes
4:32:22PM: ---- waiting on 1 changes [0/1 done] ----
4:32:22PM: ongoing: delete configmap/mymap (v1) namespace: default
4:32:22PM: Resource deletion step is waiting on following finalizers being removed: kubernetes
...
```

cppforlife: I'd like to see something like this:

```
4:32:22PM: ongoing: delete configmap/mymap (v1) namespace: default
4:32:22PM:  ^ Waiting on finalizers: kubernetes
```

i believe we have some code already with ` ^ ...` but i dont recall why it wouldnt kick in here.

redbaron: wont this message interleave when multiple resources are awaited?

sethiyash: @cppforlife we have code with ` ^ ...` but that is getting used for retryable error and in `buildParentDescMsg` no where else so I have changed the error message using `^` prefix.

cppforlife:
- https://github.com/vmware-tanzu/carvel-kapp/blob/dfb5fcad790c4dd0b729a604e75b55ef82b2bb63/pkg/kapp/clusterapply/waiting_changes.go#L71 is a place where descMsgs are returned to be printed
- https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/clusterapply/delete_change.go#L51 should return such description (does not today)
- converged_resource.go does return "^ waiting ..." in some cases. i would expect delete_change could do something similar.

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -8,10 +8,11 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/types"
+
```
cppforlife: nit: extra newline

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -110,3 +111,11 @@ func (c DeleteOrphanStrategy) Apply() error {
 	_, err = c.d.identifiedResources.Patch(c.res, types.JSONPatchType, patchJSON)
 	return err
 }
+
+func (c DeleteChange) buildDescMsg(res ctlres.Resource, doneApplying bool) []string {
+	if !doneApplying && len(res.Finalizers()) > 0 {
+		return []string{uiWaitMsgPrefix + fmt.Sprintf("Waiting on finalizers: %s",
+			strings.Join(res.Finalizers(), ","))}
```
cppforlife: nit: use `, ` (comma space)

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -110,3 +111,11 @@ func (c DeleteOrphanStrategy) Apply() error {
 	_, err = c.d.identifiedResources.Patch(c.res, types.JSONPatchType, patchJSON)
 	return err
 }
+
+func (c DeleteChange) buildDescMsg(res ctlres.Resource, doneApplying bool) []string {
```
cppforlife: in addition to adding code here, take a look at https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/resourcesmisc/deleting.go as well. it needs to also contain similar code. and once they are both have same code... can we instantiate Deleting object here to reuse it.

pkg/kapp/clusterapply/delete_change.go
sethiyash: Done

pkg/kapp/clusterapply/delete_change.go
sethiyash: Done

pkg/kapp/clusterapply/delete_change.go
sethiyash: Done

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -110,3 +110,10 @@ func (c DeleteOrphanStrategy) Apply() error {
 	_, err = c.d.identifiedResources.Patch(c.res, types.JSONPatchType, patchJSON)
 	return err
 }
+
+func (c DeleteChange) buildDescMsg(res ctlres.Resource, isDoneApplying bool) []string {
```
rohitagg2020: Either `deleteChange` or `resource` is sufficient. We need not pass both of them.

pkg/kapp/clusterapply/delete_change.go
sethiyash: Doing this way what I observed is at this point in time resource doesn't get marked as deleted and hence `resource.IsDeleting()`  returns false and that's why `NewDeleting(res)`  method returns nil.

Problem:
Deploying a resource with finalizer and trying to delete it first time this message `waiting on finalizers ...`  doesn't come up as when doing kapp delete  first time resource doesn't get marked as deleted and hence `NewDeleting(res)`  method returns `nil` and hence it doesn't call deleting method to print the waiting message.  (while on doing kapp delete second time waiting on finalizers message shows up)

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -65,7 +65,7 @@ func (c DeleteChange) IsDoneApplying() (ctlresm.DoneApplyState, []string, error)
 		return ctlresm.DoneApplyState{}, nil, err
 	}
 
-	return ctlresm.DoneApplyState{Done: !exists, Successful: true}, nil, nil
+	return ctlresm.DoneApplyState{Done: !exists, Successful: true}, buildDescMsg(res, !exists), nil
```
cppforlife: instead of this, it should be something like:

```
if !exists {
  return ctlresm.DoneApplyState{Done: true, Successful: true}, nil, nil
}

return ctlresm.NewDeleting(existingRes).IsDoneApplying(), nil, nil
```

make sure to double check that message shows up in correct place for both cases.

sethiyash: @cppforlife , While doing `ctlresm.NewDeleting(existingRes).IsDoneApplying()` we need to check for `ctlresm.NewDeleting(existingRes)` is returning non nil. At this stage resource doesn't get marked as deleting and `ctlresm.NewDeleting(existingRes)` returns nil.

sethiyash: @cppforlife  `ctlresm.NewDeleting(existingRes)` is returning nil for some resource and result in nil pointer reference. 

Observed when deploying a Service resource (without finalizer) then `Service` and its `Endpoint` resource get deleted and `exists` come `false` for them but `EndpointSlice` do exist and `deletionTimestamp` is not present in `existingRes` of `EndpointSlice` so we have to add the condition to check for `res.IsDeleting()` for such usecases. 

I have started doing this in `descMessage()`

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -110,3 +114,11 @@ func (c DeleteOrphanStrategy) Apply() error {
 	_, err = c.d.identifiedResources.Patch(c.res, types.JSONPatchType, patchJSON)
 	return err
 }
+
+func descMessage(res ctlres.Resource) []string {
+	if res.IsDeleting() {
```
sethiyash, line 119: Added this check after observing this behaviour  - 

When deploying a `Service` resource (without finalizer) then `Service` and its `Endpoint` resource get deleted and `exists` come false for them but `EndpointSlice` do exist and `deletionTimestamp` is not present in `existingRes` of `EndpointSlice` so we have to add the condition to check for `res.IsDeleting()` for such usecases.

Got [this](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#terminating) looks like `EndpointSlice` doesn't have something like `deletionTimestamp`

PR #399, opened 2021-12-22
updated go text pkg to v0.3.7 (trivy vulnerability)

sethiyash:
- Updated go text package (golang.org/x/text) to v0.3.7 as trivy reported vulnerability in v0.3.6


praveenrewar:
@sethiyash I think you might have missed running go mod vendor or building the package once.

praveenrewar (approved):

100mik:
Looks like there are no changes in the vendor files is that expected?
But I believe the `verify-no-dirty-files` script would have caught that.

praveenrewar:
> Looks like there are no changes in the vendor files is that expected?

I assumed that that's because it's part of Go itself. We can confirm to be on the safe side.

PR #402, opened 2022-01-03
Add app-change describe command and record diffs in app-change

### LINKED ISSUE #338: Add `app-change describe` command to allow the user to view diffs for deployed changes
As of today, `kapp app-change list -a app_name` lists the duration of changes deployed for an application and displays a summary of the operations performed. This feature could be extended to provide the user more insight by making the text diff for each change available.

The config maps used to track app-changes can be used to store text diffs as well. And these can be exposed to the user through an `app-change describe` command.

The end result might be similar to this:
```zsh
❯ kapp app-change list -a test                                                                                                                                                                                             ─╯
Target cluster 'https://127.0.0.1:57406' (nodes: minikube)

App changes

Name               Started At            Finished At           Successful  Description  
test-change-vx7z5  2021-10-06T09:12:36Z  2021-10-06T09:12:36Z  true        update: Op: 1 create, 0 delete, 1 update, 0 noop / Wait to: 2 reconcile, 0 delete, 0 noop  
test-change-hcwrt  2021-10-06T09:11:56Z  2021-10-06T09:11:56Z  true        update: Op: 2 create, 0 delete, 0 update, 0 noop / Wait to: 2 reconcile, 0 delete, 0 noop  

2 app changes

Succeeded

❯ kapp app-change describe -a test --app-change test-change-vx7z5

@@ create configmap/config-1-ver-2 (v1) namespace: default @@
  ...
  1,  1   data:
  2     -   foo1: bar
      2 +   foo: bar
  3,  3   kind: ConfigMap
  4,  4   metadata:
@@ update configmap/config-2 (v1) namespace: default @@
  ...
  8,  8         kind: ConfigMap
  9     -       name: config-1-ver-1
      9 +       name: config-1-ver-2
 10, 10     creationTimestamp: "2021-10-06T09:11:56Z"
 11, 11     labels:
```

**Acceptance Criteria**

- [x]  Text diffs for app changes should be serialised and stored in app-change config maps in addition to any other information that might add value.
- [x]  Expose command `app-change describe app_change_name` which presents the information for that particular change in an intuitive view.
- [ ]  Make a flag `--app-change-diff` available, which can let the user to specify that they do not want to store the diff for a particular change.

**Checks needed**

- [x] Text diffs should be truncated if greater than 0.5 MB in size keeping in mind `etcd` [limitations](https://github.com/kubernetes/kubernetes/issues/19781).
- [ ] Consider special cases that might cause noisy diffs. (Like resources with `update-strategy: skip`)

This will provide much needed observability into changes being applied by kapp in scenarios where a `kapp deploy` is automatically triggered periodically or due to an action.

---
Vote on this request

100mik:
TODO:
- [ ] Add flag which enables this feature (false by default for now)
- [x] Truncate diffs exceeding size limit
- [x] Add condition for when diffs are not stored
- [ ] Add tests

Addresses: #338

### DIFF
--- a/pkg/kapp/app/recorded_app_changes.go
+++ b/pkg/kapp/app/recorded_app_changes.go
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	isChangeLabelKey   = "kapp.k14s.io/is-app-change"
+	IsChangeLabelKey   = "kapp.k14s.io/is-app-change"
 	isChangeLabelValue = ""
 	changeLabelKey     = "kapp.k14s.io/app-change-app" // holds app name
 )
@@ -37,7 +37,7 @@ func (a RecordedAppChanges) List() ([]Change, error) {
 
 	listOpts := metav1.ListOptions{
 		LabelSelector: labels.Set(map[string]string{
-			isChangeLabelKey: isChangeLabelValue,
+			IsChangeLabelKey: isChangeLabelValue,
 			changeLabelKey:   a.appName,
 		}).String(),
 	}
@@ -69,7 +69,7 @@ func (a RecordedAppChanges) List() ([]Change, error) {
 func (a RecordedAppChanges) DeleteAll() error {
 	listOpts := metav1.ListOptions{
 		LabelSelector: labels.Set(map[string]string{
-			isChangeLabelKey: isChangeLabelValue,
+			IsChangeLabelKey: isChangeLabelValue,
 			changeLabelKey:   a.appName,
 		}).String(),
 	}
@@ -94,6 +94,7 @@ func (a RecordedAppChanges) Begin(meta ChangeMeta) (*ChangeImpl, error) {
 		StartedAt:   time.Now().UTC(),
 		Description: meta.Description,
 		Namespaces:  meta.Namespaces,
+		DiffChanges: meta.DiffChanges,
 	}
 
 	configMap := &corev1.ConfigMap{
@@ -101,7 +102,7 @@ func (a RecordedAppChanges) Begin(meta ChangeMeta) (*ChangeImpl, error) {
 			GenerateName: a.appName + "-change-",
 			Namespace:    a.nsName,
 			Labels: map[string]string{
-				isChangeLabelKey: isChangeLabelValue,
+				IsChangeLabelKey: isChangeLabelValue,
 				changeLabelKey:   a.appName,
 			},
 		},
--- a/pkg/kapp/app/recorded_app_changes.go
+++ b/pkg/kapp/app/recorded_app_changes.go
@@ -89,7 +89,17 @@ func (a RecordedAppChanges) DeleteAll() error {
 	return nil
 }
 
+const (
+	// 0.5 MB in bytes
+	diffChangesStringMaxSize = 524288
+)
+
 func (a RecordedAppChanges) Begin(meta ChangeMeta) (*ChangeImpl, error) {
+	// Ensure that information stored with app-changes respects size limit
+	if len(meta.DiffChanges)*4 > diffChangesStringMaxSize {
+		meta.DiffChanges = fmt.Sprintf("%s\n%s", meta.DiffChanges[:130000], "Diff truncated as it exceeded max size (0.5 MB) ...")
+	}
+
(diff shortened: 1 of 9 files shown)


pkg/kapp/app/recorded_app_changes.go
```diff
@@ -94,14 +94,15 @@ func (a RecordedAppChanges) Begin(meta ChangeMeta) (*ChangeImpl, error) {
 		StartedAt:   time.Now().UTC(),
 		Description: meta.Description,
 		Namespaces:  meta.Namespaces,
+		DiffChanges: meta.DiffChanges,
```
cppforlife, line 107: seems like it might be more appropriate to truncate here since changes classes know are closer to knowing how much to truncate and why (in theory there could be impl that instead of saving things into configmaps could be saving things into elsewhere and that may not need truncation)

pkg/kapp/clusterapply/change_set_view.go
```diff
@@ -45,6 +48,28 @@ func (v *ChangeSetView) Print(ui ui.UI) {
 	}
 }
 
+const (
+	// 0.5 MB in bytes
+	diffChangesStringMaxSize = 524288
+)
+
+func (v *ChangeSetView) DiffChangesString() string {
```
cppforlife: instead of defining a new method could users be using Print with a different UI (some kind of in memory ui). there is also a question of how much context to include in the diff. today i guess it's defaulted to 1 line of context i think. summary section might also be useful (but may be always turned on)?

pkg/kapp/cmd/appchange/describe.go
```diff
@@ -0,0 +1,81 @@
...
+	_, found := appChangeMap.Labels[ctlapp.IsChangeLabelKey]
+	if !found {
+		return fmt.Errorf("Not an app-change")
+	}
+
+	var configSpec ctlapp.ChangeMeta
+	err = json.Unmarshal([]byte(appChangeMap.Data[ctlapp.ChangeMetaKey]), &configSpec)
+	if err != nil {
+		return fmt.Errorf("Unmarshalling app-change spec: %s", err.Error())
+	}
```
cppforlife, line 73: we have following interface -- https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/interfaces.go#L26-L30. whats the alternative way of implementing above.

pkg/kapp/app/recorded_app_changes.go
100mik: Yep, that makes sense. Will make that change

pkg/kapp/clusterapply/change_set_view.go
100mik: It does make sense to store the summary at all times. 
You man to say that the Print statements will get buffered to this memoized UI, from where it can be flushed to the config map? I believe that will make including different parts of the output (like summary) easier.
I will explore that 🤔

pkg/kapp/cmd/appchange/describe.go
100mik: Oh I see, it will more sense to have a function added to the interface than to make the private variables public for this implementation. Will do!

100mik: We could query a list of Changes through the interface and then grab the appropriate Meta instead of querying with the core client directly. I see that the app label and app-change label are already being added as selectors there 🤔

100mik: Will play around with that approach

100mik: As of now the user is not asked to specify the app itself, but this approach will require that

100mik:
With the current implementation, the colouring of the diff is not recorded as well.
Maybe if we switch to the method where we have another IOBuffer being written to, that gets resolved as well.
Flow goes something like:
1. We have an in memory buffer.
2. We always print the diff summary to this buffer
3. We print the diff changes to the buffer depending on the value of a flag
4. We create the app change meta with the contents of the buffer

praveenrewar:
Closing it for now. We will reopen it when we start working on this again.

PR #403, opened 2022-01-03
Add initial app suffix to recorded_app

### LINKED ISSUE #146: add suffix/prefix to name of kapp created metadata configmaps to avoid configmap naming collisions
Add the ability to add a default prefix for all kapp application's state `configMap` to differentiate them from other `configMap` stored in the same namespace so it won't conflict with the application's `configMap` :smile_cat: 

- [x]  add suffix apps.k14s.io to config map
- [x]  add functionality for users to opt in to this feature
- [x]  manage backward compatibility 
- [x]  if an app is deployed with this feature opted in via the FQ name env variable, the app should continue to behave like the features has been opted in even if the env flag is unset for subsequent kapp actions on the same app.

---
_(edit: hoist solution for those who searched for the corresponding error message)_
👉🏻  TL;DR. https://carvel.dev/kapp/docs/v0.53.0/configmap-migration/

neil-hickey:
By default all new apps deployed via `kapp deploy` will have a configmap name <APP_NAME>.apps.k14s.io instead of <APP_NAME> currently.

- This PR includes a migration path for existing configmaps deployed via older kapp versions
- Includes a new environment variable: USE_OLD_CONFIGMAP_NAME that when set to `True` will maintain the current kapp configmap name / behavior on `deploy`.
- Relevant e2e tests
- One error case caveat as mentioned here : https://github.com/vmware-tanzu/carvel-kapp/pull/403#discussion_r781674108

Performance hits:
- ~~Each `deploy` command now checks initially if the configmap `foo` exists for an app named `foo`. If it does, renames / updates the configmap and stops.~~ Thanks to a suggestion from @praveenrewar , now this hit is only occur once in the Exists() function! 
- If the configmap does not exist, it continues as before
- ~~kapp `deploy`, `delete` and `inspect` will now check for the existence of the configmap given the new suffix (e.g. `foo.apps.k14s.io`) and if does not exist, checks for `foo`. This unfortunately means two hits to the kubernetes api for an app that may not exist at all in some cases.~~ As above, the Exists() function now performs this check / fall-back only once (still introduces one additional hit on each command)


test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,109 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: redis-config2
+data:
+  key: value
+`
+)
+
+func TestAppSuffix_CreateUpdateDelete(t *testing.T) {
```
praveenrewar: We don't have a hard and fast rule against using `undescores` but I think we follow PascalCase in kapp.
Also, I am thinking if we should use the existing CreateUpdateDelete test.

pkg/kapp/cmd/app/list.go
```diff
@@ -102,9 +104,11 @@ func (o *ListOptions) Run() error {
 			return err
 		}
 
+		appName := strings.TrimSuffix(item.Name(), app.AppSuffix)
```
praveenrewar: Should we move this inline?

praveenrewar (commented):
`a.name` -> `a.FullyQualifiedName()` while calling `NewRecordedAppChanges()` in 2 places.

test/e2e/app_suffix_test.go
neil-hickey: cool, funnily enough I did have it there. But it felt like it was testing two distinct things. 1 that tests that a deploy works as expected (resources created) and this one which specifically checks the config-map exists and has the suffix

pkg/kapp/cmd/app/list.go
neil-hickey: can do! Wanted to leave it like this so you could very clearly see what has to happen with this way of implementing this change.

praveenrewar: Or should we do this in the List function itself?

neil-hickey: Good call! I actually didn't see the List() call earlier, (still exploring the code). I will move it there, it makes more sense there

pkg/kapp/app/recorded_app.go
```diff
@@ -80,9 +82,24 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
 func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 	defer a.logger.DebugFunc("CreateOrUpdate").Finish()
 
-	configMap := &corev1.ConfigMap{
+	configMap, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
+	if err == nil {
+		a.renameConfigMap(configMap, a.fqName, a.nsName)
```
cppforlife: check err?

pkg/kapp/app/recorded_app.go
```diff
@@ -140,17 +157,21 @@ func (a *RecordedApp) mergeAppUpdates(cm *corev1.ConfigMap, labels map[string]st
 }
 
 func (a *RecordedApp) Exists() (bool, string, error) {
-	_, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
-	if err != nil {
-		if errors.IsNotFound(err) {
+	_, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
+	if err == nil {
+		return true, "", nil
+	} else if errors.IsNotFound(err) {
```
cppforlife: nit: no need for else, since above return, which would make this code a bit flatter.

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,146 @@
...
+	logger.Section("configmap with suffix exists and not marked as a kapp-app", func() {
+		fqName := name + app.AppSuffix
+
+		NewClusterResource(t, "configmap", fqName, env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", fqName, env.Namespace, kubectl)
+
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{IntoNs: true, AllowError: true, StdinReader: strings.NewReader(yaml1)})
+
+		require.Containsf(t, err.Error(), "kapp: Error:", "did not contain parseable app metadata")
```
neil-hickey: Something to note from the issue https://github.com/vmware-tanzu/carvel-kapp/issues/146 was my suggestion that:

```
Given: Existing application named: foo and configmap: foo.apps.k14s.io

if foo.apps.k14s.io already exists and not marked as a kapp app configmap, error and prevent the deploy continuing. Suggest to the user that the configmap should be renamed / use the environment variable USE_EXISTING_CONFIGMAP_NAME=true to prevent migration to the newly generated configmap name.
```

The existing behaviour of kapp is to fail with a "cannot parse" error if someone had manually created a configmap that matched the app name. I found it hard to find a good place for this error to be injected into the code-base such that it only happened during deploy, as kapp in other methods calls [setMeta()](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L227-L238) before [CreateOrUpdate()](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L80) even runs, therefore not even being able to check for this existence prior to migration.

pkg/kapp/app/recorded_app.go
```diff
@@ -80,9 +83,31 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
 func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 	defer a.logger.DebugFunc("CreateOrUpdate").Finish()
 
+	if os.Getenv("USE_EXISTING_CONFIGMAP_NAME") != "True" {
```
praveenrewar: I think we need to explicitly handle the scenario where `USE_EXISTING_CONFIGMAP_NAME` is set to `True`, i.e we need to use `a.name` instead of `a.fqname` if `USE_EXISTING_CONFIGMAP_NAME` is set to `True` while creating or updating the app.

pkg/kapp/app/recorded_app.go
```diff
@@ -140,17 +165,25 @@ func (a *RecordedApp) mergeAppUpdates(cm *corev1.ConfigMap, labels map[string]st
...
 func (a *RecordedApp) Exists() (bool, string, error) {
-	_, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
-	if err != nil {
+	_, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
+	if err == nil {
+		return true, "", nil
+	}
+
+	if errors.IsNotFound(err) {
+		_, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
```
praveenrewar: `Exists()` is always called before `Delete()` and `Rename()`, so should we save this information (that the app is new or old) somewhere and use it in `Delete()` and `Rename()` to avoid one additional call?

pkg/kapp/app/recorded_app.go
```diff
@@ -80,9 +83,31 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
 func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 	defer a.logger.DebugFunc("CreateOrUpdate").Finish()
 
+	if os.Getenv("USE_EXISTING_CONFIGMAP_NAME") != "True" {
+		configMap, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
+		if err == nil {
+			err = a.renameConfigMap(configMap, a.fqName, a.nsName)
```
praveenrewar: I think we should check if the old ConfigMap was created by kapp or not, otherwise we might end up renaming some user ConfigMap. 
This could happen in the following scenario:

1. User creates a ConfigMap named `xyz`.
2. User tries to create a kapp application named `xyz` with `USE_EXISTING_CONFIGMAP_NAME` being set to False.

Although it would error out after that as no meta data would be found in the ConfigMap, but the damage would have been done by then.

pkg/kapp/app/recorded_app.go
```diff
@@ -297,9 +359,16 @@ func (a *RecordedApp) BeginChange(meta ChangeMeta) (Change, error) {
 }
 
 func (a *RecordedApp) update(doFunc func(*Meta)) error {
-	change, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
+	change, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("Getting app: %s", err)
+		if errors.IsNotFound(err) {
+			change, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
```
praveenrewar: Similar to `Exists()`, we might be able to use `meta()` to reduce one additional call.

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+	logger.Section("does not migrate if USE_EXISTING_CONFIGMAP_NAME=True", func() {
+		os.Setenv("USE_EXISTING_CONFIGMAP_NAME", "True")
+
+		kubectl.RunWithOpts([]string{"apply", "-f", "-"},
+			RunOpts{IntoNs: true, StdinReader: strings.NewReader(existingApp)})
+		NewPresentClusterResource("configmap", name, env.Namespace, kubectl)
+
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+		NewPresentClusterResource("configmap", name, env.Namespace, kubectl)
```
praveenrewar: I believe this is working right now by creating an app with new ConfigMap name and because there is no resource associated with the old ConfigMap, otherwise it would give ownership error for the resources present in the old app.

pkg/kapp/app/recorded_app.go
neil-hickey: Perhaps the variable should be renamed as I would expect `USE_EXISTING_CONFIGMAP_NAME=True` to still work if I am `kapp deploying` a new app with the suffix.

If we make the behaviour of `USE_EXISTING_CONFIGMAP_NAME` that it will only work given the old name, maybe something like `USE_OLD_CONFIGMAP_NAME`

pkg/kapp/app/recorded_app.go
neil-hickey: I like that idea! Let me explore if that would be viable!

neil-hickey: Nice suggestion :) I ended up doing this, It reduces the calls by a bit! Thanks 🎉

pkg/kapp/app/recorded_app.go
neil-hickey: I went ahead and added the case where it's True and also renamed it.

pkg/kapp/app/recorded_app.go
```diff
@@ -80,9 +85,26 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
 func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
```
neil-hickey: @praveenrewar I pretty heavily refactored this function to be a bit more maintainable / small. Lemme know whatcha think!

test/e2e/app_suffix_test.go
praveenrewar: Should we add the hint in `setMeta()` itself?

pkg/kapp/app/recorded_app.go
```diff
@@ -80,9 +85,26 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
 	defer a.logger.DebugFunc("CreateOrUpdate").Finish()
 
+	if strings.ToLower(os.Getenv("USE_OLD_CONFIGMAP_NAME")) == "true" {
+		return a.createOrUpdate(a.name, labels)
+	}
+
+	configMap, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
+	if err == nil {
+		if _, ok := configMap.Labels[KappIsAppLabelKey]; !ok {
+			return fmt.Errorf("Unable to migrate, configmap '%s' is not marked as a kapp app", a.name)
```
praveenrewar: I think we can proceed on to creating the app with suffix in this case.

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml2)})
+		NewPresentClusterResource("configmap", name, env.Namespace, kubectl)
+
+		cleanUp()
+	})
+
+	logger.Section("rename", func() {
```
praveenrewar: I think we can use the same app (from deploy section) for rename and delete. Although it's more of a personal taste, but the tests would definitely be a tiny bit faster if we use the same app.

Also, should we include tests for `inspect` command?

pkg/kapp/app/recorded_app.go
praveenrewar: Looks really good now.

pkg/kapp/app/recorded_app.go
neil-hickey: Right!

test/e2e/app_suffix_test.go
neil-hickey: gotcha, I will make the tests consistent with what already exists which follows the pattern you outlined. In my mind they were separate tests and a cascading side effect felt weird. But it's just a logger section, so I am on-board with that change :)

test/e2e/app_suffix_test.go
neil-hickey: One consequence of having it there is the following:

Example Workflow:
1. `kubectl create configmap neil.apps.k14s.io` // pre-existing non-kapp configmap
2. `kapp deploy -f examples/jobs/successful-job.yml -a neil`

> kapp: Error: ConfigMap 'neil.apps.k14s.io' already exists and did not contain parseable app metadata: (hint: rename / delete this configmap or set the environment variable USE_EXISTING_CONFIGMAP_NAME=true to prevent migration)

BUT also the other command render this error, which doesn't feel right as the hint isn't really true. The env var will do nothing to help in these cases:

1. `kapp inspect -a neil`  
2. `kapp delete  -a neil`

> kapp: Error: ConfigMap 'neil.apps.k14s.io' already exists and did not contain parseable app metadata: (hint: rename / delete this configmap or set the environment variable USE_EXISTING_CONFIGMAP_NAME=true to prevent migration)

---

I am not sure what the "right" thing is here, the current behaviour is what I opted for outputting `hint: ConfigMap was overriden by another user?`. The problem lies in the fact all commands check that the app exists / read the metadata, so whatever error we throw in `meta()` will show up for all commands `deploy` , `inspect` and `delete`. Ideally this error only shows up during a `deploy` and not the other commands, but I don't think that can happen, as these other functions are always called before `CreateOrUpdate()`

praveenrewar: I am actually comfortable with the existing error as well, as it would we be very rare and we would also have it documented.

pkg/kapp/app/recorded_app.go
```diff
@@ -80,9 +85,24 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
 func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 	defer a.logger.DebugFunc("CreateOrUpdate").Finish()
 
+	if strings.ToLower(os.Getenv("USE_OLD_CONFIGMAP_NAME")) == "true" {
```
cppforlife: all envs vars should be prefix with program name to avoid any conflicts. e.g. KAPP_USE_OLD_CONFIGMAP_NAME

pkg/kapp/app/recorded_app.go
```diff
@@ -280,7 +364,7 @@ func (a *RecordedApp) LastChange() (Change, error) {
 }
 
 func (a *RecordedApp) BeginChange(meta ChangeMeta) (Change, error) {
-	change, err := NewRecordedAppChanges(a.nsName, a.name, a.coreClient).Begin(meta)
+	change, err := NewRecordedAppChanges(a.nsName, a.fqName, a.coreClient).Begin(meta)
```
cppforlife: when migration doesnt happen, shouldnt we still be using old name?

neil-hickey:
@cppforlife and I had a private chat, with some outcomes:

1. We probably want to have this be optional for the beginning rather than opt out. This also consumers to opt-in and report any unintended consequences before we rollout
2. In the context of a large number of Packages / PackageRepositories etc in kapp-controller, all the 'change' configmaps will be orphaned (as is currently the case with `kapp replace` however, is this desirable when the user hasn't opted into this manually? Kapp-controller will be released with this migration opinion baked in.
3. In the case of someone having an app named `foo.apps.k14s.io` today, the migration currently will convert this app name to being `foo` instead of maintaining it. Is there maybe an opportunity for a "migration: true" flag to indicate that an app has been migrated and thus should be ignored even if migration is requested?

---

Following up:
1. Let's go with optionality via a new ENV_VAR called `KAPP_MIGRATE_CONFIGMAP_NAMES`
2. All `-change` configmap names will remain WITHOUT the new suffix. The cost to migrate these is too expensive, and the confusion to have some with the new name and other with the old is also confusing. So the change configmaps shall remain named as is.
3. Still TBD

pkg/kapp/app/recorded_app.go
neil-hickey: Based on some conversations around the complexity of migrating the change configmaps, this PR opts for not change / renaming / migration the `change` configmaps to using the new suffix

neil-hickey:
Current state as of Jan 27th:

- This migration is hidden behind a new env_var: `KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES`
- If the user migrates an app by setting this env_var, they can turn it off, but some manual steps for renaming the configmap will be documented. This PR does not support automatic downgrades.
- The `change` configmaps will NOT be using the new app suffix, they will remain as before
- This PR was updated to allow migration for apps that already contained the `.apps.k14s.io` suffix - every migrated / new app deployed with `KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES=true` will have a new label: `kapp.k14s.io/is-configmap-migrated` added to the configmap 
---
**New Behaviours:**

- `kapp deploy`
```
If KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES is true:
    deploy will migrate the configmap
else:
    continues as before
```
- `kapp rename`
```
If KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES is true OR app was migrated previously
    app specified in --new-name will contain the new suffix
else:
    app specified in --new-name will not contain the new suffix
```
- `kapp delete`
```
If KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES is true
    will delete app assuming it's configmap contains the app suffix
else:
    will delete app assuming it's configmap does not contains the app suffix
```

- `kapp inspect`
```
Supports inspecting a migrated and non-migrated app without the env_var being set
```

pkg/kapp/app/apps.go
```diff
@@ -82,7 +85,13 @@ func (a Apps) list(additionalLabels map[string]string, nsName string) ([]App, er
...
 
 	for _, app := range apps.Items {
-		recordedApp := &RecordedApp{app.Name, app.Namespace, a.coreClient,
+		name := app.Name
+
+		if _, ok := app.Labels[KappIsConfigmapMigratedLabelKey]; ok {
+			name = strings.TrimSuffix(app.Name, AppSuffix)
+		}
+
+		recordedApp := &RecordedApp{name, name + AppSuffix, app.Namespace, false, a.coreClient,
```
praveenrewar: _isMigrated_ should be true if `KappIsConfigmapMigratedLabelKey` is present, right?

pkg/kapp/app/recorded_app.go
```diff
@@ -94,15 +99,59 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 		}.AsData(),
 	}
 
-	err := a.mergeAppUpdates(configMap, labels)
+	if strings.ToLower(os.Getenv("KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES")) == "true" {
+		migratedLabel := map[string]string{KappIsConfigmapMigratedLabelKey: KappIsConfigmapMigratedLabelKeyValue}
+
+		configMapWithSuffix, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
+		if err == nil && a.isKappApp(configMapWithSuffix) {
```
praveenrewar: If a config map with suffix exists and is not a kapp app, then we will get this err 
```bash
kapp: Error: Getting app: %!s(<nil>)
```
I think we can have the check for `a.isKappApp(configMapWithSuffix)` inside the `if err == nil` so that we do nothing if the configmap is not by created by kapp and just continue with it and get the usual err, which is,
```bash
kapp: Error: App 'foo' (namespace: default) backed by ConfigMap 'foo.apps.k14s.io' did not contain parseable app metadata:
  Parsing app metadata: unexpected end of JSON input (hint: ConfigMap was overriden by another user?)
```

pkg/kapp/app/recorded_app.go
```diff
@@ -94,15 +99,59 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
...
+				return fmt.Errorf("Updating app: %s", err)
+			}
+
+			return nil
+		} else if !errors.IsNotFound(err) {
+			return fmt.Errorf("Getting app: %s", err)
+		}
+
+		configmapWithoutSuffix, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
+		if err == nil && a.isKappApp(configmapWithoutSuffix) {
```
praveenrewar: If a config map without suffix exists and is not a kapp app, then I think we should continue on to creating a config map with suffix.

pkg/kapp/app/recorded_app.go
```diff
@@ -126,6 +175,25 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 	return nil
```
praveenrewar, line 216: Should we use setMeta() once the configmap is created?
We always use [app.UsedGVs](https://github.com/vmware-tanzu/carvel-kapp/blob/f523c713505022f4a02c43060d55a0b1042e3fdf/pkg/kapp/cmd/app/deploy.go#L103), after CreateOrUpdate, which would try to fetch the config map as meta doesn't exists yet. So we would be able to save one get call if we set the meta here. 
Or do we want to cache meta only when we try to fetch it , i.e. inside `meta()`
cc @cppforlife

pkg/kapp/app/recorded_app.go
```diff
@@ -178,29 +260,47 @@ func (a *RecordedApp) Delete() error {
...
 			return fmt.Errorf("App '%s' (namespace: %s) does not exist: %s%s",
-				a.name, a.nsName, err, a.appInDiffNsHintMsgFunc(a.name))
+				a.name, a.nsName, err, a.appInDiffNsHintMsgFunc(name))
 		}
+
 		return fmt.Errorf("Getting app: %s", err)
 	}
 
+	// use fully qualified name if app had been previously migrated
+	if a.isMigrated || strings.ToLower(os.Getenv("KAPP_USE_CONFIGMAP_SUFFIX")) == "true" {
```
praveenrewar: Should be `KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES`.

pkg/kapp/app/recorded_app.go
```diff
@@ -244,15 +353,24 @@ func (a *RecordedApp) meta() (Meta, error) {
 	}
 
 	app, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
```
praveenrewar: Should we do a check for `KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES` and do a Get on _a.fqName_ before doing a Get on _a.name_?

pkg/kapp/app/recorded_app.go
```diff
@@ -297,9 +415,16 @@ func (a *RecordedApp) BeginChange(meta ChangeMeta) (Change, error) {
...
-	change, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
+	name := a.name
+	if a.isMigrated {
+		name = a.fqName
+	}
+
+	change, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), name, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("Getting app: %s", err)
+		if errors.IsNotFound(err) {
```
praveenrewar: What happens if it's a different kind of error? Were you planning to add more changes here?

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+		NewMissingClusterResource(t, "configmap", existingName, env.Namespace, kubectl)
+		c := NewPresentClusterResource("configmap", existingName+app.AppSuffix, env.Namespace, kubectl)
+		require.Contains(t, c.Labels(), app.KappIsConfigmapMigratedLabelKey)
+
+		kapp.Run([]string{"delete", "-a", existingName})
+	})
+
+	// if a user has disabled the migration on an app that was already migrated
+	// the error we expect is that the app belongs to a different app
```
praveenrewar: _resources_ belong to a different app

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,179 @@
...
+	os.Setenv("KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES", "True")
+
+	logger.Section("with suffix and not marked as a kapp-app", func() {
+		NewClusterResource(t, "configmap", fqName, env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", fqName, env.Namespace, kubectl)
+
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{IntoNs: true, AllowError: true, StdinReader: strings.NewReader(yaml1)})
+
+		require.Containsf(t, err.Error(), "kapp: Error:", "did not contain parseable app metadata")
```
praveenrewar: We should first assert on err being non nil (using `require.Error()`). 
Also, we should test more than just `kapp: Error:`. Right now this will pass for all the errors.

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,179 @@
...
+
+		RemoveClusterResource(t, "configmap", fqName, env.Namespace, kubectl)
```
praveenrewar: We should have this inside of `cleanUp()` and remove the deletion using kapp as we will be creating only one configmap using kubectl, so it makes sense to remove it using kubectl only.

pkg/kapp/app/recorded_app.go
```diff
@@ -94,15 +99,59 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 		}.AsData(),
 	}
 
-	err := a.mergeAppUpdates(configMap, labels)
+	if strings.ToLower(os.Getenv("KAPP_EXPERIMENTAL_FQ_CONFIGMAP_NAMES")) == "true" {
```
praveenrewar: Do we want to have `EXPERIMENTAL` in there? As this might stop a lot of users or make them think twice before they use it. 
(Or is this a way of hinting them to read the docs before they use it.)

praveenrewar (commented):
I really liked that we have made the migration optional now.

pkg/kapp/app/recorded_app.go
neil-hickey: So I put `EXPERIMENTAL` because I wanted to signal that this feature has consequences that require a manual rollback if they need to downgrade. And also to really examine the docs before they use it.

I do agree though that it will probably stop users from using it as freely as if the word was not part of the flag.

pkg/kapp/app/recorded_app.go
neil-hickey: The code within `createOrUpdate` is as is from before. So no modifications. I am all for things that reduce the number of k8s api calls, but it feels like something slightly out of scope of these changes.

If you and @cppforlife decide it is a good idea, maybe I can create an accompanying PR / update this PR if the decision is made in the mean time. I am trying to reduce as much as I can the scope / changes in this PR as its been dragging on a bit

praveenrewar: I agree that it can be separate PR (if we plan on implementing it).

pkg/kapp/app/recorded_app.go
praveenrewar: Now that we are continuing [here](https://github.com/vmware-tanzu/carvel-kapp/pull/403#discussion_r795191631), we might want to add this check as well, otherwise it will error out saying configmap is not parseable.

pkg/kapp/app/recorded_app.go
praveenrewar: It does sound like a good idea to hint them about the docs, I am just thinking that this is something that we would want all the users to use at some point of time, so `EXPERIMENTAL` won't make much sense then (unless we are planning to introduce a new environment variable at a later stage).

pkg/kapp/app/recorded_app.go
```diff
@@ -23,8 +26,10 @@ const (
 )
 
 type RecordedApp struct {
-	name   string
-	nsName string
+	name       string
+	fqName     string
+	nsName     string
+	isMigrated bool
```
praveenrewar: I have been thinking about it a bit... Now that we have a label present in the configmap, do we just want to have a method for recorded app called `isMigrated()` that checks if the label is present or not? As it seems like we are storing the same information at 2 places.

neil-hickey: I was thinking all the same lines. In an earlier conversation we optimized `Rename` / `Delete` and maybe others with the `isMigrated` bool, which was populated in the `Exists()` function.

`Rename` for example would now need to get the metadata and check the label. It doesn't have access to the configmap prior to the function call. So I don't think (though I could be wrong) we can get rid of the `isMigrated` bool, as it cannot be replaced with a check for the labels in all cases (or at least the value it was providing would no longer be present).

Let me know if that doesn't make sense / I am misreading your question.

REF: https://github.com/vmware-tanzu/carvel-kapp/pull/403/files/631b31756e11cafc3d1bcd01b44901f69cc08aa5#diff-c4cadf071a66c0e934e58c35eb0a9116fee4c744abd14621dc1eeebe7f5e9fb8R267-R271

praveenrewar: > Rename for example would now need to get the metadata and check the label

True, and I think the same would need to happen for `Delete`. So `Exists()` might be another place where might want to save the meta data and we should be able to reduce one call. (We can have a seperate PR if we decide to do it).

praveenrewar: > do we just want to have a method for recorded app called isMigrated() that checks if the label is present or not? As it seems like we are storing the same information at 2 places.

Now I am having similar thoughts about `fqName` as well xD
(Maybe I am just overthinking)

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,192 @@
...
+	logger.Section("rename", func() {
+		kapp.Run([]string{"rename", "-a", name, "--new-name", newName})
+		NewMissingClusterResource(t, "configmap", name, env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", newName+app.AppSuffix, env.Namespace, kubectl)
+	})
+
+	logger.Section("delete", func() {
+		kapp.Run([]string{"delete", "-a", newName})
+
+		cleanUp()
```
praveenrewar: I am wondering if this is required?

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	os.Setenv("KAPP_FQ_CONFIGMAP_NAMES", "True")
+
+	logger.Section("without suffix and not marked as a kapp-app", func() {
+		NewClusterResource(t, "configmap", name, env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", name, env.Namespace, kubectl)
```
praveenrewar: This seems like having a test for a test (as `NewClusterResource` would error out if the resource is not created)

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+		require.NoError(t, err)
+
+		kapp.Run([]string{"delete", "-a", fqName})
+		cleanUp()
+	})
+
+	logger.Section("with suffix and not marked as a kapp-app", func() {
+		NewClusterResource(t, "configmap", fqName, env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", fqName, env.Namespace, kubectl)
```
praveenrewar: Same as above. We can remove this.

test/e2e/app_suffix_test.go
```diff
@@ -0,0 +1,192 @@
...
+		NewClusterResource(t, "configmap", name, env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", name, env.Namespace, kubectl)
+
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{IntoNs: true, AllowError: true, StdinReader: strings.NewReader(yaml1)})
+
+		require.NoError(t, err)
+
+		kapp.Run([]string{"delete", "-a", fqName})
+		cleanUp()
```
praveenrewar: Should we use `RemoveClusterResource` instead of cleanUp() as we are just trying to delete the ConfigMap created with `NewClusterResource`?

praveenrewar (commented):
Looks really good to me (except for the small nits)

pkg/kapp/app/recorded_app.go
```diff
@@ -143,6 +217,17 @@ func (a *RecordedApp) mergeAppUpdates(cm *corev1.ConfigMap, labels map[string]st
 }
 
 func (a *RecordedApp) Exists() (bool, string, error) {
+	if a.isMigrationEnabled() {
```
sethiyash, line 269: Looks like here we are making an assumption that once users enable `KAPP_FQ_CONFIGMAP_NAMES`  env variable it will never disable it again. If this happens unwillingly user will unable to find apps.

sethiyash:
@neil-hickey observed one thing - 
`kapp list` shows kapp app name when `KAPP_FQ_CONFIGMAP_NAMES` env var is set to true. But now as env var is set our kapp configmap will be created with `<app-name>.apps.k14s.io` so should we start showing fully qualified name when env var is enabled instead of showing kapp app name here?

Because if the user wants to see the configmap created by kapp they will unable to locate it by doing - 
`kubectl get cm/<app-name> -o yaml` and there will be need to figure suffix and add it by themself - `kubectl get cm/<app-name>.apps.k14s.io -o yaml`

pkg/kapp/app/recorded_app.go
neil-hickey: yep that's correct, we have a set of manual steps they can take to downgrade if they choose that path. If they don't provide the env_var they will not be able to find their migrated apps. I think this is a reasonable behaviour and we ought to be really explicit that you can't and shouldn't swap back and forth between them, we also don't want them to be able to get into a bad state with the migrated app if they forget the variable.

neil-hickey:
> @neil-hickey observed one thing - `kapp list` shows kapp app name when `KAPP_FQ_CONFIGMAP_NAMES` env var is set to true. But now as env var is set our kapp configmap will be created with `<app-name>.apps.k14s.io` so should we start showing fully qualified name when env var is enabled instead of showing kapp app name here?
> 
> Because if the user wants to see the configmap created by kapp they will unable to locate it by doing - `kubectl get cm/<app-name> -o yaml` and there will be need to figure suffix and add it by themself - `kubectl get cm/<app-name>.apps.k14s.io -o yaml`

Right, this is the behaviour we want(?):

The "app name" will always be the user supplied one. This PR is intended to keep the internal implementation of the configmap name with suffix hidden. I think this becomes especially important when we force the users down this path of migration, all CI/CD systems that run commands based on a name would break, and would be forced to include our appSuffix as part of it.

Both options have their downsides, I personally would rather users who really want the configmap using `kubectl get cm/<app-name> -o yaml` to now understand there is a suffix (as this is more advanced) versus standard use cases of kapp deploy where the app name is the most important thing.

praveenrewar (approved):
LGTM

pkg/kapp/app/recorded_app.go
sethiyash: okay got it. Sounds Good.

sethiyash:
> 



> > @neil-hickey observed one thing - `kapp list` shows kapp app name when `KAPP_FQ_CONFIGMAP_NAMES` env var is set to true. But now as env var is set our kapp configmap will be created with `<app-name>.apps.k14s.io` so should we start showing fully qualified name when env var is enabled instead of showing kapp app name here?
> > Because if the user wants to see the configmap created by kapp they will unable to locate it by doing - `kubectl get cm/<app-name> -o yaml` and there will be need to figure suffix and add it by themself - `kubectl get cm/<app-name>.apps.k14s.io -o yaml`
> 
> Right, this is the behaviour we want(?):
> 
> The "app name" will always be the user supplied one. This PR is intended to keep the internal implementation of the configmap name with suffix hidden. I think this becomes especially important when we force the users down this path of migration, all CI/CD systems that run commands based on a name would break, and would be forced to include our appSuffix as part of it.
> 
> Both options have their downsides, I personally would rather users who really want the configmap using `kubectl get cm/<app-name> -o yaml` to now understand there is a suffix (as this is more advanced) versus standard use cases of kapp deploy where the app name is the most important thing.

@neil-hickey This all sounds fair to me.

sethiyash (approved):

pkg/kapp/app/apps.go
```diff
@@ -83,7 +86,15 @@ func (a Apps) list(additionalLabels map[string]string, nsName string) ([]App, er
 	}
 
 	for _, app := range apps.Items {
-		recordedApp := &RecordedApp{app.Name, app.Namespace, app.ObjectMeta.CreationTimestamp.Time, a.coreClient,
+		name := app.Name
+		isMigrated := false
+
+		if _, ok := app.Labels[KappIsConfigmapMigratedLabelKey]; ok {
```
100mik: Should this be an annotation instead?
I think we wanna use labels when we are going to use the label as a selector while querying those resources, (example `kapp.k14s.io/is-app` used as selector when user runs `kapp ls`)
However, since we are comparing after getting the resource, I think it should be an annotation.

100mik (commented):
Was skimming through the PR. Something that caught my eye 🚀

pkg/kapp/app/apps.go
neil-hickey: this sounds good, I think we ought to use annotations also

neil-hickey: Thanks @100mik ! Good call, changed to using annotations. Looks like this is our first instance of an annotation, so I kept the label name the same just as an annotation instead

neil-hickey:
hey @praveenrewar I updated to using annotations, lemme know how you all wanna deal with merging. Should I? Do you all wanna do it, etc. I am working on docs today, so will update this PR when those are in good shape (maybe that's the signal to merge?)

---

update: docs pr https://github.com/vmware-tanzu/carvel/pull/369

pkg/kapp/app/apps.go
100mik: Sounds perfect! Thanks

100mik (approved):
LGTM!

sethiyash:
Looks Good!

praveenrewar:
Hey @neil-hickey, we were thinking about doing some upgrade testing before merging the PR and wanted to know if you had already tested some upgrade scenarios (if so, would it be possible for you to share a list of those?).

neil-hickey:
> Hey @neil-hickey, we were thinking about doing some upgrade testing before merging the PR and wanted to know if you had already tested some upgrade scenarios (if so, would it be possible for you to share a list of those?).

Hi @praveenrewar ! No, I didn't create an upgrade testing scenario. I opted for the route of a feature flag where I turned it off, deployed and then turned it on to test for the upgrade scenario.

renuy:
@neil-hickey , @praveenrewar , @cppforlife There was suggestion in the community meeting today on the suffix being used.
Do we want to move away from apps.k14s.io

PR #405, opened 2022-01-05
Use memoization to reduce api calls to get list of ResourceTypes

### LINKED ISSUE #394: Perf Enhancement
We need to look into what are the possible ways to improve the kapp performance.

To Do Task:
1. Figure out all the places we are making calls to k8s server and if we can reduce them.
2. Another possible place:
When kapp deploy is using configmaps (vs. kapp deploy -a label:my-app-name=xxx), is the usedGVs property in the main configmap an optimisation so that when querying based on labelKey/labelValue from the main configmap to identfiy app resources, it doesn't have to query every single resource type in the whole cluster, and can at least restrict it to just resources within those api groups? 
usedGVs is currently used to decide which apis we can ignore if they are not available (sometimes some api servers die, or something else happens making it not possible to list resources). we still optimistically query all types, but, we may evolve that behaviour to improve perf over time.

- [x] Use memoization to reduce redundant api calls to get the Resources supported by the K8s cluster.
- [x] In case of kapp controller, we should make the kapp inspect disable by default. Individual APP CR's can opt-in for the kapp inspect if they want. [carvel-kapp-controller#467](https://github.com/vmware-tanzu/carvel-kapp-controller/issues/467)
- [x] Service Account Scoping - to see what permissions does SA have and how APP CR's are consuming them in kapp controller. - [ry]not doing, see comment by @rohitagg2020 below
- [ ] Go through all the API calls made by kapp deploy/delete/inspect and see if we can reduce them.
- [ ] While querying the resources by label selector, we should query only the GV's which have been used earlier to deploy rather than querying all the GV's.
- [x] Match resource type associated resources with API group and resource, thus reducing API calls in waiting stage

praveenrewar:


100mik (approved):
LGTM!

rohitagg2020 (approved):

sethiyash:
Looks Good.

100mik:
Mergin it 🚀

PR #407, opened 2022-01-12
Add GK scoping to enhance performance

### LINKED ISSUE #394: Perf Enhancement

100mik:
This is primarily intended to be used consciously by AppCR/Package authors (defaults still to be decided) or by `kapp-controller` while reconciling package  repositories, to reduce the number of redundant calls being made by `kapp`

To explore:
- [x] Can we better handle how we pass the GKs we wanna scope to IdentifiedResources?
- [x] The current implementation does not remove GKs which are no longer under use, this needs to be handled better.
- [x] Backwards compatibility with deployed apps?
- [x] See how `PartialResourceRef` can be used to track Kinds as well (Extending `PartialResourceRef` will lead to changes touching too many files)


pkg/kapp/app/recorded_app.go
```diff
@@ -77,6 +77,51 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+	meta, err := a.meta()
+	if err != nil {
+		return nil, err
+	}
+
+	return meta.UsedGVKs, nil
+}
+
+func (a *RecordedApp) UpdateUsedGVKs(gvks []schema.GroupVersionKind, merge bool) ([]schema.GroupVersionKind, error) {
+	gvksByGVK := map[schema.GroupVersionKind]struct{}{}
```
rohitagg2020: nit: We can call it uniqGVKsMap

pkg/kapp/cmd/app/deploy.go
```diff
@@ -133,8 +134,13 @@ func (o *DeployOptions) Run() error {
...
 	}
 
+	usedGVKs, err := app.UpdateUsedGVKs(NewUsedGVKsScope(newResources).GVKs(), true)
+	if err != nil {
+		return err
+	}
+
 	existingResources, existingPodRs, err := o.existingResources(
-		newResources, labeledResources, resourceFilter, supportObjs.Apps)
+		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGVKs)
```
rohitagg2020: I think we can just pass app.UsedGVKs here because we are trying to get existing resources. We don't need to make function call to UpdateUsedGVKs in line 137.

pkg/kapp/resources/resource_types.go
```diff
@@ -265,3 +265,16 @@ func NonMatching(in []ResourceType, ref ResourceRef) []ResourceType {
 	}
 	return out
 }
+
+// TODO: Extend ResourceRef and PartialResourceRefd to allow GVK matching
+func MatchingAnyGVK(in []ResourceType, gvks []schema.GroupVersionKind) []ResourceType {
+	var out []ResourceType
+	for _, item := range in {
+		for _, gvk := range gvks {
+			if item.APIResource.Group == gvk.Group && item.APIResource.Version == item.APIResource.Version && item.Kind == gvk.Kind {
```
rohitagg2020: nit: Make it like `item.Kind == gvk.Kind && item.APIResource.Version == item.APIResource.Version && item.APIResource.Group == gvk.Group`. This will be faster than above one.

pkg/kapp/cmd/app/deploy.go
100mik: We need the function call on line 137 to track new GVKs that might have been introduced in the manifest.

pkg/kapp/app/recorded_app.go
100mik: I do not see why not, stuck to this because of [this](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L65) existing precedent

pkg/kapp/resources/resource_types.go
100mik: Yep! that makes sense. As of now I plan on extending the functionality [here](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/resources/refs.go) to match Kinds as well. Will keep it in mind then.

pkg/kapp/resources/resource_types.go
```diff
@@ -269,3 +269,16 @@ func NonMatching(in []ResourceType, ref ResourceRef) []ResourceType {
...
```
praveenrewar:
```suggestion
			if item.APIResource.Group == gvk.Group && item.APIResource.Version == gvk.Version && item.Kind == gvk.Kind {
```

pkg/kapp/app/interfaces.go
```diff
@@ -17,6 +17,8 @@ type App interface {
 	LabelSelector() (labels.Selector, error)
 	UsedGVs() ([]schema.GroupVersion, error)
 	UpdateUsedGVs([]schema.GroupVersion) error
+	UsedGKs() ([]schema.GroupKind, error)
+	UpdateUsedGKs([]schema.GroupKind, bool) ([]schema.GroupKind, error)
```
rohitagg2020: nit: We can return only error to be consistent with UpdateUsedGVs method. Wherever needed, we can do `app.UsedGKs()` to get the `[]schema.GroupKind`

pkg/kapp/app/recorded_app.go
```diff
@@ -77,6 +77,51 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+	gksByGK := map[schema.GroupKind]struct{}{}
+	var uniqGKs []schema.GroupKind
+
+	if merge {
+		usedGKs, err := a.UsedGKs()
+		if err != nil {
+			return nil, err
+		}
+
+		// Handle existing apps without cached GVKs
```
rohitagg2020: nit: all the comments still mention GVKs, we can call them GKs now.

pkg/kapp/app/recorded_app.go
```diff
@@ -77,6 +77,51 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+func (a *RecordedApp) UsedGKs() ([]schema.GroupKind, error) {
+	meta, err := a.meta()
+	if err != nil {
+		return nil, err
+	}
+
+	return meta.UsedGKs, nil
+}
+
+func (a *RecordedApp) UpdateUsedGKs(gvks []schema.GroupKind, merge bool) ([]schema.GroupKind, error) {
```
rohitagg2020: nit: we can call them as gks.

pkg/kapp/app/recorded_app.go
```diff
@@ -77,6 +77,51 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+	var uniqGKs []schema.GroupKind
+
+	if merge {
+		usedGKs, err := a.UsedGKs()
+		if err != nil {
+			return nil, err
+		}
+
+		// Handle existing apps without cached GVKs
+		// These apps can cache and scope to GVKs in subsequent deploys
```
praveenrewar:
```suggestion
		// Handle existing apps without cached GKs
		// These apps can cache and scope to GKs in subsequent deploys
```

pkg/kapp/cmd/app/deploy.go
```diff
@@ -178,6 +184,12 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	// Cache possibly untracked GVKs in existing resources (handles older apps)
```
praveenrewar:
```suggestion
	// Cache possibly untracked GKs in existing resources (handles older apps)
```

pkg/kapp/cmd/app/used_gvks_scope.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2020 VMware, Inc.
```
praveenrewar: Should we change the file name to `used_gk_scope.go`?

pkg/kapp/app/recorded_app.go
praveenrewar:
```suggestion
func (a *RecordedApp) UpdateUsedGKs(gks []schema.GroupKind, merge bool) ([]schema.GroupKind, error) {
```

pkg/kapp/app/recorded_app.go
```diff
@@ -77,6 +77,51 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+
+		for _, gk := range usedGKs {
+			if _, found := gksByGK[gk]; !found {
+				gksByGK[gk] = struct{}{}
+				uniqGKs = append(uniqGKs, gk)
+			}
+		}
+	}
+
+	for _, gk := range gvks {
```
praveenrewar:
```suggestion
	for _, gk := range gks {
```

pkg/kapp/cmd/app/deploy.go
```diff
@@ -293,6 +305,11 @@ func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 		// Prevent accidently overriding kapp state records
 		DisallowedResourcesByLabelKeys: []string{ctlapp.KappIsAppLabelKey},
 		LabelErrorResolutionFunc:       labelErrorResolutionFunc,
+
+		//Scope resource searching to UsedGVKs
```
praveenrewar:
```suggestion
		//Scope resource searching to UsedGKs
```

pkg/kapp/cmd/app/used_gvks_scope.go
rohitagg2020: nit: rename the file to `used_gks_scope.go`

pkg/kapp/resources/identified_resources_list_test.go
```diff
@@ -76,3 +76,4 @@ func (r *FakeResourceTypes) Find(ctlres.Resource) (ctlres.ResourceType, error) {
 	return ctlres.ResourceType{}, nil
 }
 func (r *FakeResourceTypes) CanIgnoreFailingGroupVersion(schema.GroupVersion) bool { return true }
+func (r *FakeResourceTypes) ScopeToUsedGVKs() bool                                 { return false }
```
praveenrewar: I think this is not required now.

pkg/kapp/resources/resource.go
```diff
@@ -19,7 +19,9 @@ import (
 
 type Resource interface {
 	GroupVersionResource() schema.GroupVersionResource
+	GroupVersionKind() schema.GroupVersionKind
```
rohitagg2020: Do we need this as we have started relying on `GroupKind` now.

pkg/kapp/resources/resource.go
```diff
@@ -138,6 +140,17 @@ func (r *ResourceImpl) GroupVersionResource() schema.GroupVersionResource {
 	return r.resType.GroupVersionResource
 }
 
+func (r *ResourceImpl) GroupVersionKind() schema.GroupVersionKind {
```
rohitagg2020: Do we need this as we have started relying on `GroupKind` now.

pkg/kapp/resources/resource_types.go
```diff
@@ -269,3 +269,16 @@ func NonMatching(in []ResourceType, ref ResourceRef) []ResourceType {
 	}
 	return out
 }
+
+// TODO: Extend ResourceRef and PartialResourceRefd to allow GVK matching
+func MatchingAnyGK(in []ResourceType, gks []schema.GroupKind) []ResourceType {
+	var out []ResourceType
+	for _, item := range in {
+		for _, gk := range gks {
+			if (GKResourceRef{gk}).Matches(item) {
```
praveenrewar, line 278: Maybe send _item.GroupVersionKind_ to `.Matches()`?

pkg/kapp/resources/resource.go
praveenrewar: Not required now.

pkg/kapp/resources/resource.go
```diff
@@ -138,6 +140,17 @@ func (r *ResourceImpl) GroupVersionResource() schema.GroupVersionResource {
...
+	return r.un.GroupVersionKind()
+}
```
praveenrewar: Not required now.

pkg/kapp/resources/resource.go
```diff
@@ -138,6 +140,17 @@ func (r *ResourceImpl) GroupVersionResource() schema.GroupVersionResource {
...
+
+func (r *ResourceImpl) GroupKind() schema.GroupKind {
+	return schema.GroupKind{
+		Group: r.APIGroup(),
+		Kind:  r.Kind(),
+	}
```
praveenrewar: Maybe use `r.un.GroupVersionKind().GroupKind()`?

test/e2e/crd_versions_with_gk_scoping_test.go
```diff
@@ -0,0 +1,160 @@
...
+				"reconcile_info":  "",
+				"reconcile_state": "",
+				"wait_to":         "reconcile",
+			},
+		}
+
+		require.Exactlyf(t, expected, respRows, "Expected to see correct changes")
+	})
+
+	logger.Section("stop serving CRD on v1beta1 and remove cr from manifest", func() {
```
praveenrewar: Should we first deploy the crd with the changes and then do another deployment with the CR being removed?

pkg/kapp/app/recorded_app.go
```diff
@@ -77,6 +77,51 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+
+	return meta.UsedGKs, nil
+}
+
+func (a *RecordedApp) UpdateUsedGKs(gvks []schema.GroupKind, merge bool) ([]schema.GroupKind, error) {
+	gksByGK := map[schema.GroupKind]struct{}{}
+	var uniqGKs []schema.GroupKind
+
+	if merge {
+		usedGKs, err := a.UsedGKs()
```
praveenrewar: Ideally, we should not use one public method inside another public method.

pkg/kapp/app/interfaces.go
100mik: `app.UsedGKs()` returns values from the cached app meta which is not updated. And we need to account for new KGs added in the manifest. Will explore if we can update the cache here as well. Not sure about it though.

This is so because so far:
1. All updates to the app meta were meant to be cached for the next deploy
2. The values used for the current deploy was the cached meta fetched from the config map.

100mik:
I was expecting the reviews to catch a lot of missed renames. Handling the mixed renames caught by y'all now 🚀 
Also removing the now unused functions. 
(Responding to this thread instead of replying to individual comments)

pkg/kapp/resources/resource.go
100mik: Yep, definitely better. Adopted this.

pkg/kapp/app/recorded_app.go
100mik: As of now we are using it basically as an internal function so I do not see any harm in making this internal.
The alternative would be to get the meta and then get the used GKs from that. I was trying to avoid duplication.

I am considering two possibilities now:
1. Update cache and use UsedGK to fetch GKs in `deploy.go`. We could have `usedGK()` and `UsedGK()` in this case (Rohit also noted this in his comments)
2. Make `UsedGKs` non exported and continue returning GKs with `UpdateUsedGKs()`

test/e2e/crd_versions_with_gk_scoping_test.go
100mik: We could definitely do that. I went with this because ordering should kick in and have the same affect.
You think doing it that way will make the case more legible?

pkg/kapp/app/recorded_app.go
praveenrewar: I think the first approach makes more sense.

test/e2e/crd_versions_with_gk_scoping_test.go
praveenrewar: I think this test would work for GVK scoping as well. But if we delete the cr in the next step, then the results for GK scoping and GVK scoping will be different.

100mik: In case of GVK scoping.
There won't be a delete as we won't list v1 resources.
Could you elaborate ? Not sure if I am missing something here

praveenrewar: > we won't be listing v1 resources

We won't have to. kapp would list the v1beta1 and show the summary based on that, only once the crd is updated, the v1beta1 version would cease to exists.

pkg/kapp/app/interfaces.go
100mik: Updating cache, and using UsedGVKs now 🚀

test/e2e/crd_versions_with_gk_scoping_test.go
100mik: Oh yes, I see it now, makes complete sense. Will make the change

pkg/kapp/app/recorded_app.go
100mik: Incorporated the first approach 🚀

pkg/kapp/resources/resource_types.go
100mik: That was my original intention, but `ResourceType` does not seem to expose that function.
Would be prefer constructing a `GroupKind` struct here and passing it over passing a `ResourceType` struct?
(I could implement the function but chose to not do that as the other functions surfacing group, kind, etc are due to the structs embedded in `ResourceType`)

test/e2e/crd_versions_with_gk_scoping_test.go
100mik: Made tests stricter 🚀

pkg/kapp/resources/resource_types.go
praveenrewar: > GroupKind struct here and passing it over passing a ResourceType struct?

We can do that. I don't have any strong reason apart from the fact that the other information is unnecessary/noisy in `Matches()`, so I will it up to you.

rohitagg2020:
LGTM

pkg/kapp/resources/resource_types.go
100mik: That is reasonable. I however did that in favour of keeping the function `MatchingAnyGK` concise.
I will leave this thread unresolved in case anyone wants to add any thoughts 🤔

praveenrewar (approved):
LGTM

sethiyash (approved):
Looks Good.

pkg/kapp/app/recorded_app.go
```diff
@@ -77,6 +77,53 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return meta.UsedGKs, nil
+}
+
+func (a *RecordedApp) UsedGKs() ([]schema.GroupKind, error) { return a.usedGKs() }
+
+func (a *RecordedApp) UpdateUsedGKs(gks []schema.GroupKind, merge bool) error {
```
cppforlife: can this method be simplified to just do updating instead of merging, can merging be done outside?

pkg/kapp/cmd/app/deploy.go
```diff
@@ -133,8 +134,18 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	err = app.UpdateUsedGKs(NewUsedGKsScope(newResources).GKs(), true)
+	if err != nil {
+		return err
+	}
```
cppforlife: why do we need to save new resource GKs so early? (we dont even know if user wants to do a deploy at this point, or may be is using --diff-run)

pkg/kapp/cmd/app/deploy.go
```diff
@@ -178,6 +189,12 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	// Cache possibly untracked GKs in existing resources (handles older apps)
+	err = app.UpdateUsedGKs(NewUsedGKsScope(append(newResources, existingResources...)).GKs(), false)
```
cppforlife: shouldnt we be eliminating old GKs in touch.Do below?

pkg/kapp/resources/identified_resources_list.go
```diff
@@ -36,6 +37,10 @@ func (r IdentifiedResources) List(labelSelector labels.Selector, resRefs []Resou
 		schema.GroupVersionResource{Version: "v1", Resource: "componentstatuses"},
 	})
 
+	if opts.GKsScope != nil {
```
cppforlife: it's a little weird to check slice against nil... is nil and empty here different meaning (seems to be)?

should we be using `resRefs []ResourceRef` somehow here instead of additional opts?

pkg/kapp/cmd/app/deploy.go
cppforlife: in my high level understanding, we should be doing something like:

- run diff with newRes+existingRes GKs
- save newRes+existingRes GKs
- run apply changes
- save newRes+existingRes GKs
  - interesting... when do GKs get removed for delete changes? (next deploy i guess?)

100mik: It makes complete sense to have the UsedGKs from the cluster merged with the new GKs from the YAML and then to update the app with latest cache after user confirms.

Yep GKs for delete changes are removed in the next deploy.

pkg/kapp/app/recorded_app.go
100mik: We can have something like `newAndUsedGKs()` in the deploy once we start updating them after the user confirms

pkg/kapp/resources/identified_resources_list.go
100mik: I do believe I have used `nil` and empty slices inter changeably. We could definitely switch it up to use empty slices everywhere instead,
Maybe we could check on length instead.

pkg/kapp/cmd/app/deploy.go
100mik: We could I believe update here to eliminate old GKs, if we are sure at this point that all resources that were meant to be deleted have been deleted (which is the case to the best of my knowledge)

pkg/kapp/resources/identified_resources_list.go
100mik: The ResourceRef struct is heavily based on GVR matching where as our focus is on GKs. Since we cannot obtain GVRs from manifests reliably, it does not let us scope aggressively.

100mik: Will verify that we are comparing slices appropriately everywhere

100mik: Went through the places where we were comparing this slice and switched to comparing lengths instead 🚀

pkg/kapp/cmd/app/resource_types_flags.go
```diff
@@ -21,6 +23,9 @@ func (s *ResourceTypesFlags) Set(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&s.ScopeToFallbackAllowedNamespaces, "dangerous-scope-to-fallback-allowed-namespaces",
 		false, "Scope resource searching to fallback allowed namespaces")
+
+	cmd.Flags().BoolVar(&s.DisableGKScoping, "dangerous-disable-gk-scoping",
+		false, "Disable scoping of resource searching to used GroupKinds")
```
100mik, line 28: Not the happiest with the flag name, open to suggestions!

pkg/kapp/app/recorded_app.go
```diff
@@ -80,6 +80,33 @@ func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
...
+	}
+
+	return meta.UsedGKs, nil
+}
+
+func (a *RecordedApp) UsedGKs() ([]schema.GroupKind, error) { return a.usedGKs() }
+
+func (a *RecordedApp) UpdateUsedGKs(gks []schema.GroupKind) error {
+	gksByGK := map[schema.GroupKind]struct{}{}
+	var uniqGKs []schema.GroupKind
```
cppforlife, line 97: lets make sure to sort them before saving

pkg/kapp/cmd/app/deploy.go
```diff
@@ -215,35 +234,77 @@ func (o *DeployOptions) Run() error {
...
+	var uniqGKs []schema.GroupKind
+
+	usedGKs, err := app.UsedGKs()
+	if err != nil {
+		return nil, err
+	}
+
+	// Handle existing apps without cached GKs
+	// These apps can cache and scope to GKs in subsequent deploys
+	lastChange, err := app.LastChange()
```
cppforlife: why do we need to check on last change?

pkg/kapp/cmd/app/deploy.go
```diff
@@ -203,6 +215,13 @@ func (o *DeployOptions) Run() error {
 		if err != nil {
 			return err
 		}
+
+		// Prunes unused GKs
+		err = app.UpdateUsedGKs(NewUsedGKsScope(newResources).GKs())
```
cppforlife, line 220: can UpdateUsedGKs and UpdateUsedGVs be combined so that we make less api calls

pkg/kapp/cmd/app/deploy.go
100mik: If LastChange is not nil, and there are no UsedGVs stored, it means that the app was created by an older version of kapp. (LastChange not being nil indicates that it has been deployed once).
Is there a better way of determining whether an app has been deployed by an older version of kapp?

pkg/kapp/app/recorded_app.go
100mik: We definitely could.
I am guessing the priority would be sort by Group, then Version and then Kind.
Is there a specific way in which that will help? (Just trying to understand why how it will help us)
Do we not do this explicitly for GroupVersions because it is done at some other point?

pkg/kapp/cmd/app/deploy.go
100mik: Yep I took note of this too, but I was hoping to pursue that as a separate PR.
(Right after this one)

pkg/kapp/app/recorded_app.go
cppforlife: any criteria is fine, as long as it is deterministic.

pkg/kapp/cmd/app/deploy.go
cppforlife: ok

pkg/kapp/app/recorded_app.go
100mik: Oh so our end result is to be deterministic.
Will make the change 🚀

pkg/kapp/cmd/app/deploy.go
cppforlife: LastChange does not really indicate that app was deployed by older version of kapp. newer version of kapp also have LastChange (so do super old ones, right?).

i think what you probably want is to make usedGKs a pointer (*[]stuff) so that you can see a distinction between field being present or not.

100mik: If we wanna use something like `usedGKs == nil` to differentiate between  a new deploy by new kapp and an old deploy done using old kapp, we will have to set `UsedGKs` to an empty list [here](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L86-L98) along with `labelKey` and `labelValue`.

I am making this change for now, we can revert it if we prefer the `lastChange` approach.

PR #409, opened 2022-01-13
Filtering and sorting on app change list 

### LINKED ISSUE #339: Improve how the user can sort and filter app-changes while listing them
Today we do not provide any flags with the `app-change list` command which allows the user to filter or list changes for a particular time frame easily.

Allowing the following flags will allow the user to find specific app-changes more effectively.

- `--before=<timestamp>`, `--after=<timestamp>`
    - Can be used to specify a time interval and list app-changes in that interval
- `--sort`
    - allowed values : `newest-first`,`oldest-first`
    - default value : `newest-first`
    - Let's the user decide how they want to list app-changes
- Would it make sense to have a `--duration` flag as well?
    - Example : `--duration 2h` gets changes for the past 2 hours 

---
Vote on this request

sethiyash:
- Added `--before`  `--after` flag to filter app-changes list by passing time stamp.
  - supported formats for time stamp `2006-01-02` and `2006-01-02T15:04:05Z07:00`
- ~~Change the sorting of app-change list by using `--sort` flag allowed values : `newest-first` (default), `oldest-first`~~

Example - `kapp app-change list -a <app-name> --before 2022-01-10T12:15:25Z`


pkg/kapp/cmd/appchange/list.go
```diff
@@ -4,23 +4,25 @@
...
-	"strings"
-
 	"github.com/cppforlife/go-cli-ui/ui"
 	uitable "github.com/cppforlife/go-cli-ui/ui/table"
 	ctlapp "github.com/k14s/kapp/pkg/kapp/app"
 	cmdapp "github.com/k14s/kapp/pkg/kapp/cmd/app"
 	cmdcore "github.com/k14s/kapp/pkg/kapp/cmd/core"
 	"github.com/k14s/kapp/pkg/kapp/logger"
 	"github.com/spf13/cobra"
+	"strings"
```
rohitagg2020: nit: Add a newline

pkg/kapp/cmd/appchange/time_flags.go
```diff
@@ -0,0 +1,24 @@
...
+	Before     string
+	After      string
+	BeforeTime time.Time
+	AfterTime  time.Time
+
+	Duration string
+}
+
+func (t *TimeFlags) Set(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&t.Before, "before", "", "", "List app-changes happened before given time stamp (format: before=<timestamp>)")
```
rohitagg2020: Can we also the timestamp format which is expected?

pkg/kapp/cmd/appchange/list.go
sethiyash: Done

pkg/kapp/cmd/appchange/time_flags.go
sethiyash: Done

pkg/kapp/app/apps.go
```diff
@@ -8,11 +8,12 @@ import (
...
 	"strings"
 
-	"github.com/k14s/kapp/pkg/kapp/logger"
-	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
+
+	"github.com/k14s/kapp/pkg/kapp/logger"
+	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
```
praveenrewar: Did auto save or build script mess up or was this intended?

pkg/kapp/app/apps.go
```diff
@@ -8,11 +8,12 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/k14s/kapp/pkg/kapp/logger"
-	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
+
```
cppforlife: nit: no newline

pkg/kapp/cmd/appchange/list.go
```diff
@@ -5,22 +5,26 @@ package appchange
 
 import (
 	"strings"
+	"time"
 
 	"github.com/cppforlife/go-cli-ui/ui"
 	uitable "github.com/cppforlife/go-cli-ui/ui/table"
+	"github.com/spf13/cobra"
+
```
praveenrewar: Same as above (Did auto save or build script mess up or was this intended?)

pkg/kapp/cmd/appchange/sort_flag.go
```diff
@@ -0,0 +1,21 @@
...
+package appchange
+
+import "github.com/spf13/cobra"
+
+type SortFlag struct {
+	Sort string
+}
+
+func (s *SortFlag) Set(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&s.Sort, "sort", "s", "newest-first", "Sort app-changes list (allowed values : newest-first, oldest-first)")
```
praveenrewar: What do you think about using a boolean flag called `newest-first` with default set to true?
Also, since the sorting is happening based on just the time field, do we want to keep this with `time_flags` itself?

pkg/kapp/cmd/appchange/list.go
```diff
@@ -75,14 +97,31 @@ func (t AppChangesTable) Print(ui ui.UI) {
...
+			{Column: 1, Asc: true},
+			{Column: 0, Asc: true},
+		}
 	}
 
+	isFilterApplied := !t.TimeFlags.BeforeTime.IsZero() || !t.TimeFlags.AfterTime.IsZero()
+
 	for _, change := range t.Changes {
+
+		if isFilterApplied {
```
praveenrewar: The behaviour would be same even if we remove this `if isFilterApplied` condition, right? Or am I just overthinking 👀

pkg/kapp/cmd/appchange/time_flags.go
```diff
@@ -0,0 +1,24 @@
...
+	After      string
+	BeforeTime time.Time
+	AfterTime  time.Time
+
+	Duration string
+}
+
+func (t *TimeFlags) Set(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&t.Before, "before", "", "", "List app-changes happened before given time stamp (format: before=2022-01-10T12:15:25Z)")
+	cmd.Flags().StringVarP(&t.After, "after", "", "", "List app-changes happened after given time stamp (format: after=2022-01-10T12:15:25Z)")
```
praveenrewar: I don't recall the exact use case here but the timestamp format makes me wonder if this will be "usable" 🤔

test/e2e/app_change_test.go
```diff
@@ -0,0 +1,86 @@
...
+kind: Service
+metadata:
+  name: redis-primary
+spec:
+  ports:
+  - port: 6380
+    targetPort: 6380
+  selector:
+    app: redis
+    tier: frontend
```
praveenrewar: I am not a big fan of having a lot of formatting to reduce some lines of code, but in this case I think we can make do with just one yaml and use `fmt.Sprintf()` to update the `tier` value, as it takes a lot of concentration to find one small difference in 2 almost same yamls.

pkg/kapp/app/apps.go
sethiyash: This is was the recommendation by Goland to sort the imports and it groups the one from same source. I can undo this if we don't want to follow this.

pkg/kapp/cmd/appchange/list.go
sethiyash: I believe this is was due to my Goland setting I have just reverted it.

pkg/kapp/cmd/appchange/time_flags.go
sethiyash: This is what I asked also in carvel-ind once. but as we are showing timestamps in `kapp app-change list` user can pick the timestamps from there and apply the filter I believe.

pkg/kapp/cmd/appchange/sort_flag.go
sethiyash: Yeah can certainly we can do that but just `--newest-first` by name is not indicating that we gonna sort by newest first. We can discuss this further and get thoughts from others to. 

I think if we decide to go just with the boolean flag then it make sense to add it in `time_flags`

pkg/kapp/app/apps.go
sethiyash: I believe this is was due to my Goland setting I have just reverted it.

pkg/kapp/cmd/appchange/list.go
sethiyash: Yeah removing this will also do the job. Looks like I was overthinking.

test/e2e/app_change_test.go
sethiyash: Done.

pkg/kapp/cmd/appchange/list.go
100mik: I believe we group by inbuilt packages and then other named/unnamed packages, Looks like there are 3 groups being created here. I guess we would not wanna do this change without reason.

pkg/kapp/cmd/appchange/sort_flag.go
100mik: I have been thinking how we can avoid a string flag, so it makes sense to do that. I wonder if we could name it better tho (`--newest-first false`  does not _necessarily_ indicate oldest first). I think newest-first is alright but worth some thought.

pkg/kapp/cmd/appchange/sort_flag.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+func (s *SortFlag) IsSortByNewestFirst() bool {
```
100mik: If we do go ahead with the string flag. We should definitely validate that the values are one of the allowed values, and error out otherwise while prompting the user with the acceptable values

pkg/kapp/cmd/appchange/time_flags.go
100mik: My thought behind going with that is that users will be copying values after listing them. I think it will be cool if we can support multiple valid formats tho. I do not think this is supported by the `time` package out of the box. 
@sethiyash today it errors out if not adhering exactly to the specific time format?

100mik: Also, would an option to filter on _just_ date be an useful feature?
(I guess the usability of that depends on how frequently the app is deployed.)

pkg/kapp/cmd/appchange/list.go
```diff
@@ -5,22 +5,26 @@ package appchange
...
 
 type ListOptions struct {
 	ui          ui.UI
 	depsFactory cmdcore.DepsFactory
 	logger      logger.Logger
 
-	AppFlags cmdapp.Flags
+	AppFlags  cmdapp.Flags
+	TimeFlags TimeFlags
+	SortFlag  SortFlag
```
100mik: Maybe group SortFlag into TimeFlags?

pkg/kapp/cmd/appchange/list.go
```diff
@@ -49,14 +55,30 @@ func (o *ListOptions) Run() error {
...
+		}
+	}
+
+	if o.TimeFlags.After != "" {
+		o.TimeFlags.AfterTime, err = time.Parse(time.RFC3339, o.TimeFlags.After)
+		if err != nil {
+			return err
+		}
+	}
+
```
100mik, line 75: I think we should verify that the After time is less than the before time if both are non-nil values?
I think today we will simply list no values if it is not as none of the records satisfy those conditions.
Maybe we should error out as a user might be doing this erroneously?

pkg/kapp/cmd/appchange/sort_flag.go
100mik: +1 on merging sort flag into time flags

pkg/kapp/cmd/appchange/list.go
100mik: +1 looks like a double check
Not sure if we should keep `isFilterApplied` or makes the condition below simpler. 🤔

100mik (commented):
Put some thoughts down!
(also couple of missing checks I think should be there)

pkg/kapp/cmd/appchange/time_flags.go
sethiyash: Yeah, we can support just date also. Let's get thoughts from others also on this.

pkg/kapp/cmd/appchange/list.go
sethiyash: Looks like you are willing to say we should through an error when before time is less than after time.

Here after time is less than before time but this can be a valid usecase
`kapp appchange list -a <app-name> --before 2022-01-10T02:00:25Z --after 2022-01-09T11:15:25Z`
(here user wants app change list from 09 Jan 11:15 to 10 Jan 2:00)

While when before time is less than after time then it doesn't make sense. something like - 
`kapp appchange list -a <app-name> --before 2022-01-10T02:00:25Z --after 2022-01-10T11:15:25Z`

is my understanding correct?

100mik: Yes that's correct. Switched them up by mistake.

sethiyash: Done

pkg/kapp/util/time.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+func ParseTime(input string, formats []string) (time.Time, error) {
+	for _, format := range formats {
+		t, err := time.Parse(format, input)
+		if err == nil {
```
100mik: Are we sure errors are only thrown when there is an improper format?
(I think it is, but might be worth verifying)

test/e2e/app_change_test.go
```diff
@@ -0,0 +1,81 @@
...
+		resp := uitest.JSONUIFromBytes(t, []byte(out))
+
+		require.Equal(t, 2, len(resp.Tables[0].Rows), "Expected to have 2 app-changes")
+
+		require.Equal(t, "update: Op: 0 create, 0 delete, 1 update, 0 noop, 0 exists / Wait to: 1 reconcile, 0 delete, 0 noop", resp.Tables[0].Rows[0]["description"], "Expected description to match")
+		require.Equal(t, "update: Op: 1 create, 0 delete, 0 update, 0 noop, 0 exists / Wait to: 1 reconcile, 0 delete, 0 noop", resp.Tables[0].Rows[1]["description"], "Expected description to match")
+	})
+
+	logger.Section("app change list filter with after flag", func() {
+		out, _ := kapp.RunWithOpts([]string{"app-change", "ls", "-a", name, "--after", time.Now().Add(24 * time.Hour).Format("2006-01-02"), "--json"}, RunOpts{})
```
100mik, line 79: Is the time added here 24 hours because the intention was to check both formats?

100mik (commented):
LGTM, just some clarifications!
Just left a few comments and thoughts that struck me while going through it.
I don't remember if we made a call about whether we are gonna handle time zones specially or not. If we are sure that we are not gonna go for that, I am gonna go ahead and approve this. 🚀

test/e2e/app_change_test.go
sethiyash: Yes, that's correct. Want to check just the date also.

pkg/kapp/cmd/appchange/time_flags.go
rohitagg2020: As we are supporting 2 formats, can we mention both of them ?

test/e2e/app_change_test.go
```diff
@@ -0,0 +1,81 @@
...
+		resp := uitest.JSONUIFromBytes(t, []byte(out))
+
+		require.Equal(t, 2, len(resp.Tables[0].Rows), "Expected to have 2 app-changes")
+
+		require.Equal(t, "update: Op: 0 create, 0 delete, 1 update, 0 noop, 0 exists / Wait to: 1 reconcile, 0 delete, 0 noop", resp.Tables[0].Rows[0]["description"], "Expected description to match")
+		require.Equal(t, "update: Op: 1 create, 0 delete, 0 update, 0 noop, 0 exists / Wait to: 1 reconcile, 0 delete, 0 noop", resp.Tables[0].Rows[1]["description"], "Expected description to match")
+	})
+
+	logger.Section("app change list filter with before flag", func() {
+		out, _ := kapp.RunWithOpts([]string{"app-change", "ls", "-a", name, "--before", time.Now().Add(2 * time.Second).Format(time.RFC3339), "--json"}, RunOpts{})
```
rohitagg2020: nit: Add a comment which mentions which format it will test.

pkg/kapp/util/time.go
sethiyash: Yeah that I have verified it will only return an error when none of the formats matches.

test/e2e/app_change_test.go
sethiyash: Done

pkg/kapp/cmd/appchange/time_flags.go
sethiyash: Yeah, thanks for pointing it out. Done.

sethiyash:
@100mik as discussed earlier we will not be handling time zones here.

pkg/kapp/cmd/appchange/list.go
```diff
@@ -20,9 +22,12 @@ type ListOptions struct {
 	depsFactory cmdcore.DepsFactory
 	logger      logger.Logger
 
-	AppFlags cmdapp.Flags
+	AppFlags  cmdapp.Flags
+	TimeFlags TimeFlags
 }
 
+const dateFormat = "2006-01-02"
```
cppforlife: nit: since its only being used in one place, lets inline

pkg/kapp/cmd/appchange/list.go
```diff
@@ -98,3 +129,13 @@ func (t AppChangesTable) Print(ui ui.UI) {
 
 	ui.PrintTable(table)
 }
+
+func parseTime(input string, formats []string) (time.Time, error) {
```
cppforlife: lets put this on `func (o *ListOptions)` to reduce scope where this function is available.

pkg/kapp/cmd/appchange/time_flags.go
```diff
@@ -0,0 +1,24 @@
...
+	Before     string
+	After      string
+	BeforeTime time.Time
+	AfterTime  time.Time
+
+	Duration string
+}
+
+func (t *TimeFlags) Set(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&t.Before, "before", "", "List app-changes happened before given time stamp (formats: before=2022-01-10T12:15:25Z, before=2022-01-10)")
```
cppforlife: nit: `List app-changes happened before given time stamp` => `List app changes before given time`

cppforlife: no need to specify `before=` in the examples

pkg/kapp/cmd/appchange/list.go
sethiyash: Adopted.

pkg/kapp/cmd/appchange/list.go
sethiyash: Adopted.

pkg/kapp/cmd/appchange/time_flags.go
sethiyash: Adopted.

rohitagg2020:
LGTM

pkg/kapp/cmd/appchange/list.go
```diff
@@ -75,14 +110,18 @@ func (t AppChangesTable) Print(ui ui.UI) {
 			uitable.NewHeader("Description"),
 			nsHeader,
 		},
-
```
praveenrewar: nit: I think we can continue having this new line.

pkg/kapp/cmd/appchange/list.go
```diff
@@ -75,14 +110,18 @@ func (t AppChangesTable) Print(ui ui.UI) {
...
 		SortBy: []uitable.ColumnSort{
 			{Column: 1, Asc: false},
 			{Column: 0, Asc: true}, // in case start time are same
 		},
 	}
 
 	for _, change := range t.Changes {
+		if (!t.TimeFlags.BeforeTime.IsZero() && !change.Meta().StartedAt.Before(t.TimeFlags.BeforeTime)) ||
+			(!t.TimeFlags.AfterTime.IsZero() && !change.Meta().StartedAt.After(t.TimeFlags.AfterTime)) {
```
praveenrewar: *question*: Do we need this `t.TimeFlags.AfterTime.IsZero()`?
I am thinking that if `AfterTime` is `Zero`, then `change.Meta().StartedAt.After(t.TimeFlags.AfterTime)` would always be false. Or maybe I am missing some edge case!

praveenrewar: Also, would it make sense to have the `!` outside, in my mind that would be more readable.
```suggestion
			!(t.TimeFlags.AfterTime.IsZero() && change.Meta().StartedAt.After(t.TimeFlags.AfterTime)) {
```

test/e2e/app_change_test.go
praveenrewar: Did we add this part `.Format(time.RFC3339)` after adding the comment as the format that will be tested against is clearly visible right now.
(No strong opinions on it though)

praveenrewar: `time.Now().Add(2 * time.Second)`
I am not sure why we are adding those 2 seconds, even `time.Now()` should be good.

praveenrewar: *In addition to this*, we could `Subtract` let's say a Minute from `time.Now()` and show that none of the changes get's listed because they did not occur before that time?

test/e2e/app_change_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+		resp := uitest.JSONUIFromBytes(t, []byte(out))
+
+		require.Equal(t, 2, len(resp.Tables[0].Rows), "Expected to have 2 app-changes")
+		require.Equal(t, "update: Op: 0 create, 0 delete, 1 update, 0 noop, 0 exists / Wait to: 1 reconcile, 0 delete, 0 noop", resp.Tables[0].Rows[0]["description"], "Expected description to match")
+		require.Equal(t, "update: Op: 1 create, 0 delete, 0 update, 0 noop, 0 exists / Wait to: 1 reconcile, 0 delete, 0 noop", resp.Tables[0].Rows[1]["description"], "Expected description to match")
+	})
+
+	logger.Section("app change list filter with after flag", func() {
+		// testing with just date based format
```
praveenrewar: nit: Same as above (might not be required)

test/e2e/app_change_test.go
```diff
@@ -0,0 +1,81 @@
...
+		out, _ := kapp.RunWithOpts([]string{"app-change", "ls", "-a", name, "--after", time.Now().Add(24 * time.Hour).Format("2006-01-02"), "--json"}, RunOpts{})
```
praveenrewar, line 79: We could have another test by subtracting 1 Minute from `time.Now()` and use it with `after` to show both the changes being present.

pkg/kapp/cmd/appchange/list.go
renuy: @praveenrewar this would change the condition....Please check. However, if the condition can be made simpler, I am all for it

renuy: !(t.TimeFlags.AfterTime.IsZero() || change.Meta().StartedAt.After(t.TimeFlags.AfterTime))

praveenrewar: My bad. @renuy Thank for pointing it out.

praveenrewar, line 121:
```suggestion
		if !(!t.TimeFlags.BeforeTime.IsZero() && change.Meta().StartedAt.Before(t.TimeFlags.BeforeTime) && change.Meta().StartedAt.After(t.TimeFlags.AfterTime))  {
```
How about this? 
Basically, here the internal conditional evaluates to true whenever the current change is in between before time and after time and then the `!` outside will help us continue for all the other changes.

praveenrewar: I think we can move the append statement inside a simplified if condition instead of using continue.

sethiyash: > I think we can move the append statement inside a simplified if condition instead of using continue.

I have used `continue` here as we just have to skip which are not satisfying the filter criteria and append everything else. Also looks like if we start appending inside some `if condition` then we also have to handle a scenario when no filter is applied append all of them.

sethiyash: Also up for discussing it on call

test/e2e/app_change_test.go
sethiyash: > time.Now().Add(2 * time.Second)
> I am not sure why we are adding those 2 seconds, even time.Now() should be good.

With `time.Now()` testcase fails as it is deployed in the same second and when we do filter by `--before ` the given time is not inclusive.

pkg/kapp/cmd/appchange/list.go
sethiyash: added back.

test/e2e/app_change_test.go
sethiyash: removed.

test/e2e/app_change_test.go
sethiyash: > In addition to this, we could Subtract let's say a Minute from time.Now() and show that none of the changes get's listed because they did not occur before that time?

Added.

test/e2e/app_change_test.go
sethiyash: I was thinking one-one testcase is sufficient for each `before` and `after` but no harm in having as many as possible. Added it.

100mik (approved):
LGTM! Praveen seems to have covered most concerns ^

praveenrewar (approved):
LGTM

PR #411, opened 2022-01-14
Preserve status for all resources on update

scothis:
Update the default config to ignore 'new' status, and instead preserve
the a resource's 'existing' status.

Resources that define a status will no longer try to update the server
to make that status empty. Status may not be defined when creating a
resource (this is enforced by k8s) and should not be defined on update,
as this state should be owned by the controller for that resource.

Many tools will include "empty" scaffolding for a resource when
unmarshaling a struct. These values will now be ignored for status. For
example, controllergen includes a complex empty object for the status
of CustomResourceDefinitions that kapp will now ignore.

Signed-off-by: Scott Andrews <andrewssc@vmware.com>

### DIFF
--- a/test/e2e/config_test.go
+++ b/test/e2e/config_test.go
@@ -413,3 +413,89 @@ func asYAML(t *testing.T, val interface{}) string {
 	require.NoError(t, err)
 	return string(bs)
 }
+
+func TestDefaultConfig_PreserveExistingStatus(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	kubectl := Kubectl{t, env.Namespace, logger}
+
+	yaml := `
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
+  name: tests.kapp.example
+spec:
+  group: kapp.example
+  names:
+    kind: Test
+    listKind: TestList
+    plural: tests
+    singular: test
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        properties:
+          apiVersion:
+            type: string
+          kind:
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties: {}
+            type: object
+          status:
+            properties: {}
+            type: object
+        required:
+        - spec
+        type: object
+    served: true
+    storage: true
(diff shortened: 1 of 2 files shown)


100mik (approved):
LGTM!

test/e2e/config_test.go
```diff
@@ -413,3 +413,89 @@ func asYAML(t *testing.T, val interface{}) string {
...
+	kubectl := Kubectl{t, env.Namespace, logger}
+
+	yaml := `
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
```
praveenrewar: Should we remove this and other such fields which are not relevant to the test (Just to make the tests smaller)?

test/e2e/config_test.go
```diff
@@ -413,3 +413,89 @@ func asYAML(t *testing.T, val interface{}) string {
...
+	defer cleanUp()
+
+	var crd ClusterResource
+
+	logger.Section("initial deploy", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml)})
+
+		crd = NewPresentClusterResource("customresourcedefinition", "tests.kapp.example", env.Namespace, kubectl)
+		kind := crd.RawPath(ctlres.NewPathFromStrings([]string{"status", "acceptedNames", "kind"})).(string)
```
praveenrewar, line 465: I think we can get the kind inline like [this](https://github.com/vmware-tanzu/carvel-kapp/blob/1d41764d78ac38baf4432c8c042076191aa09999/test/e2e/config_test.go#L238) as we are not using crd anywhere else.

test/e2e/config_test.go
scothis: I can prune it back to the core essence. I included a full resource as generated by controller-gen because that's where I encountered the issue

test/e2e/config_test.go
scothis: that's going to be a very long line

test/e2e/config_test.go
praveenrewar: I think that having just the core essence should be good enough as the functionality effects only a specific section of the resource.

test/e2e/config_test.go
praveenrewar: Maybe split it across 2 lines? I don't have a strong opinion so I will leave it upto you :)

praveenrewar (approved):
LGTM

rohitagg2020 (approved):

PR #413, opened 2022-01-19
Pinning trivy cloning to branch v0.22.0

rohitagg2020:
We can remove this and the git clone code once trivy release v0.23.0

- there was a breaking change introduced on the trivy develop branch that moved the sarif template into go code and introduced the --format sarif flag. Once they release that we can move to using it


sethiyash (approved):
Looks good.

praveenrewar:
> We can remove this and the git clone code once trivy release v0.23.0

I would actually prefer to continue using a specific version instead of using the latest changes.

praveenrewar (approved):

PR #416, opened 2022-01-24
drop require block to allow - go install to work

dprotaso:
Ran the `./hack/test-all.sh` locally and it passed

Right now `go install` fails with the following error

```
$ go install  github.com/k14s/kapp/cmd/kapp@latest
go install: github.com/k14s/kapp/cmd/kapp@latest (in github.com/k14s/kapp@v0.44.0):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```


dprotaso:
@cppforlife note you'll want to cut a release for `go install to work with @latest` version

dprotaso:
`@develop` works in the meantime

PR #420, opened 2022-01-25
Add newer versions/releases of tools in examples

praveenrewar:
The `apiextensions.k8s.io/v1beta1` API version of CustomResourceDefinition is [no longer served as of v1.22.](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122)

These tools require updating to a newer version to support working on Kubernetes v1.22+
- [x] cert-manager
- [x] istio
- [x] knative

We will also add these tools to our external tests.
- [x] gatekeeper
- [x] pinniped


hack/test-external.sh
```diff
@@ -2,19 +2,28 @@
...
-time kapp delete -y -a cf
+# TODO Add cf-for-k8s-v5.4.3
+
+time kapp deploy -y -a gk -f examples/gatekeeper-v3.7.0/config.yml
+time kapp delete -y -a gk
+
+time kapp deploy -y -a pinniped -f examples/pinniped-v0.13.0/
+time kapp delete -y -a pinniped
+
+pkill -9 'minikube'
```
rohitagg2020, line 27: Can we add something specific to `tunnel`. I think this might stop `minikube` sometime.

praveenrewar: It should not, as minikube doesn't run as a process. However, `minikube tunnel` does require password to run, so I am not sure how it would behave in the GitHub Actions environment.

rohitagg2020: that's strange, minikube doesn't run as process. So, you are saying minikube doesn't run as process but minikube tunnel does ?

praveenrewar: `minikube` runs as a _service_ and `minikube tunnel `runs as a _process_. You can try using `pgrep` to confirm.

rohitagg2020:
LGTM

PR #430, opened 2022-02-03
kapp deploy without disable original annotation when resource size exceed

### LINKED ISSUE #410: disable-original by default if

When a resource is exceeds Kubernetes' max annotation length, you need to set `kapp.k14s.io/disable-original`.  

From https://carvel.dev/kapp/docs/latest/diff/#kappk14siodisable-original: 

> kapp.k14s.io/disable-original annotation controls whether to record provided resource copy (rarely wanted)
> 
> Possible values: "" (empty). In some cases it’s not possible or wanted to record applied resource copy into its annotation kapp.k14s.io/original. One such case might be when resource is extremely lengthy (e.g. long ConfigMap or CustomResourceDefinition) and will exceed annotation value max length of 262144 bytes.

kapp should automatically disable original when saving it would exceed Kubernetes' limit. 

---
Vote on this request

sethiyash:
- no need to set `kapp.k14s.io/disable-original` annotation when resource size exceeds the limit of Kubernetes max annotation length. 
  - will not be recording resource in `kapp.k14s.io/original` annotation when resource limit exceeds.

### DIFF
--- a/pkg/kapp/clusterapply/add_or_update_change.go
+++ b/pkg/kapp/clusterapply/add_or_update_change.go
@@ -5,7 +5,6 @@ package clusterapply
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
 	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
@@ -236,16 +235,13 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
 		}
 
 		// Record last applied change on the latest version of a resource
-		latestResWithHistoryUpdated, isAnnValueMaxLengthExceeded, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
+		latestResWithHistoryUpdated, madeAnyModifications, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
 		if err != nil {
-			if strings.Contains(err.Error(), "unable to record resource, limit exceed") {
-				return true, nil
-			}
 			return true, fmt.Errorf("Recording last applied resource: %s", err)
 		}
 
-		// exceed annotation value max length then don't record the resource
-		if isAnnValueMaxLengthExceeded {
+		// when annotation value max length exceed then don't record the resource hence not making any modification
+		if !madeAnyModifications {
 			return true, nil
 		}
 
--- a/pkg/kapp/clusterapply/add_or_update_change.go
+++ b/pkg/kapp/clusterapply/add_or_update_change.go
@@ -5,7 +5,7 @@ package clusterapply
 
 import (
 	"fmt"
-	"os"
+	"strings"
 	"time"
 
 	ctldiff "github.com/k14s/kapp/pkg/kapp/diff"
@@ -24,10 +24,6 @@ const (
 	updateStrategyFallbackOnReplaceAnnValue ClusterChangeApplyStrategyOp = "fallback-on-replace"
 	updateStrategyAlwaysReplaceAnnValue     ClusterChangeApplyStrategyOp = "always-replace"
 	updateStrategySkipAnnValue              ClusterChangeApplyStrategyOp = "skip"
-)
-
-var (
-	resourceWithHistoryDebug = os.Getenv("KAPP_DEBUG_RESOURCE_WITH_HISTORY") == "true"
 )
 
 type AddOrUpdateChangeOpts struct {
@@ -223,28 +219,6 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
 		return fmt.Errorf("Calculating change after the save: %s", err)
 	}
 
-	// Use compact representation to take as little space as possible
-	// because annotation value max length is 262144 characters
-	// (https://github.com/vmware-tanzu/carvel-kapp/issues/48).
-	appliedResBytes, err := applyChange.AppliedResource().AsCompactBytes()
-	if err != nil {
-		return fmt.Errorf("Compact representation of applied resource: %s", err)
-	}
-
-	diff := applyChange.OpsDiff()
-
-	const resourceSizeMaxLen = 262144
-
-	// kapp deploy should work without adding disable annotation when resource size exceed
-	// (https://github.com/vmware-tanzu/carvel-kapp/issues/410)
-	if len(appliedResBytes) > resourceSizeMaxLen || len(diff.MinimalMD5()) > resourceSizeMaxLen {
-		if resourceWithHistoryDebug {
-			fmt.Printf("Unable to record resource, exceed limit of %d bytes (this can result in dumb diffing) \n",
-				resourceSizeMaxLen)
-		}
-		return nil
-	}
-
 	// first time, try using memory copy
 	latestResWithHistory := &savedResWithHistory
 
@@ -262,8 +236,11 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
 		}
 
 		// Record last applied change on the latest version of a resource
-		latestResWithHistoryUpdated, err := latestResWithHistory.RecordLastAppliedResource(appliedResBytes, diff)
+		latestResWithHistoryUpdated, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
 		if err != nil {
+			if strings.Contains(err.Error(), "unable to record resource, limit exceed") {
+				return true, nil
+			}
 			return true, fmt.Errorf("Recording last applied resource: %s", err)
 		}
 
--- a/pkg/kapp/clusterapply/add_or_update_change.go
+++ b/pkg/kapp/clusterapply/add_or_update_change.go
@@ -236,12 +236,17 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
 		}
 
 		// Record last applied change on the latest version of a resource
-		latestResWithHistoryUpdated, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
+		latestResWithHistoryUpdated, isAnnValueMaxLengthExceeded, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
 		if err != nil {
 			if strings.Contains(err.Error(), "unable to record resource, limit exceed") {
 				return true, nil
 			}
 			return true, fmt.Errorf("Recording last applied resource: %s", err)
+		}
+
+		// exceed annotation value max length then don't record the resource
+		if isAnnValueMaxLengthExceeded {
+			return true, nil
 		}
 
 		_, err = c.identifiedResources.Update(latestResWithHistoryUpdated)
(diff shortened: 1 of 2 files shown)


pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -218,6 +223,28 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
...
+		return fmt.Errorf("Compact representation of applied resource: %s", err)
+	}
+
+	diff := applyChange.OpsDiff()
+
+	const resourceSizeMaxLen = 262144
+
+	// kapp deploy should work without adding disable annotation when resource size exceed
+	// (https://github.com/vmware-tanzu/carvel-kapp/issues/410)
+	if len(appliedResBytes) > resourceSizeMaxLen || len(diff.MinimalMD5()) > resourceSizeMaxLen {
```
sethiyash: Checking here for `len(diff.MinimalMD5())` as we were doing this earlier also.

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -218,6 +223,28 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
...
+		if resourceWithHistoryDebug {
+			fmt.Printf("Unable to record resource, exceed limit of %d bytes (this can result in dumb diffing) \n",
```
100mik: I do not think we wanna use the term "dumb" in our code
Maybe "might result in the diff showing deletion of fields applied by external agents"

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -25,6 +26,10 @@ const (
 	updateStrategySkipAnnValue              ClusterChangeApplyStrategyOp = "skip"
 )
 
+var (
+	resourceWithHistoryDebug = os.Getenv("KAPP_DEBUG_RESOURCE_WITH_HISTORY") == "true"
```
100mik: `resourceWithHistoryDebug` is used to debug the hashing and ops diffing we do while deciding what kind of diff we do.
Is there a particular reason why we are printing the warning only in the debug mode and not for our users?

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -218,6 +223,28 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
...
 
+	// Use compact representation to take as little space as possible
+	// because annotation value max length is 262144 characters
+	// (https://github.com/vmware-tanzu/carvel-kapp/issues/48).
+	appliedResBytes, err := applyChange.AppliedResource().AsCompactBytes()
+	if err != nil {
+		return fmt.Errorf("Compact representation of applied resource: %s", err)
+	}
+
+	diff := applyChange.OpsDiff()
```
100mik: I think this should stay in the diff package, with the rest of the diffing logic and related functions.
Is it not feasible to add this check in the diff package itself?
(Taking a look myself as well)

100mik (commented):
Some thoughts

cppforlife:
i see @100mik took a stab at reviewing. ill add to his comments...

is there a reason why we couldnt localize this logic within RecordLastAppliedResource. the point of having ResourceWithHistory is to manage necessary on the metadata on the resource. it seems like we are spilling it out into other parts of the code (AddOrUpdateChange).

pkg/kapp/clusterapply/add_or_update_change.go
sethiyash: Yeah, make sense. My initial thought to validate it here as we are checking for `disable-original` annotation and in case we just want to check for `len(appliedResBytes) > resourceSizeMaxLen` then we don't have to calculate diff here. 
But earlier also we are checking the length of `diff.MinimalMD5()` along with resource length than its better to keep it under `RecordLastAppliedResource`

pkg/kapp/clusterapply/add_or_update_change.go
sethiyash: Sounds Good. Done.

sethiyash:
My initial thought to validate it here as we are checking for disable-original annotation and in case we just want to check for `len(appliedResBytes) > resourceSizeMaxLen` then we don't have to calculate diff here.
But earlier also we are checking the length of `diff.MinimalMD5()` along with resource length than I believe its better to keep it under `RecordLastAppliedResource` only.

pkg/kapp/clusterapply/add_or_update_change.go
sethiyash: I think this is because the users didn't much care about the warnings shown. But this is debatable if others also think that it is good to show a warning by default then I can make that change. 
(Keeping this unresolved to get the inputs)

100mik: I definitely think that we should let the users know, as this behaviour actually results in them seeing diffs they otherwise would not have.

pkg/kapp/diff/resource_with_history.go
```diff
@@ -80,6 +80,19 @@ func (r ResourceWithHistory) RecordLastAppliedResource(appliedChange Change) (ct
 
 	diff := appliedChange.OpsDiff()
 
+	const resourceSizeMaxLen = 262144
+
+	// kapp deploy should work without adding disable annotation when resource size exceed
+	// (https://github.com/vmware-tanzu/carvel-kapp/issues/410)
+	if len(appliedResBytes) > resourceSizeMaxLen || len(diff.MinimalMD5()) > resourceSizeMaxLen {
+		if resourceWithHistoryDebug {
+			fmt.Printf("Unable to record resource, exceed limit of %d bytes "+
```
100mik: I would still insist that this information is made available to the users, in a way that gives them an idea of how this affects their experience

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -237,6 +238,9 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
 		// Record last applied change on the latest version of a resource
 		latestResWithHistoryUpdated, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
 		if err != nil {
+			if strings.Contains(err.Error(), "unable to record resource, limit exceed") {
```
rohitagg2020: Is there any error code passed specifically for limit exceed? If yes, we should use that. Error messages should be the last option as they are very fragile and can be changed over time.

pkg/kapp/diff/resource_with_history.go
```diff
@@ -80,6 +80,19 @@ func (r ResourceWithHistory) RecordLastAppliedResource(appliedChange Change) (ct
 
 	diff := appliedChange.OpsDiff()
 
+	const resourceSizeMaxLen = 262144
```
rohitagg2020: nit: Add a comment that it is 256KB

pkg/kapp/diff/resource_with_history.go
```diff
@@ -80,6 +80,19 @@ func (r ResourceWithHistory) RecordLastAppliedResource(appliedChange Change) (ct
...
+
+	// kapp deploy should work without adding disable annotation when resource size exceed
+	// (https://github.com/vmware-tanzu/carvel-kapp/issues/410)
+	if len(appliedResBytes) > resourceSizeMaxLen || len(diff.MinimalMD5()) > resourceSizeMaxLen {
+		if resourceWithHistoryDebug {
+			fmt.Printf("Unable to record resource, exceed limit of %d bytes "+
+				"(this might result in the diff showing deletion of fields applied by external agents) \n",
+				resourceSizeMaxLen)
+		}
+		return nil, fmt.Errorf("unable to record resource, limit exceed")
```
rohitagg2020: It would be better if we can mention resource name.

pkg/kapp/diff/resource_with_history.go
sethiyash: Yeah, we can take inputs from others also on this.

pkg/kapp/diff/resource_with_history.go
sethiyash: This is just for our internal purpose this is not something that will be shown to the user.

pkg/kapp/diff/resource_with_history.go
sethiyash: 262144 bytes get converted to 262.144KB. I'm not sure will this be of more help.

pkg/kapp/diff/resource_with_history.go
sethiyash: This is for our internal purpose this is not something shown to the user.

pkg/kapp/clusterapply/add_or_update_change.go
sethiyash: This is just for internal purposes, not something shown to the user. I believe `508` error code is used for `Resource Limit Reached`  but looks like we don't use the error code in our cli.

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -235,11 +236,19 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
 		}
 
 		// Record last applied change on the latest version of a resource
-		latestResWithHistoryUpdated, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
+		latestResWithHistoryUpdated, isAnnValueMaxLengthExceeded, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
 		if err != nil {
+			if strings.Contains(err.Error(), "unable to record resource, limit exceed") {
```
cppforlife: this should not be checked anymore if bool is in play.

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -235,11 +236,19 @@ func (c AddOrUpdateChange) recordAppliedResource(savedRes ctlres.Resource) error
 		}
 
 		// Record last applied change on the latest version of a resource
-		latestResWithHistoryUpdated, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
+		latestResWithHistoryUpdated, isAnnValueMaxLengthExceeded, err := latestResWithHistory.RecordLastAppliedResource(applyChange)
```
cppforlife: isAnnValueMaxLengthExceeded => madeAnyModifications

we should make this transparent to consumers why we have not make any modifications.

sethiyash: Sure. I have started checking the negation of `madeAnyModifications` as I think it makes more sense and reverted all the returning bool values.

pkg/kapp/clusterapply/add_or_update_change.go
sethiyash: Yeah got missed. Removed now.

PR #436, opened 2022-02-16
Merge cm updates

100mik:
Reduce the number of calls being made to update app config map

### DIFF
--- a/pkg/kapp/app/interfaces.go
+++ b/pkg/kapp/app/interfaces.go
@@ -19,9 +19,8 @@ type App interface {
 
 	LabelSelector() (labels.Selector, error)
 	UsedGVs() ([]schema.GroupVersion, error)
-	UpdateUsedGVs([]schema.GroupVersion) error
 	UsedGKs() (*[]schema.GroupKind, error)
-	UpdateUsedGKs([]schema.GroupKind) error
+	UpdateUsedGVsAndGKs([]schema.GroupVersion, []schema.GroupKind) error
 
 	CreateOrUpdate(map[string]string) error
 	Exists() (bool, string, error)
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -65,22 +65,6 @@ func (a *RecordedApp) UsedGVs() ([]schema.GroupVersion, error) {
 	return meta.UsedGVs, nil
 }
 
-func (a *RecordedApp) UpdateUsedGVs(gvs []schema.GroupVersion) error {
-	gvsByGV := map[schema.GroupVersion]struct{}{}
-	var uniqGVs []schema.GroupVersion
-
-	for _, gv := range gvs {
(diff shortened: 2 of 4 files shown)


100mik:
Started sorting GVs before storing them as well to make them deterministic
cc: @cppforlife

pkg/kapp/app/recorded_app.go
```diff
@@ -92,7 +76,17 @@ func (a *RecordedApp) usedGKs() (*[]schema.GroupKind, error) {
 
 func (a *RecordedApp) UsedGKs() (*[]schema.GroupKind, error) { return a.usedGKs() }
 
-func (a *RecordedApp) UpdateUsedGKs(gks []schema.GroupKind) error {
+func (a *RecordedApp) UpdateUsedGVsAndGKs(gvs []schema.GroupVersion, gks []schema.GroupKind) error {
```
rohitagg2020, line 79: nit: As this function has grown in size, we can call two private functions 1. `func (a *RecordedApp) updateUsedGVs(gvs []schema.GroupVersion) error` and other `func (a *RecordedApp) updateUsedGKs(gks []schema.GroupKind) error` and then call 
```
a.update(func(meta *Meta) {
		meta.UsedGVs = uniqGVs
		meta.UsedGKs = &uniqGKs
})```

instead of writing everything inside it.

File: pkg/kapp/app/interfaces.go [hunk 2]
```diff
@@ -19,9 +19,8 @@ type App interface {
 
 	LabelSelector() (labels.Selector, error)
 	UsedGVs() ([]schema.GroupVersion, error)
-	UpdateUsedGVs([]schema.GroupVersion) error
 	UsedGKs() (*[]schema.GroupKind, error)
```
rohitagg2020, line 22: nit: Do we need a reference here. I am asking because `UsedGVs()` doesn't return the reference.

100mik: Yep we use it to determine whether an existing app was created by the older or newer kapp.
 This pointer would be nil for older apps as we set it [here](https://github.com/vmware-tanzu/carvel-kapp/commit/98b377c63a797aa00c50867ea7c2129cabb0c709#diff-c4cadf071a66c0e934e58c35eb0a9116fee4c744abd14621dc1eeebe7f5e9fb8R133)

pkg/kapp/app/recorded_app.go
100mik: I would say it is pretty straight forward (with the exact same operations being performed on 2 lists of things)
I did consider it initially but chose not to do that as it would be something like `uniqGKs()` and `uniqGVs()` since the functions would just deduplicate the list and the actual update happens in the function itself.

I decided to stick with this because it is more logically straightforward. Leaving this open for discussion.

cppforlife: im ok with current way.

PR #443, opened 2022-02-23
Fixed Release Automation

rohitagg2020:
Fixed Release Automation
- Added checksum.
- Added post Release workflow as well.
- Moved back from `goreleaser` to `uploading the binaries by script` as we faced issue with goreleaser earlier. The checksum calculated and uploaded by goreleaser were different.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -117,3 +117,4 @@ jobs:
           ${{steps.get-checksums-from-draft-release.outputs.result}}
           EOF
           )
+          
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,16 +1,14 @@
-name: goreleaser
+name: kapp-release
 
 on:
   push:
     tags:
-      - 'v*'
+      - "v*"
 
 jobs:
-  goreleaser:
+  kappreleaser:
+    name: kapp release
     runs-on: ubuntu-latest
-    # Set permissions of github token. See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
-    permissions:
-      contents: write
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -75,23 +100,20 @@ jobs:
...
-            ${checksums['kapp-linux-arm64']}  ./kapp-linux-arm64
-            ${checksums['kapp-windows-amd64.exe']}  ./kapp-windows-amd64.exe`
+            return `${checksums['kapp-darwin-amd64']}  kapp-darwin-amd64
+            ${checksums['kapp-darwin-arm64']}  kapp-darwin-arm64
+            ${checksums['kapp-linux-amd64']}  kapp-linux-amd64
+            ${checksums['kapp-linux-arm64']}  kapp-linux-arm64
+            ${checksums['kapp-windows-amd64.exe']}  kapp-windows-amd64.exe`
 
       - name: verify uploaded artifacts
+        if: startsWith(github.ref, 'refs/tags/')
```
joaopapereira, line 110: We have this step with goreleaser because we want to make sure that nothing extra was added to the binaries during the compilation process. Since yll are not using goreleaser I think that this and the previous step are not needed in your release process

joaopapereira (commented):
I added a comment about the steps I feel like are not needed due to ylls release process

.github/workflows/release.yml
rohitagg2020: I understand that we are trying to be extra cautious here but I think it's ok to have it. As manually also, we are building on two different machines and matching the checksum.

.github/workflows/release.yml
```diff
@@ -75,23 +100,20 @@ jobs:
...
-
-          ./hack/build-binaries.sh "$VERSION" > ./go-checksums
-          cat ./go-checksums
-          diff ./go-checksums <(cat <<EOF
+          cat ./checksums.txt
+          diff ./checksums.txt <(cat <<EOF
           ${{steps.get-checksums-from-draft-release.outputs.result}}
           EOF
-          )
+          )
```
praveenrewar: nit: add new line

hack/test-external.sh
```diff
@@ -13,8 +13,9 @@ time kapp delete -y -a istio
 time kapp deploy -y -a cert-manager -f examples/cert-manager-v1.6.1/
 time kapp delete -y -a cert-manager
 
-time kapp deploy -y -a knative -f examples/knative-v1.1.0/
-time kapp delete -y -a knative
+# TODO Add knative - Commenting because it need > resources than what github action provides.
```
praveenrewar, line 16: Question: Does it need more RAM and CPU?

.github/workflows/release.yml
praveenrewar: +1 to what @joaopapereira has mentioned, I don't see any use of having this check.

> As manually also, we are building on two different machines and matching the checksum.

I think we initially did that as we saw the checksums mismatching from goreleaser compilation and local compilation, but I don't see how this would happen while directly using the binaries built from the scripts.

rohitagg2020: Not the `if` loop, I think he is mentioning about `verifying the uploaded artifact` step.

hack/test-external.sh
rohitagg2020: Yes.

.github/workflows/release.yml
rohitagg2020: Adopted

.github/workflows/release.yml
praveenrewar: I was talking about the same `verifying the uploaded artifact`, but I am okay for being extra cautious.

praveenrewar (approved):
LGTM

100mik (approved):
LGTM!

Issue #432, opened 2022-02-14
kapp creates state configmap on --diff-run


schmurfy:
**What steps did you take:**
kapp deploy -a myapp -f . --diff-run

**What happened:**
kapp shows the diff but created an empty configmap for this app, when running `kapp list` afterwards the app is shown in the list with "-" as namespace, an inspect list nothing, as it should

**What did you expect:**
nothing changed on the cluster

**Anything else you would like to add:**
[Additional information that will assist in solving the issue.]

**Environment:**

- kapp version (use `kapp --version`): kapp version 0.42.0 (I checked but no changes on this appear on the changelogs for the following releases)
- OS (e.g. from `/etc/os-release`): osx, linux
- Kubernetes version (use `kubectl version`)
```
lient Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T18:49:28Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.12-gke.1500", GitCommit:"d32c0db9a3ccd0ac73b0b3abd0532505217b376e", GitTreeState:"clean", BuildDate:"2021-11-17T09:30:02Z", GoVersion:"go1.15.15b5", Compiler:"gc", Platform:"linux/amd64"}
```

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

100mik:
Looks like this issue is also being tracked in this one #23

rohitagg2020:
@rohitagg2020 Please check: When `diff-run` is performed against the labeled app, we don't create configmap. Can we convert recorded app to labeled app and do the `diff-run`.

schmurfy:
@rohitagg2020 What are labeled apps and recorded apps ?

cppforlife:
[...]

PR #444, opened 2022-02-23
Nothing should be changed on the cluster while doing --diff-run (#432)

sethiyash:
-  deleting configmap created by kapp during `--diff-run`


pkg/kapp/cmd/app/deploy.go
```diff
@@ -165,6 +165,11 @@ func (o *DeployOptions) Run() error {
 	}
 
 	if o.DiffFlags.Run || hasNoChanges {
+		// delete configmap created by kapp while --diff-run (#432)
+		err := app.Delete()
```
rohitagg2020, line 169: I think it will delete the full app. Can you please confirm.

Issue #235, opened 2021-06-24
remove conditions column from inspect/diff since it's misleadingly colored for resources like HPA which always have a false condition


StarpTech:
**What steps did you take:**

```
Namespace  test  
Name       test-71295  
Kind       HorizontalPodAutoscaler  
Status     conditions:  
           - lastTransitionTime: "2021-06-24T11:56:39Z"  
             message: recommended size matches current size  
             reason: ReadyForNewScale  
             status: "True"  
             type: AbleToScale  
           - lastTransitionTime: "2021-06-24T12:41:27Z"  
             message: the HPA was able to successfully calculate a replica count from cpu resource  
               utilization (percentage of request)  
             reason: ValidMetricFound  
             status: "True"  
             type: ScalingActive  
           - lastTransitionTime: "2021-06-24T11:58:24Z"  
             message: the desired count is within the acceptable range  
             reason: DesiredWithinRange  
             status: "False"  
             type: ScalingLimited  
           currentMetrics:  
           - resource:  
               current:  
                 averageUtilization: 3  
                 averageValue: 16m  
               name: cpu  
             type: Resource  
           currentReplicas: 1  
           desiredReplicas: 1  
           lastScaleTime: "2021-06-24T12:45:06Z"
```

```
Namespace           Name                              Kind                     Owner    Conds.  Rs  Ri  Age  
test  				test-71295                        HorizontalPodAutoscaler  kapp     2/3 t   ok  -   52m  
```

**What happened:**

`ScalingLimited: false` is interpreted as an error.

**What did you expect:**
ScalingLimited indicates that autoscaling is not allowed because a maximum or minimum replica count was reached.

A True condition indicates that you need to raise or lower the minimum or maximum replica count in order to scale.

A False condition indicates that the requested scaling is allowed.

Source: https://docs.openshift.com/container-platform/3.9/dev_guide/pod_autoscaling.html

**Environment:**

- kapp version (use `kapp --version`): 0.37.0
- OS (e.g. from `/etc/os-release`): Ubuntu
- Kubernetes version (use `kubectl version`): 1.19.4

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

cppforlife:
[...]

PR #445, opened 2022-02-25
#235: removed condition column from inspect table

kumaritanushree:
Removed condition column from inspect table as it was giving some misleading information in some cases where conditions are false always.


pkg/kapp/clusterapply/changes_view.go
```diff
@@ -93,16 +89,13 @@ func (v *ChangesView) Print(ui ui.UI) {
 		}
 
 		if resource.IsProvisioned() {
-			condVal := cmdcore.NewConditionsValue(resource.Status())
 
 			row = append(row,
 				// TODO erroneously colors empty value
```
praveenrewar: We could remove this and other similar TODOs since we have removed the column itself.

pkg/kapp/clusterapply/changes_view.go
```diff
@@ -93,16 +89,13 @@ func (v *ChangesView) Print(ui ui.UI) {
...
-				uitable.ValueFmt{V: condVal, Error: condVal.NeedsAttention()},
```
praveenrewar, line 100: Now that we are not using `ConditionsValue` anywhere, we could also remove [this](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/cmd/core/conditions_value.go) file.

sethiyash: +1. There is no point in keeping this file now.

pkg/kapp/clusterapply/changes_view.go
kumaritanushree: Removed

pkg/kapp/clusterapply/changes_view.go
kumaritanushree: Removed

pkg/kapp/clusterapply/changes_view.go
```diff
@@ -93,16 +89,13 @@ func (v *ChangesView) Print(ui ui.UI) {
 		}
 
 		if resource.IsProvisioned() {
-			condVal := cmdcore.NewConditionsValue(resource.Status())
 
 			row = append(row,
-				// TODO erroneously colors empty value
-				uitable.ValueFmt{V: condVal, Error: condVal.NeedsAttention()},
+
```
praveenrewar: nit: Remove new line.
I think you added a new line in all the places while removing the comments.

kumaritanushree: Removed

praveenrewar (approved):
LGTM

rohitagg2020:
LGTM

100mik (approved):
LGTM 🚀

sethiyash (commented):
Looks Good.

PR #450, opened 2022-03-02
Allow kapp deploy to migrate app configmap

neil-hickey:
Following on from https://github.com/vmware-tanzu/carvel-kapp/issues/146, this PR introduces a new flag (`--prev-app`) on the `deploy` command.

This `--prev-app` command performs the functionality of `kapp rename` but within the `deploy` command, this is important for controllers such as `kapp-controller` that cannot introduce another command on every reconciliation loop. The purpose of this flag is to optimize the `deploy` case to handle renaming / migrating when necessary without needing `kapp rename`.

This flag also interacts with the recently introduced `KAPP_FQ_CONFIGMAP_NAMES` environment variable. The tests cases are therefore pretty extensive to cover all the various ways these options can be used together. There are two main useful test cases that explicity expect a failure:

1. An app has been migrated previously, and now the user has not set `KAPP_FQ_CONFIGMAP_NAMES` or explicity set it to `false`
- https://github.com/vmware-tanzu/carvel-kapp/pull/450/files#diff-019ca500f7c4cf4ec2758417db106bda75824729b9a3884a797c6719e76714ddR192-R210
- https://github.com/vmware-tanzu/carvel-kapp/pull/450/files#diff-019ca500f7c4cf4ec2758417db106bda75824729b9a3884a797c6719e76714ddR228-R242
2. The reason for this is mainly due to the complexity of allowing migrations back and forth, the codebase is not setup in a way that this plays out nicely. Let me know if this is a deal breaker and we can work out some way to make it possible to do.

Let me know if you have any questions!

For more information / details, checkout the [doc](https://docs.google.com/document/d/1gE-gwrTeYSZ91zPYcjalwMp-zMJSOwNW1wSPTKYxLcQ/edit?usp=sharing)

### DIFF
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -219,6 +219,21 @@ func (a *RecordedApp) RenamePrevApp(prevAppName string, labels map[string]string
 	var c *corev1.ConfigMap
 	var err error
 
+	newConf := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      a.name,
+			Namespace: a.nsName,
+			Labels: map[string]string{
+				KappIsAppLabelKey: kappIsAppLabelValue,
+			},
+		},
+		Data: Meta{
+			LabelKey:   kappAppLabelKey,
+			LabelValue: fmt.Sprintf("%d", time.Now().UTC().UnixNano()),
+			UsedGKs:    &[]schema.GroupKind{},
+		}.AsData(),
+	}
+
 	if a.isMigrationEnabled() {
 		c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
 		if err == nil {
@@ -248,6 +263,14 @@ func (a *RecordedApp) RenamePrevApp(prevAppName string, labels map[string]string
 					c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), prevAppName, metav1.GetOptions{})
 					if err == nil {
 						return a.migrate(c, labels, a.fqName)
+					} else if errors.IsNotFound(err) {
+						newConf.Name = a.fqName
+						err = a.mergeAppAnnotationUpdates(newConf, map[string]string{KappIsConfigmapMigratedAnnotationKey: KappIsConfigmapMigratedAnnotationValue})
+						if err != nil {
+							return err
+						}
+
+						return a.createOrUpdate(newConf, labels)
 					}
 				}
 			}
@@ -260,11 +283,9 @@ func (a *RecordedApp) RenamePrevApp(prevAppName string, labels map[string]string
 				}
 
 				return a.updateApp(c, labels)
+			} else if errors.IsNotFound(err) {
+				return a.createOrUpdate(newConf, labels)
 			}
-		}
-
-		if errors.IsNotFound(err) {
-			return a.CreateOrUpdate(labels)
 		}
 	}
 
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -191,12 +191,7 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
 				return fmt.Errorf("Getting app: %s", err)
 			}
 
-			err = a.updateApp(existingConfigMap, labels)
-			if err != nil {
-				return err
-			}
-
-			return nil
+			return a.updateApp(existingConfigMap, labels)
 		}
 
 		return fmt.Errorf("Creating app: %s", err)
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -214,7 +214,7 @@ func (a *RecordedApp) updateApp(existingConfigMap *corev1.ConfigMap, labels map[
 	return nil
 }
 
-func (a *RecordedApp) RenamePrevApp(prevAppName string, labels map[string]string) (string, error) {
+func (a *RecordedApp) RenamePrevApp(prevAppName string, labels map[string]string) error {
 	defer a.logger.DebugFunc("RenamePrevApp").Finish()
 	var c *corev1.ConfigMap
 	var err error
@@ -222,19 +222,19 @@ func (a *RecordedApp) RenamePrevApp(prevAppName string, labels map[string]string
 	if a.isMigrationEnabled() {
 		c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
 		if err == nil {
-			return "", a.updateApp(c, labels)
+			return a.updateApp(c, labels)
 		} else if err != nil {
 			if errors.IsNotFound(err) {
 				c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
 				if err == nil {
-					return "", a.migrate(c, labels, a.fqName)
+					return a.migrate(c, labels, a.fqName)
 				}
 			}
 		}
 	} else {
 		c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
 		if err == nil {
-			return "", a.updateApp(c, labels)
+			return a.updateApp(c, labels)
 		}
 	}
 
@@ -242,13 +242,12 @@ func (a *RecordedApp) RenamePrevApp(prevAppName string, labels map[string]string
 		if a.isMigrationEnabled() {
 			c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), prevAppName+AppSuffix, metav1.GetOptions{})
 			if err == nil {
-				return "", a.migrate(c, labels, a.fqName)
+				return a.migrate(c, labels, a.fqName)
 			} else if err != nil {
 				if errors.IsNotFound(err) {
 					c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), prevAppName, metav1.GetOptions{})
 					if err == nil {
-						msg := fmt.Sprintf("Renaming '%s' (namespace: %s) to '%s' (app changes will not be renamed)", prevAppName, c.Namespace, a.name)
-						return msg, a.migrate(c, labels, a.fqName)
+						return a.migrate(c, labels, a.fqName)
 					}
 				}
 			}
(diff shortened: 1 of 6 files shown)


pkg/kapp/cmd/app/deploy_flags.go
```diff
@@ -5,6 +5,7 @@ package app
 
 import (
 	"fmt"
+
 	"github.com/spf13/cobra"
```
praveenrewar: nit: remove new line

pkg/kapp/app/recorded_app.go
```diff
@@ -196,16 +191,11 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
...
 			if err != nil {
 				return err
 			}
 
-			_, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Update(context.TODO(), existingConfigMap, metav1.UpdateOptions{})
-			if err != nil {
-				return fmt.Errorf("Updating app: %s", err)
-			}
-
 			return nil
```
praveenrewar: Should we directly return err here?

praveenrewar (commented):
I really like that we are *not* using the existing CreateOrUpdate() for prev-app :)

pkg/kapp/cmd/app/deploy_flags.go
neil-hickey: hmm, I think my IDE put that in there, will remove 👍

pkg/kapp/app/recorded_app.go
neil-hickey: absolutely. I missed that one, thanks!

pkg/kapp/cmd/app/deploy_flags.go
neil-hickey: everytime I save in VSCode it brings the line back 🙄 .

praveenrewar: Actually I was talking about the newline after `"github.com/spf13/cobra"`, which I think has been there for some time now 😅 . We usually group the imports into 2 sections (separated by a new line), where first section consists of the standard libraries provided by Go (`fmt` in this case) and rest all go into the second section. 
So the new line you had added is perfectly fine, we just need to remove the other new line.

I shouldn't have mentioned it as `nit` I guess, my bad 😅

neil-hickey: ahhhhh, gotcha. I was confused because I thought I introduced this, whereas you were asking to help fixup the existing thing. Gotcha! I fixed up it, `go fmt` rearranged the second set of imports so the named one was first also.

pkg/kapp/app/recorded_app.go
```diff
@@ -196,27 +191,83 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
...
+					c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), prevAppName, metav1.GetOptions{})
+					if err == nil {
+						return a.migrate(c, labels, a.fqName)
+					}
+				}
+			}
+		} else {
+			c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), prevAppName, metav1.GetOptions{})
+			if err == nil {
+				return a.migrate(c, labels, a.name)
```
praveenrewar: I think this would add the `KappIsConfigmapMigratedAnnotationKey` to the configmap even though migration is disabled. Maybe it was best to keep the `mergeAppAnnotationUpdates()` outside the migrate function? 
(Although I did like that right now we don't have to call it every time we use `migrate()`)

neil-hickey: This is a great callout, I updated the tests to all check for the presence / absence of that annotation and I updated this path to not use the `migrate()` function - thus keeping the annotation addition in that function.

praveenrewar:
@neil-hickey The PR seems to be in a very good state to me. However there is one point that was brought up by @renuy, that just like rename, the app-changes won't be renamed while using the `prev-app` flag. Do you think it's worth adding a confirmation just like rename?
```
kapp rename -a test-2 --new-name test-3               
Target cluster 'https://192.168.64.81:8443' (nodes: minikube)

Renaming 'test-2' (namespace: default) to 'test-3' (namespace: default) (app changes will not be renamed)

Continue? [yN]: y

Succeeded
```

I haven't given much thought to the flow but since the flag works with deploy, there's already going to be a confirmation present, so maybe just add a message (only when the rename is going to happen) ?

neil-hickey:
> I haven't given much thought to the flow but since the flag works with deploy, there's already going to be a confirmation present, so maybe just add a message (only when the rename is going to happen) ?

Interesting! Yeh I have been mostly thinking of this in the context of KC. Where there is no need for a prompt, for a kapp user I think it makes sense to have some message in the deploy confirm prompt that this migration is happening and that app changes won't be affected.

pkg/kapp/app/recorded_app.go
```diff
@@ -196,27 +191,88 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
...
+				}
 
-			return nil
+				return a.updateApp(c, labels)
+			}
 		}
 
-		return fmt.Errorf("Creating app: %s", err)
+		if errors.IsNotFound(err) {
+			return a.CreateOrUpdate(labels)
```
praveenrewar: I have been wondering about this a bit. This will result in a lot of duplicate calls when both the app doesn't exist. But I guess that's okay to have as it will only happen once when the app is deployed for the first time (vs we will have to create the configmap here itself based on the value of isMigrationEnabled which will result in code duplication).

neil-hickey: Right, I think most of this code introduces a potential for lots of calls in the worst case. But it's optimized for the case of I have an app already running in the cluster. I think the hit is ok also because it's only the first time. And tbh I am not really sure how we can avoid these duplicate calls given the stateless nature of kapp.

praveenrewar: True. I guess this should be fine.

praveenrewar (approved):
LGTM!

pkg/kapp/app/recorded_app.go
praveenrewar: I just remembered something. We should ideally *not* use a public method inside another public method. I don't think it's a rule but rather a good to have thing. We are already calling `CreateOrUpdate` in deploy after `RenamePrevApp` (although in an else), maybe we could somehow use that to do our job?

(I have already approved the PR, but this is something that struck my mind while working on something else and thought of sharing 😅 )

neil-hickey: Cool, I updated the code. It makes things a little longer in that method, but I think it's a little more readable also.

praveenrewar (approved):

pkg/kapp/app/recorded_app.go
```diff
@@ -196,27 +191,109 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
...
+			LabelKey:   kappAppLabelKey,
+			LabelValue: fmt.Sprintf("%d", time.Now().UTC().UnixNano()),
+			UsedGKs:    &[]schema.GroupKind{},
+		}.AsData(),
+	}
+
+	if a.isMigrationEnabled() {
+		c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
+		if err == nil {
+			return a.updateApp(c, labels)
```
sethiyash: Should we also check here the configMap we got is having a kapp app label `kapp.k14s.io/is-app` or not using `isKappApp()` method before updating it ? To make sure this is the configmap created by kapp only. 

As of now looks like we do this in `CreateOrUpdate()` method.

neil-hickey:
## Acceptance Criteria

##### Important Note

The word _migrated_ indicates that the configmap contains the `apps.k14s.io` suffix and the `is-configmap-migrated` annotation. 

Conversely _not migrated_ indicates the lack of these.

### Case KAPP_FQ_CONFIGMAP_NAMES=True

`KAPP_FQ_CONFIGMAP_NAMES=True kapp deploy -a APP --prev-app PREV_APP`

1. Non-existant APP && non-existant PREV_APP
   > Creates `APP`
   > `APP` is *migrated*
2. Existing unmigrated APP 
   > Renames configmap from `APP` to `APP.apps.k14s.io`
   > `APP` is *migrated*
3. Existing migrated APP
   > updates APP
   > `APP` is *migrated*
4. Non-existant APP && existing unmigrated PREV_APP
   > Renames configmap from `PREV_APP` to `APP.apps.k14s.io`
   > `APP`  is *migrated*
5. Non-existant APP && existing migrated PREV_APP
   > Renames configmap from `PREV_APP.apps.k14s.io` to `APP.apps.k14s.io`
   > `APP` is *migrated*

### Case KAPP_FQ_CONFIGMAP_NAMES=False

`KAPP_FQ_CONFIGMAP_NAMES=False kapp deploy -a APP --prev-app PREV_APP`

1. Non-existant APP && non-existant PREV_APP
   > Creates `APP` as though --prev-app was not supplied
   > `APP` is not *migrated*
2. Existing unmigrated APP 
   > Updates `APP` as though --prev-app was not supplied
   > `APP` is not *migrated*
3. Existing migrated APP
   > This is an error case - cannot go from migrated back to unmigrated app
   > Error will indicate that the app is already owned by another app
4. Non-existant APP && existing unmigrated PREV_APP
   > Renames `PREV_APP` to `APP` behaves the same as `kapp rename -a PREV_APP --new-name APP`
   > `APP` is not *migrated*
5. Non-existant APP && existing migrated PREV_APP
   > This is an error case - cannot go from migrated back to unmigrated app
   > Error will indicate that the app is already owned by another app

pkg/kapp/app/recorded_app.go
rohitagg2020: Can we use the same `a.updateApp(c, labels)` even in case of `CreateOrUpdate` function . https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L142-L147

pkg/kapp/app/recorded_app.go
```diff
@@ -196,27 +181,113 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
...
+			LabelValue: fmt.Sprintf("%d", time.Now().UTC().UnixNano()),
+			UsedGKs:    &[]schema.GroupKind{},
+		}.AsData(),
+	}
+
+	if a.isMigrationEnabled() {
+		c, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
+		if err == nil {
+			if a.isKappApp(c) {
+				return a.updateApp(c, labels)
```
rohitagg2020, line 231: Can we use the same logic here as well:
https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L141-L147

praveenrewar: I think that's already done (As part of [this](https://github.com/vmware-tanzu/carvel-kapp/pull/450/commits/e37be4d6344782ef0cbc8b1287e0f07f1c628670) commit)

rohitagg2020:
LGTM

100mik (approved):
LGTM!

sethiyash (approved):
Looks Good!

PR #453, opened 2022-03-07
WIP - Add non-blocking-wait annotation

praveenrewar:
Add non-blocking-wait annotation to enable disable waiting of a resource by the dependent resources.
Example: 
```
---
kind: Deployment
name: dep-1
...
    kapp.k14s.io/change-group: "group-1"
    kapp.k14s.io/non-blocking-wait: ""
...
---
kind: ConfigMap
name: cm-1
...
    kapp.k14s.io/change-rule: "upsert after upserting group-1"
...
```
```
...
---- applying 1 changes [0/2 done] ----
create deployment/dep-1 (batch/v1) namespace: default
---- waiting on 1 changes [0/2 done] ----
ongoing: reconcile deployment/dep-1 (batch/v1) namespace: default
^ Waiting to complete (0 active, 0 failed, 0 succeeded)
---- applying 1 changes [1/2 done] ----
create configmap/cm-1 (v1) namespace: default
---- waiting on 2 changes [0/2 done] ----
...
```

### DIFF
--- a/pkg/kapp/clusterapply/waiting_changes.go
+++ b/pkg/kapp/clusterapply/waiting_changes.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	nonBlockingOrderingAnnKey = "kapp.k14s.io/non-blocking-ordering" // value is ignored
+	nonBlockingWaitAnnKey = "kapp.k14s.io/non-blocking-wait" // value is ignored
 )
 
 type WaitingChangesOpts struct {
@@ -105,7 +105,7 @@ func (c *WaitingChanges) WaitForAny() ([]WaitingChange, error) {
 			case !state.Done:
 				newInProgressChanges = append(newInProgressChanges, change)
 
-				if _, found := change.Cluster.Resource().Annotations()[nonBlockingOrderingAnnKey]; found {
+				if _, found := change.Cluster.Resource().Annotations()[nonBlockingWaitAnnKey]; found {
 					// TODO Improve if condition
 					if !strings.Contains(state.Message, "Waiting for generation") {
 						doneChanges = append(doneChanges, change)
--- a/pkg/kapp/clusterapply/waiting_changes.go
+++ b/pkg/kapp/clusterapply/waiting_changes.go
@@ -5,7 +5,6 @@ package clusterapply
 
 import (
 	"fmt"
-	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -100,6 +105,13 @@ func (c *WaitingChanges) WaitForAny() ([]WaitingChange, error) {
 			case !state.Done:
 				newInProgressChanges = append(newInProgressChanges, change)
 
+				if _, found := change.Cluster.Resource().Annotations()[nonBlockingOrderingAnnKey]; found {
+					// TODO Improve if condition
+					if !strings.Contains(state.Message, "Waiting for generation") {
```
cppforlife: i dont think we need to check on any info here. if resource is marked as "wait is non-blocking" then we should unblock other changes, but also keep waiting for change to finish.

pkg/kapp/clusterapply/waiting_changes.go
```diff
@@ -5,13 +5,18 @@ package clusterapply
...
+	"strings"
 	"time"
 
 	ctldgraph "github.com/k14s/kapp/pkg/kapp/diffgraph"
 	ctlresm "github.com/k14s/kapp/pkg/kapp/resourcesmisc"
 	"github.com/k14s/kapp/pkg/kapp/util"
 )
 
+const (
+	nonBlockingOrderingAnnKey = "kapp.k14s.io/non-blocking-ordering" // value is ignored
```
cppforlife: how about `kapp.k14s.io/non-blocking-wait: ""`

pkg/kapp/clusterapply/waiting_changes.go
praveenrewar: > Waiting for the resource to be created/updated on the api may be sufficient, or if a wait rule defines supportsObservedGeneration: true wait for the generations to match.

The check is inline with this ^. I am also wondering if we should allow a value for the annotation to make this check opt in 🤔

pkg/kapp/clusterapply/waiting_changes.go
praveenrewar: Sounds perfect. Thanks :)

pkg/kapp/clusterapply/waiting_changes.go
praveenrewar: Removed the check, but keeping this open for now.

PR #454, opened 2022-03-08
Stopped creating kapp app configmap during --diff-run

### LINKED ISSUE #432: kapp creates state configmap on --diff-run

sethiyash:
- kapp should not create kapp app configmap while doing any operation with `--diff-run` flag


pkg/kapp/app/recorded_app.go
```diff
@@ -208,10 +214,13 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
 
 			return nil
 		}
-
 		return fmt.Errorf("Creating app: %s", err)
 	}
 
+	if isDiffRun {
```
rohitagg2020: I think we can remove the condition. It doesn't matter if it is `diffRun` or not, you should call the setMeta fn

pkg/kapp/app/recorded_app.go
```diff
@@ -115,7 +115,7 @@ func (a *RecordedApp) UpdateUsedGVsAndGKs(gvs []schema.GroupVersion, gks []schem
 	})
 }
 
-func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
+func (a *RecordedApp) CreateOrUpdate(labels map[string]string, isDiffRun bool) error {
```
rohitagg2020, line 130: @sethiyash Can somebody try to use `diffRun` while doing migration. If yes, are all those cases handled ?

pkg/kapp/app/recorded_app.go
sethiyash: We can but if we call setMeta for all we have to modify our [`meta()` method](https://github.com/vmware-tanzu/carvel-kapp/blob/6f5f582e71bc41e8c5fb53af5f1c6a4cb57d8d02/pkg/kapp/app/recorded_app.go#L412) to first check for migrated app and then check for memoizedMeta is non nil

sethiyash: As if we start setting meta here then `a.memoizedMeta` will never be nil and hence even when migration is enabled it `a.isMigrated ` will not get set to `True` and hence will unable to find the app in `Rename()` method as now we have to get configMap with fully qualified name.

sethiyash: And if we start setting the `a.isMigrated` flag here based on ` a.isMigrationEnabled()` then it may be a reduplication of logic

pkg/kapp/app/recorded_app.go
sethiyash: tested with different scenario.

sethiyash: Resource| KAPP_FQ_CONFIGMAP_NAMES | diff-run | create  | update  | delete    | CM created |
| --------- | ---------------- | -------- | ------- | ------- | --------- | ------- |
| new       | False            | no       | Success | Success | Success   | yes     |
| new       | False            | yes      | Success | Success | Success | no      |
| new       | True         | no       |   Success | Success        |   Success        |   yes      |
| new       |  True         | yes|  Success |   Success      |   Success        |  No       |
| existing  | True     |    yes |  Success  |   Success     |   Success        |     No    |
| existing       | True             | no      |   Success      |     Success    |  Success         |      yes   |

cppforlife:
would recommend blocking merge of this issue on https://github.com/vmware-tanzu/carvel-kapp/pull/450

pkg/kapp/app/recorded_app.go
```diff
@@ -187,6 +195,8 @@ func (a *RecordedApp) createOrUpdate(c *corev1.ConfigMap, labels map[string]stri
 		return fmt.Errorf("Creating app: %s", err)
 	}
 
+	a.setMeta(*configmap)
```
sethiyash, line 198: Started setting `meta` here as in case of `--diff-run` we will unable to get the configmap later. In without `--diff-run` scenario it will reduce the `get` calls.

100mik:
I think we can rebase this on develop now.

pkg/kapp/app/recorded_app.go
```diff
@@ -162,18 +162,26 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 		}
 
 		configMap.ObjectMeta.Name = a.fqName
+		// setting migration as true as we will be creating configmap with suffix
+		a.isMigrated = true
```
sethiyash, line 166: Started setting `isMigrated` flag to true here as we will be creating configmap with fully qualified name and in case of `--diff-run` we unable to get the configmap and the flag will never set as we do [here](https://github.com/vmware-tanzu/carvel-kapp/blob/d667147b688102aea0f729741c2841c35e1015d2/pkg/kapp/app/recorded_app.go#L481).

pkg/kapp/app/recorded_app.go
sethiyash: @rohitagg2020 we can't start setting meta everytime as in case when migration is enabled (i.e `KAPP_FQ_CONFIGMAP_NAMES=True`) and we set the meta here then `a.isMigrated` flag always remain false as it will always read from cache (i.e. memoizedMeta) and hence won't set the  `a.isMigrated` flag to true and will try to get the configmap with non suffixed name and won't able to find it and give an error. 

In my opinion, let's just do it only in case of diff-run and let remain the logic of setMeta in normal flow intact. 

Letting it open for the feedback.

praveenrewar: > we can't start setting meta everytime as in case when migration is enabled (i.e KAPP_FQ_CONFIGMAP_NAMES=True) and we set the meta here then a.isMigrated flag always remain false

With the new changes (i.e. always check for migrated app first) merged, we will have to set `isMigrated` in more than 1 place in `CreateOrUpdate()` and `RenamePrevApp` as compared to only 1 place.

sethiyash: I'm no more setting `isMigrated` anywhere we need to do that when we were calling `setMeta()` for each every call but now as I'm doing it just in case of `--diff-run`  so not needed.

praveenrewar: Any specific reason for not setting meta in all scenarios?

sethiyash: Setting meta in for all scenarios is making an issue in the case when there is an existing migrated app and we try to update it making `KAPP_FQ_CONFIGMAP_NAMES` env variable False. (basically in [this](https://github.com/vmware-tanzu/carvel-kapp/blob/d667147b688102aea0f729741c2841c35e1015d2/test/e2e/create_update_delete_test.go#L194) test scenario)
 
Then to handle this case we have to start setting `a.isMigrated` flag to True in `CreateOrUpdate()` and `RenamePrevApp()` method and it can be controversial as we are making `RecordedApp` as migrated before even creating it that's why removed setting meta for all calls. 

Please let me know your thoughts also on this. 

cc: @praveenrewar

praveenrewar: > it can be controversial as we are making RecordedApp as migrated before even creating it

IMO this is not controversial as we will be setting `isMigrated` to true only when we are sure that app that will be created is going to be a migrated app (not that it will be created first and then migrated)

praveenrewar: The only concern I have is that we will have to set `isMigrated` to true in multiple scenarios.

sethiyash: Yeah we have to set it in both  `CreateOrUpdate()` and `RenamePrevApp()` methods.

pkg/kapp/app/recorded_app.go
```diff
@@ -197,7 +197,15 @@ func (a *RecordedApp) create(labels map[string]string) error {
...
 	}
 
-	_, err = a.coreClient.CoreV1().ConfigMaps(a.nsName).Create(context.TODO(), configMap, metav1.CreateOptions{})
+	var dryRunValue []string = nil
+
+	if isDiffRun {
+		dryRunValue = []string{metav1.DryRunAll}
+	}
+
+	app, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Create(context.TODO(), configMap, metav1.CreateOptions{DryRun: dryRunValue})
```
praveenrewar: nit
```suggestion
	createOpts := metav1.CreateOptions{}
	if isDiffRun {
		createOpts.DryRun = []string{metav1.DryRunAll}
	}

	app, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Create(context.TODO(), configMap, createOpts)
```

test/e2e/diff_test.go
```diff
@@ -375,3 +375,30 @@ data:
 
 	require.Containsf(t, out, expectedOutput, "Did not find expected diff output")
 }
+
+func TestKappNotCreateConfigmapOnDiffRun(t *testing.T) {
```
praveenrewar:
```suggestion
func TestDiffRun(t *testing.T) {
```
I think we don't have an existing test that asserts _just_ diff run output. Should we also add that to the same test?

praveenrewar:
I am thinking if the effects of `--diff-run` should just be limited to creation of ConfigMap or should they also extend to updates as well. For example, if we use `--labels x=y`, along with `--diff-run` while updating an app, should these labels be added to the ConfigMap?

neil-hickey:
> I am thinking if the effects of `--diff-run` should just be limited to creation of ConfigMap or should they also extend to updates as well. For example, if we use `--labels x=y`, along with `--diff-run` while updating an app, should these labels be added to the ConfigMap?

I would expect nothing to change on the cluster if I use `--diff-run`. No configmaps created, no labels added, no changes. So in my mind, the answer to should the labels be added is no.

kumaritanushree:
LGTM

sethiyash:
Hi @praveenrewar, 
As @neil-hickey correctly pointed out in the case of `--diff-run` ideally nothing should change on the cluster but currently when there is an update operation and the user do - `kapp deploy -a test -f test.yaml --labels x=y --diff-run` we are actually updating the configmap with labels even if it's a diff-run. 

Please let me know your thoughts on changing this behaviour to not updating the configmap as part of this PR.

neil-hickey:
> Please let me know your thoughts on changing this behaviour to not updating the configmap as part of this PR.

I'd vote for a separate issue / PR. This one should probably remain scoped to just the creation of the configmap.

neil-hickey (approved):
LGTM

sethiyash:
After discussion with the team, we have decided to address the update flow as part of this PR only.

pkg/kapp/app/recorded_app.go
```diff
@@ -137,7 +137,7 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 
 	if foundMigratedApp {
 		a.isMigrated = true
-		return a.updateApp(app, labels)
+		return a.updateApp(app, labels, isDiffRun)
```
praveenrewar: Maybe just return if `isDiffRun`?

pkg/kapp/app/recorded_app.go
```diff
@@ -146,13 +146,13 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 	}
 
 	if foundNonMigratedApp {
-		if a.isMigrationEnabled() {
+		if a.isMigrationEnabled() && !isDiffRun {
```
praveenrewar: At this point, we would wanna `return nil` if `isDiffRun`, otherwise it will move to the next part of the code.

test/e2e/diff_test.go
```diff
@@ -375,3 +375,30 @@ data:
...
+	defer cleanUp()
+
+	yaml1 := `
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config
+`
+
+	_, _ = kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name,
```
100mik, line 399: I think we could add another case for the update scenario (with labels) to have more coverage

100mik (commented):
No super strong feelings about this one, just a thought

test/e2e/diff_test.go
100mik: Create app -> update app -> Verify that labels were not added

praveenrewar (commented):
A couple of thoughts I had while taking a look.
- If we are making sure that diff-run doesn't result in any change in the cluster, then we might have to make changes to the `UsedGVs` and `UsedGKs` because they will also be updating the ConfigMap. 
- Since we will have to make changes in so many places, it might make sense to have the `isDiffRun` in the RecordedApp itself so that we don't have to pass it along in so many places and use it just when required.
- Another thing that might be worth exploring is if it's possible to have the dry run set for the whole client itself (but I guess that's not possible for the core client).

(I am totally okay with creating a separate PR for the updates)

pkg/kapp/app/recorded_app.go
sethiyash: Yeah, that may be better. I will address this in a separate PR for update scenarios.

pkg/kapp/app/recorded_app.go
sethiyash: It will go to update only and there is no change that will happen when it's diff-run. But yeah returning nil earlier itself may be better. I will make the change in a separate PR.

test/e2e/diff_test.go
sethiyash: Yeah sure, I will add a test case for this scenario also in a separate PR.

pkg/kapp/app/recorded_app.go
sethiyash: I think It will be fine as it will go ahead and call `updateApp()` and as the DiffRun is True the update will tale place with DryRun and nothing will change on the cluster. 

I believe [here](https://github.com/vmware-tanzu/carvel-kapp/blob/4a6a3a60851b1425cbdb4acb40566d4e7e788d72/pkg/kapp/app/recorded_app.go#L259) we have to take care that to `return nil` as both will gonna make changes on the cluster

test/e2e/diff_test.go
sethiyash: Done.

pkg/kapp/app/recorded_app.go
sethiyash: started checking for `diffRun` inside `updateApp()` method itself

praveenrewar (approved):
LGTM!

PR #455, opened 2022-03-10
Fix crash when applying rebase rule for nil map key

nicholasmhughes:
Found an issue when attempting to apply the Rancher 2.6 Helm chart via kapp.

The template for the deployment has logic which can result in an empty/nil value for the `annotations` metadata key.

```
metadata:
  name: {{ template "rancher.fullname" . }}
  annotations:
{{- if (lt (int .Values.replicas) 0) }}
    management.cattle.io/scale-available: "{{ sub 0 (int .Values.replicas)}}"
{{- end  }}
  labels:
{{ include "rancher.labels" . | indent 4 }}
```

kapp is unable to deploy the chart when the annotations are empty, throwing this error:

```
kapp: Error: Applying rebase rule to deployment/rancher (apps/v1) namespace: cattle-system:
  FieldCopyMod for path 'metadata,annotations,deployment.kubernetes.io/revision' on resource 'deployment/rancher (apps/v1) namespace: cattle-system':
    Unexpected non-map found: <nil>
```

The type checking for key values should allow nil, as a map can fully overwrite the empty value.

### DIFF
--- a/pkg/kapp/diff/change_set_test.go
+++ b/pkg/kapp/diff/change_set_test.go
@@ -267,7 +267,7 @@ metadata:
 	require.Equal(t, expectedDiff, actualDiff, "Expected diff to match")
 }
 
-func TestChangeSet_RebaseWithNilNew_And_UnexpectedChanges(t *testing.T) {
+func TestRebaseWithNilNewAndUnexpectedChanges(t *testing.T) {
 	newRes := ctlres.MustNewResourceFromBytes([]byte(`
 metadata:
   name: my-res
--- a/pkg/kapp/diff/change_set_test.go
+++ b/pkg/kapp/diff/change_set_test.go
@@ -266,3 +266,60 @@ metadata:
 
 	require.Equal(t, expectedDiff, actualDiff, "Expected diff to match")
 }
+
+func TestChangeSet_RebaseWithNilNew_And_UnexpectedChanges(t *testing.T) {
+	newRes := ctlres.MustNewResourceFromBytes([]byte(`
+metadata:
+  name: my-res
+  annotations:
+`))
+
+	existingRes := ctlres.MustNewResourceFromBytes([]byte(`
+metadata:
+  name: my-res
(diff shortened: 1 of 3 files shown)


sethiyash:
Hi @nicholasmhughes,
Thank you so much for the contribution. This looks like a reasonable change to me. We will discuss and put our thoughts on this shortly.

sethiyash:
LGTM! As we are already handling a `nil` scenario in [remove strategy](https://github.com/vmware-tanzu/carvel-kapp/blob/6f5f582e71bc41e8c5fb53af5f1c6a4cb57d8d02/pkg/kapp/resources/mod_field_remove.go#L42). Good to see this handled in copy strategy as well.

sethiyash:
LGTM! As we are already handling a `nil` scenario in [remove strategy](https://github.com/vmware-tanzu/carvel-kapp/blob/6f5f582e71bc41e8c5fb53af5f1c6a4cb57d8d02/pkg/kapp/resources/mod_field_remove.go#L42). Good to see this handled in copy strategy as well.

sethiyash (approved):

pkg/kapp/diff/change_set_test.go
```diff
@@ -266,3 +266,60 @@ metadata:
 
 	require.Equal(t, expectedDiff, actualDiff, "Expected diff to match")
 }
+
+func TestChangeSet_RebaseWithNilNew_And_UnexpectedChanges(t *testing.T) {
```
100mik: We generally stick to Pascal Case for test names. How does `TestRebaseWithNilNewAndUnexpectedChanges` sound?

(not a strong opinion, so you could stick with this if it makes more sense)

nicholasmhughes: I can definitely change that. I just copied the format from the other tests above, so I'm not opinionated one way or another.  :smile:

rohitagg2020:
LGTM

PR #468, opened 2022-03-25
452: handle delete strategy for inoperable resources

### LINKED ISSUE #452: Allow `kapp delete` even if default namespace was part of the application configuration
If the default namespace is included as a resource in a kapp-deployed application, you can then not delete the application. You get the following error and you are stuck:

```
Error: resource deletion failed: kapp: Error: Applying delete namespace/default (v1) cluster:
  Deleting resource namespace/default (v1) cluster: namespaces "default" is forbidden: this namespace may not be deleted (reason: Forbidden)
. Delete failed: Error (see .status.usefulErrorMessage for details)
Error: exit status 1
```

I don't think it is realistic to prevent the default namespace to be part of an application configuration, so maybe one of these two approaches could work:
- If default namespace is one of the resources in the YAML, ignore it but deploy the rest of the resources in the application
- If default namespace is one of the resources in the kapp-deployed application, do not try to delete the namespace during `kapp delete...`. Simply ignore it and delete the rest of the application resources.

Here is one case in which I ran into this problem:

[https://github.com/vmware-tanzu/community-edition/issues/3067](https://github.com/vmware-tanzu/community-edition/issues/3067)

It is not the only situation in which I've experienced this issue, but I think it's a good example.

kumaritanushree:
**Why this PR and what it is doing**
Issue was:  Inoperable resources like default namespace are not allowed to delete which was throwing error on kapp delete.
Solution:  For all inoperable resources kapp will orphaned the requested resource instead of trying to delete.

TODO:
- [x]  Manual testing for the changes 
- [x] Add e2e test case

NOTE: If you feel there should be more resources in the list please suggest.

### DIFF
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -44,6 +44,18 @@ var inoperableResourceList = []uniqueResourceRef{
 		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
 		Name:      "default",
 	},
+	{
+		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		Name:      "kube-node-lease",
+	},
+	{
+		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		Name:      "kube-public",
+	},
+	{
+		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		Name:      "kube-system",
+	},
 }
 
 func (c DeleteChange) ApplyStrategy() (ApplyStrategy, error) {
@@ -118,7 +130,7 @@ func (c DeleteOrphanStrategy) Op() ClusterChangeApplyStrategyOp { return deleteS
 func (c DeleteOrphanStrategy) Apply() error {
 	mergePatch := []interface{}{
 		// TODO currently we do not account for when '-a label:foo=bar' used
-		//
+		// Reason: In labeled app labels are not accessible in delete operation now. Without the label info kapp can not apply the changes
 		map[string]interface{}{
 			"op":   "remove",
 			"path": "/metadata/labels/" + jsonPointerEncoder.Replace(appLabelKey),
--- a/pkg/kapp/clusterapply/delete_change.go
+++ b/pkg/kapp/clusterapply/delete_change.go
@@ -33,34 +33,33 @@ type DeleteChange struct {
 	identifiedResources ctlres.IdentifiedResources
 }
 
-type uniqueResourceRef struct {
+type inoperableResourceRef struct {
 	schema.GroupKind
 	Name string
 }
 
-// if any new reource will encountered which kapp can not delete and required to orphaned then resource info will be added here.
-var inoperableResourceList = []uniqueResourceRef{
+// List of resources use the "orphan" delete strategy implicitly as they cannot be deleted by end users
+var inoperableResourceList = []inoperableResourceRef{
 	{
-		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		GroupKind: schema.GroupKind{Group: "", Kind: "Namespace"},
 		Name:      "default",
 	},
 	{
-		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		GroupKind: schema.GroupKind{Group: "", Kind: "Namespace"},
 		Name:      "kube-node-lease",
 	},
 	{
-		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		GroupKind: schema.GroupKind{Group: "", Kind: "Namespace"},
 		Name:      "kube-public",
 	},
 	{
-		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		GroupKind: schema.GroupKind{Group: "", Kind: "Namespace"},
 		Name:      "kube-system",
 	},
 }
 
 func (c DeleteChange) ApplyStrategy() (ApplyStrategy, error) {
 	res := c.change.ExistingResource()
-
 	strategy := res.Annotations()[deleteStrategyAnnKey]
 
 	if c.isInoperableResource() {
@@ -162,10 +161,8 @@ func descMessage(res ctlres.Resource) []string {
 func (c DeleteChange) isInoperableResource() bool {
 
 	res := c.change.ExistingResource()
-
 	for _, r := range inoperableResourceList {
-		strings.EqualFold(r.Name, res.Name())
-		if strings.EqualFold(r.Name, res.Name()) && strings.EqualFold(r.Kind, res.GroupKind().Kind) && strings.EqualFold(r.Group, res.GroupKind().Group) {
+		if r.Name == res.Name() && r.Kind == res.GroupKind().Kind && r.Group == res.GroupKind().Group {
 			return true
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/kapp/clusterapply/delete_change.go
```diff
@@ -122,3 +139,12 @@ func descMessage(res ctlres.Resource) []string {
 	}
 	return []string{}
 }
+
+func isInoperableResource(res ctlres.Resource) bool {
```
100mik: We should scope this function to DeleteChange: `func (c DeleteChange) isInoperableResource(...`

kumaritanushree: updated

vrabbi:
i think this list should also include the namespaces:
kube-system
kube-public
kube-node-lease

these are all default namespaces in a cluster and as such should be treated i believe the same as the default namespace in terms of deletion strategy behavior

kumaritanushree:
> i think this list should also include the namespaces: kube-system kube-public kube-node-lease
> 
> these are all default namespaces in a cluster and as such should be treated i believe the same as the default namespace in terms of deletion strategy behavior

Updated the list as suggested. Thanks @vrabbi

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -122,3 +158,16 @@ func descMessage(res ctlres.Resource) []string {
 	}
 	return []string{}
 }
+
+func (c DeleteChange) isInoperableResource() bool {
+
+	res := c.change.ExistingResource()
+
+	for _, r := range inoperableResourceList {
+		strings.EqualFold(r.Name, res.Name())
```
renuy: ??

kumaritanushree: EqualFold do the non case-sensitive comparison of two strings. Which I found better than converting strings in one case and then compare.

100mik: If we have decide3d to go ahead with matching Group and Kind. We can change the values to have capital case (i.e. "Namespace") as that is always the case and go ahead with simply checking for equality

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -32,10 +33,40 @@ type DeleteChange struct {
 	identifiedResources ctlres.IdentifiedResources
 }
 
+type uniqueResourceRef struct {
+	schema.GroupKind
+	Name string
+}
+
+// if any new reource will encountered which kapp can not delete and required to orphaned then resource info will be added here.
```
100mik: nit: (just trying to make it a bit clearer)
```suggestion
// List of resources use the "orphan" delete strategy implicitly as they cannot be deleted by end users
```

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -32,10 +33,40 @@ type DeleteChange struct {
 	identifiedResources ctlres.IdentifiedResources
 }
 
+type uniqueResourceRef struct {
```
100mik: How about we change this to "inoperableResourceRef" as we are loosely matching it without being strict about versions now? It is also not being used anywhere else.
(Others might have other ideas as well)

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -32,10 +33,40 @@ type DeleteChange struct {
...
+	{
+		GroupKind: schema.GroupKind{Group: "", Kind: "namespace"},
+		Name:      "kube-system",
+	},
+}
+
 func (c DeleteChange) ApplyStrategy() (ApplyStrategy, error) {
 	res := c.change.ExistingResource()
+
 	strategy := res.Annotations()[deleteStrategyAnnKey]
```
100mik, line 64: nit: do not think we need this new line

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -122,3 +158,16 @@ func descMessage(res ctlres.Resource) []string {
...
 	return []string{}
 }
+
+func (c DeleteChange) isInoperableResource() bool {
+
+	res := c.change.ExistingResource()
+
+	for _, r := range inoperableResourceList {
+		strings.EqualFold(r.Name, res.Name())
+		if strings.EqualFold(r.Name, res.Name()) && strings.EqualFold(r.Kind, res.GroupKind().Kind) && strings.EqualFold(r.Group, res.GroupKind().Group) {
```
100mik: As renu pointed out, I do not think we should do a match which ignores cases here if we have decided on using Group-Kinds to match as they are always the same.

test/e2e/delete_inoperable_test.go
```diff
@@ -0,0 +1,75 @@
...
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	kubectl := Kubectl{t, env.Namespace, logger}
+
+	yaml1 := `
+--- 
+apiVersion: v1
+kind: Namespace
+metadata: 
+  name: default
+
```
100mik: nit: we do not need this empty line

100mik (commented):
Just some thoughts 🤔 
will take a closer look at the test as it is slightly diferent for what we have for the orphaning test.

100mik:
The failing lint should be fixed by #470

pkg/kapp/clusterapply/delete_change.go
kumaritanushree: Updated

pkg/kapp/clusterapply/delete_change.go
kumaritanushree: Agree with the suggestion and updated the same. thanks

pkg/kapp/clusterapply/delete_change.go
kumaritanushree: thanks for pointing this out, removed it.

test/e2e/delete_inoperable_test.go
kumaritanushree: removed it

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -122,3 +157,14 @@ func descMessage(res ctlres.Resource) []string {
 	}
 	return []string{}
 }
+
+func (c DeleteChange) isInoperableResource() bool {
+
```
rohitagg2020: nit: No need of new line

100mik (approved):
LGTM!

I see that there is an extra case (one more the orphaned test)
But I believe it addresses two cases, one where the GVR of the resource being deleted is available and the other when it is not.

I am fine with having this 🚀

pkg/kapp/clusterapply/delete_change.go
```diff
@@ -122,3 +157,14 @@ func descMessage(res ctlres.Resource) []string {
...
+	res := c.change.ExistingResource()
+	for _, r := range inoperableResourceList {
+		if r.Name == res.Name() && r.Kind == res.GroupKind().Kind && r.Group == res.GroupKind().Group {
```
sethiyash: > 
```suggestion
		if r.Name == res.Name() && r.Kind == res.Kind() && r.Group == res.APIGroup() {
```
Looks like we have separate methods for Kind and Group so better to use them.

cppforlife: here is a trick: 

```
if r == inoperableResourceRef{Name: res.Name(), GroupKind: res.GroupKind()} {
```

pkg/kapp/clusterapply/delete_change.go
kumaritanushree: removed

pkg/kapp/clusterapply/delete_change.go
kumaritanushree: Updated

rohitagg2020:
LGTM

PR #470, opened 2022-04-01
Explicit set Go version for linting in CI

pivotaljohn:
On Friday, 4/1, `ytt` lint GH actions started to fail. Details: https://kubernetes.slack.com/archives/CH8KCCKA5/p1648847055663899

It appears that `kapp` is using the same `golangci-lint-action` version+configuration.

Here's an attempt to help save all y'all from having to do the same troubleshooting. 👍🏻


100mik (approved):
Thanks for putting this in John!
LGTM 🚀

PR #475, opened 2022-04-07
Defining wait rules for custom resources 

### LINKED ISSUE #442: Add waitRules for Custom Resources that don't have the "conditions" field in their status.
kapp's waitRules currently only supports conditionMatchers for resources that have the "Conditions" field in their status. We are using kapp to deploy a Custom Resource that has no "Conditions" field in its status so we are not able to detect if the CR actually got deployed successfully. The CR does have a Status field though which could be used for adding waitRules

A waitRules configuration that allows users to configure arbitrary rules. For example if the CR has this in its status field:
`status:
  currentState: Running
`
the conditionMatchers could be 
`- type: currentState
     statusMessage: Running
`
The above is just a suggestion but because this could be arbitrary, the conditionMatcher may need to support more complex data structures. 

---
Vote on this request

sethiyash:
- kapp's waitRules currently only supports conditionMatchers for resources that have the "Conditions" field in their status. We now added a support to define the waitRules for custom resources in Config under `rules.star` file.

### DIFF
--- a/pkg/kapp/config/config.go
+++ b/pkg/kapp/config/config.go
@@ -44,6 +44,7 @@ type WaitRule struct {
 	SupportsObservedGeneration bool
 	ConditionMatchers          []WaitRuleConditionMatcher
 	ResourceMatchers           []ResourceMatcher
+	Ytt                        *WaitRuleYtt
 }
 
 type WaitRuleConditionMatcher struct {
@@ -52,6 +53,18 @@ type WaitRuleConditionMatcher struct {
 	Failure                    bool
 	Success                    bool
 	SupportsObservedGeneration bool
+}
+
+type WaitRuleYtt struct {
+	// Contracts are named and versioned (eg v1)
+	// to provide a stable interface to rule authors.
+	// Multiple contracts will be offered at the same time
+	// so that existing rules do not not break as we decide to evolve running environment.
+	WaitRuleContractV1 *WaitRuleYttContractV1 `json:"waitRuleContractV1"`
+}
+
+type WaitRuleYttContractV1 struct {
+	RulesStar string `json:"rules.star"`
 }
 
 type RebaseRule struct {
@@ -208,6 +221,21 @@ func (r RebaseRule) Validate() error {
 	}
 	if len(r.Path) == 0 && len(r.Paths) == 0 {
 		return fmt.Errorf("Expected either path or paths to be specified")
+	}
+	return nil
+}
+
+func (r WaitRule) AsMods() *yttresmod.WaitRuleContractV1Mod {
+	if r.Ytt != nil {
+		switch {
+		case r.Ytt.WaitRuleContractV1 != nil:
+			return &yttresmod.WaitRuleContractV1Mod{
+				ResourceMatcher: ctlres.AnyMatcher{
+					Matchers: ResourceMatchers(r.ResourceMatchers).AsResourceMatchers(),
+				},
+				Starlark: r.Ytt.WaitRuleContractV1.RulesStar,
+			}
+		}
 	}
 	return nil
 }
--- a/pkg/kapp/config/config.go
+++ b/pkg/kapp/config/config.go
@@ -60,10 +60,10 @@ type WaitRuleYtt struct {
 	// to provide a stable interface to rule authors.
 	// Multiple contracts will be offered at the same time
 	// so that existing rules do not not break as we decide to evolve running environment.
-	WaitRuleContractV1 *WaitRuleYttContractV1 `json:"waitRuleContractV1"`
+	FuncContractV1 *funcContractV1 `json:"funcContractV1"`
 }
 
-type WaitRuleYttContractV1 struct {
+type funcContractV1 struct {
 	RulesStar string `json:"rules.star"`
 }
 
@@ -225,15 +225,15 @@ func (r RebaseRule) Validate() error {
 	return nil
 }
 
-func (r WaitRule) AsMods() *yttresmod.WaitRuleContractV1Mod {
+func (r WaitRule) GetContract() *yttresmod.WaitRuleContractV1 {
 	if r.Ytt != nil {
 		switch {
-		case r.Ytt.WaitRuleContractV1 != nil:
-			return &yttresmod.WaitRuleContractV1Mod{
+		case r.Ytt.FuncContractV1 != nil:
+			return &yttresmod.WaitRuleContractV1{
 				ResourceMatcher: ctlres.AnyMatcher{
 					Matchers: ResourceMatchers(r.ResourceMatchers).AsResourceMatchers(),
 				},
-				Starlark: r.Ytt.WaitRuleContractV1.RulesStar,
+				Starlark: r.Ytt.FuncContractV1.RulesStar,
 			}
 		}
 	}
--- a/pkg/kapp/config/config.go
+++ b/pkg/kapp/config/config.go
@@ -60,11 +60,11 @@ type WaitRuleYtt struct {
 	// to provide a stable interface to rule authors.
 	// Multiple contracts will be offered at the same time
 	// so that existing rules do not not break as we decide to evolve running environment.
-	FuncContractV1 *funcContractV1 `json:"funcContractV1"`
+	FuncContractV1 *FuncContractV1 `json:"funcContractV1"`
 }
 
-type funcContractV1 struct {
-	RulesStar string `json:"rules.star"`
+type FuncContractV1 struct {
+	Resource string `json:"resource.star"`
 }
 
 type RebaseRule struct {
@@ -221,21 +221,6 @@ func (r RebaseRule) Validate() error {
 	}
 	if len(r.Path) == 0 && len(r.Paths) == 0 {
 		return fmt.Errorf("Expected either path or paths to be specified")
-	}
-	return nil
-}
-
-func (r WaitRule) GetContract() *yttresmod.WaitRuleContractV1 {
-	if r.Ytt != nil {
-		switch {
-		case r.Ytt.FuncContractV1 != nil:
-			return &yttresmod.WaitRuleContractV1{
-				ResourceMatcher: ctlres.AnyMatcher{
-					Matchers: ResourceMatchers(r.ResourceMatchers).AsResourceMatchers(),
-				},
-				Starlark: r.Ytt.FuncContractV1.RulesStar,
-			}
-		}
 	}
 	return nil
 }
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -42,16 +42,6 @@ type customWaitingResourceCondition struct {
 }
 
 func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
-	if mod := s.waitRule.AsMods(); mod != nil {
-		configObj, err := mod.ApplyYttWaitRule(s.resource)
-		if err != nil {
-			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
-				"Error: Applying ytt wait rule: %s", err.Error())}
-		}
-		return DoneApplyState{Done: configObj.Result.Done, Successful: configObj.Result.Successful,
-			Message: configObj.Result.Message}
-	}
-
 	obj := customWaitingResourceStruct{}
 
 	err := s.resource.AsUncheckedTypedObj(&obj)
@@ -63,6 +53,16 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 	if s.waitRule.SupportsObservedGeneration && obj.Metadata.Generation != obj.Status.ObservedGeneration {
 		return DoneApplyState{Done: false, Message: fmt.Sprintf(
 			"Waiting for generation %d to be observed", obj.Metadata.Generation)}
+	}
+
+	if contract := s.waitRule.GetContract(); contract != nil {
+		configObj, err := contract.ApplyYttWaitRule(s.resource)
+		if err != nil {
+			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
+				"Error: Applying ytt wait rule: %s", err.Error())}
+		}
+		return DoneApplyState{Done: configObj.Done, Successful: configObj.Successful,
+			Message: configObj.Message}
 	}
 
 	hasConditionWaitingForGeneration := false
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -60,7 +60,7 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			ResourceMatcher: ctlres.AnyMatcher{
 				Matchers: ctlconf.ResourceMatchers(s.waitRule.ResourceMatchers).AsResourceMatchers()},
 			Starlark: s.waitRule.Ytt.FuncContractV1.Resource,
-		}.ApplyYttWaitRule(s.resource)
+		}.applyYttWaitRule(s.resource)
 		if err != nil {
 			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
 				"Error: Applying ytt wait rule: %s", err.Error())}
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -8,7 +8,6 @@ import (
 
 	ctlconf "github.com/k14s/kapp/pkg/kapp/config"
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
-	"github.com/k14s/kapp/pkg/kapp/yttresmod"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -57,9 +56,11 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 	}
 
 	if s.waitRule.Ytt != nil {
-		configObj, err := yttresmod.WaitRuleContractV1{ResourceMatcher: ctlres.AnyMatcher{
-			Matchers: ctlconf.ResourceMatchers(s.waitRule.ResourceMatchers).AsResourceMatchers()},
-			Starlark: s.waitRule.Ytt.FuncContractV1.Resource}.ApplyYttWaitRule(s.resource)
+		configObj, err := WaitRuleContractV1{
+			ResourceMatcher: ctlres.AnyMatcher{
+				Matchers: ctlconf.ResourceMatchers(s.waitRule.ResourceMatchers).AsResourceMatchers()},
+			Starlark: s.waitRule.Ytt.FuncContractV1.Resource,
+		}.ApplyYttWaitRule(s.resource)
 		if err != nil {
 			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
 				"Error: Applying ytt wait rule: %s", err.Error())}
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -42,6 +42,16 @@ type customWaitingResourceCondition struct {
 }
 
 func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
+	if mod := s.waitRule.AsMods(); mod != nil {
+		configObj, err := mod.ApplyYttWaitRule(s.resource)
+		if err != nil {
+			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
+				"Error: Applying ytt wait rule: %s", err)}
+		}
+		return DoneApplyState{Done: configObj.Result.Done, Successful: configObj.Result.Successful,
+			Message: configObj.Result.Message}
+	}
+
 	obj := customWaitingResourceStruct{}
 
 	err := s.resource.AsUncheckedTypedObj(&obj)
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -46,7 +46,7 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 		configObj, err := mod.ApplyYttWaitRule(s.resource)
 		if err != nil {
 			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
-				"Error: Applying ytt wait rule: %s", err)}
+				"Error: Applying ytt wait rule: %s", err.Error())}
 		}
 		return DoneApplyState{Done: configObj.Result.Done, Successful: configObj.Result.Successful,
 			Message: configObj.Result.Message}
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -60,7 +60,7 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			ResourceMatcher: ctlres.AnyMatcher{
 				Matchers: ctlconf.ResourceMatchers(s.waitRule.ResourceMatchers).AsResourceMatchers()},
 			Starlark: s.waitRule.Ytt.FuncContractV1.Resource,
-		}.applyYttWaitRule(s.resource)
+		}.Apply(s.resource)
 		if err != nil {
 			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
 				"Error: Applying ytt wait rule: %s", err.Error())}
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -8,6 +8,7 @@ import (
 
 	ctlconf "github.com/k14s/kapp/pkg/kapp/config"
 	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+	"github.com/k14s/kapp/pkg/kapp/yttresmod"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -55,8 +56,10 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			"Waiting for generation %d to be observed", obj.Metadata.Generation)}
 	}
 
-	if contract := s.waitRule.GetContract(); contract != nil {
-		configObj, err := contract.ApplyYttWaitRule(s.resource)
+	if s.waitRule.Ytt != nil {
+		configObj, err := yttresmod.WaitRuleContractV1{ResourceMatcher: ctlres.AnyMatcher{
+			Matchers: ctlconf.ResourceMatchers(s.waitRule.ResourceMatchers).AsResourceMatchers()},
+			Starlark: s.waitRule.Ytt.FuncContractV1.Resource}.ApplyYttWaitRule(s.resource)
 		if err != nil {
 			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
 				"Error: Applying ytt wait rule: %s", err.Error())}
--- a/pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
+++ b/pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
@@ -1,7 +1,7 @@
 // Copyright 2020 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-package yttresmod
+package resourcesmisc
 
 import (
 	"fmt"
@@ -18,17 +18,17 @@ type WaitRuleContractV1 struct {
 	Starlark        string
 }
 
-type ConfigYAMLObj struct {
-	Result DoneApplyState
+type waitRuleContractV1Result struct {
+	Result waitRuleContractV1ResultDetails
 }
 
-type DoneApplyState struct {
-	Done       bool
-	Successful bool
-	Message    string
+type waitRuleContractV1ResultDetails struct {
+	Done       bool   `json:"done"`
+	Successful bool   `json:"successful"`
+	Message    string `json:"message"`
 }
 
-func (t WaitRuleContractV1) ApplyYttWaitRule(res ctlres.Resource) (*DoneApplyState, error) {
+func (t WaitRuleContractV1) ApplyYttWaitRule(res ctlres.Resource) (*waitRuleContractV1ResultDetails, error) {
 	if !t.ResourceMatcher.Matches(res) {
 		return nil, nil
 	}
@@ -36,7 +36,7 @@ func (t WaitRuleContractV1) ApplyYttWaitRule(res ctlres.Resource) (*DoneApplySta
 	return t.evalYtt(res)
 }
 
-func (t WaitRuleContractV1) evalYtt(res ctlres.Resource) (*DoneApplyState, error) {
+func (t WaitRuleContractV1) evalYtt(res ctlres.Resource) (*waitRuleContractV1ResultDetails, error) {
 	opts := cmdtpl.NewOptions()
 
 	opts.DataValuesFlags.FromFiles = []string{"values.yml"}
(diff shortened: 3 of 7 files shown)


pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -41,6 +41,16 @@ type customWaitingResourceCondition struct {
 }
 
 func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
+	if mod := s.waitRule.AsMods(); mod != nil {
+		configObj, err := mod.ApplyYttWaitRule(s.resource)
+		if err != nil {
+			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
+				"Error: Applying ytt wait rule: %s", err)}
```
rohitagg2020: It should be err.Error() instead of err

sethiyash: oh Thanks. Adopted.

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+          def check_status():
+              if state == "Failed":
+                return {"Done":True, "Successful": False, "Message": ""}
+              elif state == "Running":
+                return {"Done":True, "Successful": True, "Message": ""}
+              else:
+                return {"Done":False, "Successful": False, "Message": "Not in Failed or Running state"}
+              end
+          end
```
cppforlife: @jtigger any opinions on this interface presented via kapp (e.g. should return value be something else than hash, etc.)?

pkg/kapp/config/config.go
```diff
@@ -53,6 +54,18 @@ type WaitRuleConditionMatcher struct {
 	Success bool
 }
 
+type WaitRuleYtt struct {
+	// Contracts are named and versioned (eg v1)
+	// to provide a stable interface to rule authors.
+	// Multiple contracts will be offered at the same time
+	// so that existing rules do not not break as we decide to evolve running environment.
+	WaitRuleContractV1 *WaitRuleYttContractV1 `json:"waitRuleContractV1"`
```
cppforlife: waitRule in the name is redundant (since rules are under waitRules). in following example [0], ive used overlayContractV1 because main logic is done via ytt "overlay". should we call this something like funcContractV1 since main logic is done via a function?

[0] https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/config/default.go#L125-L126

pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -41,6 +41,16 @@ type customWaitingResourceCondition struct {
 }
 
 func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
+	if mod := s.waitRule.AsMods(); mod != nil {
```
cppforlife: i think this is done too early. we should still be abiding by supportsObservedGeneration field check.

pkg/kapp/yttresmod/wait_rule_contract_v1_mod.go
```diff
@@ -0,0 +1,90 @@
...
+	Result DoneApplyState
+}
+
+type DoneApplyState struct {
+	Done       bool
+	Successful bool
+	Message    string
+}
+
+func (t WaitRuleContractV1Mod) ApplyYttWaitRule(res ctlres.Resource) (*ConfigYAMLObj, error) {
```
cppforlife: any reason why this should not return DoneApplyState?

pkg/kapp/yttresmod/wait_rule_contract_v1_mod.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return &configObj, nil
+}
+
+func (t WaitRuleContractV1Mod) valuesYAML(res ctlres.Resource) ([]byte, error) {
+	return yaml.Marshal(res.DeepCopyRaw())
+}
+
+func getConfigYAML() []byte {
```
cppforlife: should be private method

test/e2e/custom_wait_rules_test.go
pivotaljohn: _(I'm mindful that this feedback is "rather late to the party" ... my hope is that it's helpful...)_

## Hook in the form of a `ytt` Library for Future Value?

I'm interested in the possibilities kept open by the content of `waitRuleContractV1` being a `ytt` Library (i.e. collection of files that are executed together).

_However_, doing so brings some ceremony that the user can't value: the user is naming _a lot_ of objects (a file, a function, unpacking the `state` variable) which they have no other need to do other than satisfy this interface.

~Have we considered a simple interface where the value of the key is the body of a function?~ _(on further thought, I end-up retracting this suggestion — noted [below](https://github.com/vmware-tanzu/carvel-kapp/pull/475#discussion_r846410278).)_


## Return value of `check_status()` through a function?

Given that the desired output is _not_ arbitrary, but very specific in shape (i.e. is equally part of the contract), using an open-ended data structure like a hash invites the possibility of errors without much upside.

Would it be worth it to provide a function to format the result?

```python
load("@kapp:status", "status")
state = status.currentState
def check_status():
  if state == "Failed":
    return status.Result(done=True, successful=False, message="")
  elif state == "Running":
    return status.Result(done=True, successful=True, message="")
  else:
    return status.Result(done=False, successful=False, message="Not in Failed or Running state")
  end
end
```

## Separate code declaration from reference?

If I'm reading this right, the ytt Library declaration _is_ also the reference.
So, if I want to declare multiple custom wait rules, I need to repeat that library structure.

I am imagining that the user might want to provide multiple custom wait rules. Would it help to extract the library into a separate value?

```yaml
waitRuleYttLibrary:
  rules.star: |
    ...
    def crontab_status():
    ...
waitRules:
- resourceMatchers:
  - apiVersionKindMatcher: {apiVersion: corp.com/v1, kind: DatabaseInstance}
  conditionMatchers:
  - type: Failed
    status: "True"
    failure: true
  - type: Deployed
    status: "True"
    success: true
  supportsObservedGeneration: true
- resourceMatchers:
   - apiVersionKindMatcher: {apiVersion: stable.example.com/v1, kind: CronTab}
   customStateMatcher: "crontab_status"
```

sethiyash: Hi @jtigger

Thanks for this much detailed feedback  💯 . _(You are definitely not late to the party yet 😄 )_

> However, doing so brings some ceremony that the user can't value: the user is naming a lot of objects (a file, a function, unpacking the state variable) which they have no other need to do other than satisfy this interface.

Totally Agree with this part. 

> `load("@kapp:status", "status")`

Not clear about this how we can import `status` from `@kapp:status`  and what is `@kapp:status` here ?

sethiyash: > - resourceMatchers:
     - apiVersionKindMatcher: {apiVersion: stable.example.com/v1, kind: CronTab}
     customStateMatcher: "crontab_status"
   
   Here `crontab_status` is in quotes by this way `crontab_status` function will get called and return the result ?

pivotaljohn: > Not clear about this how we can import `status` from `@kapp:status` and what is `@kapp:status` here ?

Yeah, apologies, I'm working off of some ideas that haven't yet been broadly considered.

Today, `ytt` comes with what you could think of as its "standard library". We call it `@ytt`.

A Library (in this sense) is a collection of "modules" (that's a Starlark concept: https://github.com/google/starlark-go/blob/master/doc/spec.md#module-execution). Examples are: `@ytt:data` is the "data" module, `@ytt:template` is the "template" module.

We don't have the full facility to do this in `ytt`, yet, but... I was imaging that we might collect one or more modules that are specifically in support of Starlark as it appears in `kapp`.
So as an aspirational sketch, I'm saying, "what if ytt allowed you to register your own library as well... and further, what if `kapp` _did_ register its own library... and we called it `@kapp`?"

`@kapp:status` could be the first of such modules in the `@kapp` ytt library.
This module could have utilities for presenting and processing resource status.

Now, under the covers, it may well be that the means by which we plumb the status of resource(s) _is_ as data values. And that makes a ton of sense to do so.
But I'm suggesting that in this particular context, that might be an implementation detail that the user neither cares about nor needs to understand. And would be the implementation _within_ this `@kapp:status` module.

Coincidentally, that module also can provide functions like `status.Result()` that helps catch/prevent syntactic errors and overall provide a better experience.

pivotaljohn: I previously suggested:
> Have we considered a simple interface where the value of the key is the body of a function?

But on second thought, I'm seeing that limiting the "viewport" of this field to being only the inside of a function would cut-off opportunities to gracefully handle more complex cases.
Namely, ensuring the user continues to have the ability to load `@ytt:` modules will be valuable: there's facilities for encoding/decoding, parsing, and manipulating values.

Also, I could imagine that as adoption of a feature like this increases, it's not unreasonable to expect a need to easily share/reuse such rules (e.g. `vendir sync'ed` and referenced via a path rather than inline) would emerge.
So, making the unit of `ytt` code inclusion being a proper ytt library is making more sense, to me.

pivotaljohn: BTW, I'm realizing that I've been using a term that's rather overloaded: "library".

In `ytt` there are three meanings of this word (sigh):
- the set of files/directories that you `-f` into `ytt` is known as a "Library".
  - when I'm musing about what shape should the user's input be: a function body? a module? a library? — I'm using _this_ definition of library, here.
  - note that we tend to only talk about so-called "private libraries", like here: [What is a library?](https://carvel.dev/ytt/docs/v0.40.0/lang-ref-ytt-library/#what-is-a-library); but note the mention of the "root library" in that description.
- a collection of loadable modules is _also_ known as a library: `@ytt:` is the primary (and currently, only) example here (it's this: [Built-in ytt library](https://carvel.dev/ytt/docs/v0.40.0/lang-ref-ytt/#docs))
  - when talking about this thing `@kapp:` _this_ is the definition of library I'm using there.
- when a file has the `.lib.yaml` filename extension, we colloquially refer to that as a "library file" (mostly because of that `.lib.` extension). It's more accurate to call it a "loadable module". (as described in [Load Statements](https://carvel.dev/ytt/docs/v0.40.0/lang-ref-load/))
  - this definition hasn't come up in this convo; just adding it for completeness.
 
Hopefully, this helps! 😬

pkg/kapp/yttresmod/wait_rule_contract_v1_mod.go
cppforlife: private (to package) function:
```
func getConfigYAML() []byte {
```

this would be private (on struct) method:
```
func (t WaitRuleContractV1Mod) getConfigYAML() []byte {
```

being private to struct implies that it is not meant to be used elsewhere in the package.

sethiyash: Adopted.

pkg/kapp/yttresmod/wait_rule_contract_v1_mod.go
sethiyash: Adopted.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
sethiyash: moved it below after supportsObservedGeneration field check.

pkg/kapp/config/config.go
sethiyash: Adopted. Changed it to `funcContractV1 `

pkg/kapp/config/config.go
```diff
@@ -53,6 +54,18 @@ type WaitRuleConditionMatcher struct {
...
 
+type WaitRuleYtt struct {
+	// Contracts are named and versioned (eg v1)
+	// to provide a stable interface to rule authors.
+	// Multiple contracts will be offered at the same time
+	// so that existing rules do not not break as we decide to evolve running environment.
+	FuncContractV1 *funcContractV1 `json:"funcContractV1"`
+}
+
+type funcContractV1 struct {
```
cppforlife: this should not be private.

cppforlife: this should not be private.

pkg/kapp/config/config.go
```diff
@@ -211,6 +224,21 @@ func (r RebaseRule) Validate() error {
 	return nil
 }
 
+func (r WaitRule) GetContract() *yttresmod.WaitRuleContractV1 {
```
cppforlife: i dont think we need this method at all. see comment below.

cppforlife: i dont think we need this method at all. see comment below.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -54,6 +54,16 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			"Waiting for generation %d to be observed", obj.Metadata.Generation)}
 	}
 
+	if contract := s.waitRule.GetContract(); contract != nil {
```
cppforlife:
```
if s.waitRule.Ytt != nil {
   ...
}
```

cppforlife: s.waitRule.Ytt != nil {

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestCustomWaitRules(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	//kubectl := Kubectl{t, env.Namespace, logger}
```
cppforlife: nit: remove dead code

cppforlife: nit: remove dead code

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,109 @@
...
+              properties:
+                currentState:
+                  type: string
+  scope: Namespaced
+  names:
+    plural: crontabs
+    singular: crontab
+    kind: CronTab
+    shortNames:
+      - ct
```
cppforlife: unnecessary code

cppforlife: unnecessary code

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,109 @@
...
+apiVersion: "stable.example.com/v1"
+kind: CronTab
+metadata:
+  name: my-new-cron-object-1
+spec:
+  cronSpec: "* * * * */5"
+  image: my-awesome-cron-image
+status:
+  currentState: %s
+---
```
cppforlife, line 88: nit: remove ---

cppforlife, line 88: nit: remove ---

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,109 @@
...
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy when resource current state as running)", func() {
+		res, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name},
+			RunOpts{IntoNs: true, StdinReader: strings.NewReader(crdYaml + fmt.Sprintf(crYaml, "Running") + config)})
```
cppforlife: no need for `IntoNs: true`

> fmt.Sprintf(crYaml, "Running")

can we just inline this?

cppforlife: no need for `IntoNs: true`

> fmt.Sprintf(crYaml, "Running")

can we just inline this?

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			require.Errorf(t, err, "")
+		}
+		require.Contains(t, res, "Succeeded")
```
cppforlife: what is this checking?

cppforlife: what is this checking?

pkg/kapp/yttresmod/wait_rule_contract_v1.go
```diff
@@ -0,0 +1,91 @@
...
+
+	err := yaml.Unmarshal(file.Bytes(), &configObj)
+	if err != nil {
+		return nil, fmt.Errorf("Deserializing result: %s", err)
+	}
+
+	return &configObj.Result, nil
+}
+
+func (t WaitRuleContractV1) valuesYAML(res ctlres.Resource) ([]byte, error) {
```
cppforlife: nit: this function can be inlined

cppforlife: nit: this function can be inlined

pkg/kapp/config/config.go
```diff
@@ -53,6 +54,18 @@ type WaitRuleConditionMatcher struct {
...
+type WaitRuleYtt struct {
+	// Contracts are named and versioned (eg v1)
+	// to provide a stable interface to rule authors.
+	// Multiple contracts will be offered at the same time
+	// so that existing rules do not not break as we decide to evolve running environment.
+	FuncContractV1 *funcContractV1 `json:"funcContractV1"`
+}
+
+type funcContractV1 struct {
+	RulesStar string `json:"rules.star"`
```
cppforlife: im thinking we should rename this field to `resource.star` and function should be called `is_done(resource)`

cppforlife: im thinking we should rename this field to `resource.star` and function should be called `is_done(resource)`

pkg/kapp/config/config.go
sethiyash: Adopted.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
sethiyash: Adopted.

test/e2e/custom_wait_rules_test.go
sethiyash: Started using it to check for resource present in cluster.

test/e2e/custom_wait_rules_test.go
sethiyash: removed.

test/e2e/custom_wait_rules_test.go
sethiyash: As while using it in RunOpts I'm giving it in this order `crdYaml + fmt.Sprintf(crYaml, "Failed") + config` so this is added to separate `Config` kind resource from CR resource.

sethiyash: As while using it in RunOpts I'm giving it in this order `crdYaml + fmt.Sprintf(crYaml, "Failed") + config` so this is added to separate `Config` kind resource from CR resource.

test/e2e/custom_wait_rules_test.go
sethiyash: Adopted.

test/e2e/custom_wait_rules_test.go
sethiyash: Bad idea to check for the resource existence started doing it with `NewPresentClusterResource("CronTab", "my-new-cron-object-1", env.Namespace, kubectl)`

pkg/kapp/yttresmod/wait_rule_contract_v1.go
sethiyash: Adopted.

pkg/kapp/config/config.go
sethiyash: Adopted.

sethiyash: Adopted.

test/e2e/custom_wait_rules_test.go
sethiyash: Bad idea to check for the resource existence started doing it with `NewPresentClusterResource("CronTab", "my-new-cron-object-1", env.Namespace, kubectl)`

pkg/kapp/config/config.go
sethiyash: Removed the method.

sethiyash: Removed.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -55,6 +56,18 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			"Waiting for generation %d to be observed", obj.Metadata.Generation)}
 	}
 
+	if s.waitRule.Ytt != nil {
+		configObj, err := yttresmod.WaitRuleContractV1{ResourceMatcher: ctlres.AnyMatcher{
```
cppforlife: nit: newline after `{`, and before closing `}`

pkg/kapp/yttresmod/wait_rule_contract_v1.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type ConfigYAMLObj struct {
+	Result DoneApplyState
+}
+
+type DoneApplyState struct {
+	Done       bool
+	Successful bool
+	Message    string
```
cppforlife: use `json:"done"` (lowercase naming)

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,115 @@
...
+	name := "test-custom-wait-rule-contract-v1"
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy resource with current state as running", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{StdinReader: strings.NewReader(crdYaml + fmt.Sprintf(crYaml, "Running") + config)})
```
cppforlife: nit/style: newline after `{`

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			require.Errorf(t, err, "Expected CronTab to be deployed")
+		}
+		NewPresentClusterResource("CronTab", "my-new-cron-object-1", env.Namespace, kubectl)
+	})
+
+	cleanUp()
+
+	logger.Section("deploy resource with current state as failed", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{StdinReader: strings.NewReader(crdYaml +
```
cppforlife: nit/style: newline after `{`

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,115 @@
...
+kind: Config
+
+waitRules:
+  - ytt:
+      funcContractV1:
+        resource.star: |
+          def is_done(resource):
+              state = resource.status.currentState
+              if state == "Failed":
+                return {"Done":True, "Successful": False, "Message": ""}
```
cppforlife: should use `done` (lowercase lettering)

cppforlife: nit/style: space after `:`

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,115 @@
...
+		NewPresentClusterResource("CronTab", "my-new-cron-object-1", env.Namespace, kubectl)
+	})
+
+	cleanUp()
+
+	logger.Section("deploy resource with current state as failed", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{StdinReader: strings.NewReader(crdYaml +
+			fmt.Sprintf(crYaml, "Failed") + config), AllowError: true})
+
+		require.Contains(t, err.Error(), "kapp: Error: waiting on reconcile crontab/my-new-cron-object-1")
```
cppforlife, line 119: lets also assert that the message is returned as well.

pkg/kapp/yttresmod/wait_rule_contract_v1.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2020 VMware, Inc.
```
cppforlife: i think this file should be in resourcesmisc package since it's all about waiting.

pkg/kapp/yttresmod/wait_rule_contract_v1.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/k14s/ytt/pkg/cmd/ui"
+	"github.com/k14s/ytt/pkg/files"
+)
+
+type WaitRuleContractV1 struct {
+	ResourceMatcher ctlres.ResourceMatcher
+	Starlark        string
+}
+
+type ConfigYAMLObj struct {
```
cppforlife: ConfigYAMLObj => waitRuleContractV1Result

pkg/kapp/yttresmod/wait_rule_contract_v1.go
```diff
@@ -0,0 +1,91 @@
...
+	Result DoneApplyState
+}
+
+type DoneApplyState struct {
```
cppforlife: DoneApplyState => waitRuleContractV1ResultDetails

pkg/kapp/config/config.go
sethiyash: adopted.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
sethiyash: Adopted.

pkg/kapp/yttresmod/wait_rule_contract_v1.go
sethiyash: Adopted.

test/e2e/custom_wait_rules_test.go
sethiyash: Adopted.

test/e2e/custom_wait_rules_test.go
sethiyash: adopted.

pkg/kapp/yttresmod/wait_rule_contract_v1.go
sethiyash: moved.

pkg/kapp/yttresmod/wait_rule_contract_v1.go
sethiyash: adopted.

test/e2e/custom_wait_rules_test.go
sethiyash: started checking for not empty.

pkg/kapp/yttresmod/wait_rule_contract_v1.go
sethiyash: adopted.

test/e2e/custom_wait_rules_test.go
sethiyash: adopted.

pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
```diff
@@ -0,0 +1,91 @@
...
+	Result waitRuleContractV1ResultDetails
+}
+
+type waitRuleContractV1ResultDetails struct {
+	Done       bool   `json:"done"`
+	Successful bool   `json:"successful"`
+	Message    string `json:"message"`
+}
+
+func (t WaitRuleContractV1) applyYttWaitRule(res ctlres.Resource) (*waitRuleContractV1ResultDetails, error) {
```
cppforlife: applyYttWaitRule => Apply. should be public since we expect other classes to use it.

pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
```diff
@@ -0,0 +1,91 @@
...
+
+	err := yaml.Unmarshal(file.Bytes(), &configObj)
+	if err != nil {
+		return nil, fmt.Errorf("Deserializing result: %s", err)
+	}
+
+	return &configObj.Result, nil
+}
+
+func (t WaitRuleContractV1) valuesYAML(res ctlres.Resource) ([]byte, error) {
```
cppforlife: nit/style: recommend inlining to where its used, since it's such a small method.

test/e2e/custom_wait_rules_test.go
cppforlife: i did not mean asserting that err is non nil. what i meant is we should assert that message returned from wait rule `Not in Failed or Running state` should show up in the output.

pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
sethiyash: Did that earlier but just noticed that during `./hack/build.sh` it again makes it multiline. Did this as of now but whenever we run build script it will make it multiline and show it as diff.

pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
sethiyash: Making it public also means that the return type should also be public otherwise lint check is getting failed. 

Made `WaitRuleContractV1ResultDetails` public.

pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
sethiyash: I believe it will be better to remain it in multiline otherwise every time we build it will show it as diff. Open for thoughts and suggestions.

test/e2e/custom_wait_rules_test.go
sethiyash: Got it. Added a new scenario with `currentState` as "Invalid" to get the message and validate it.

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,129 @@
...
+	cleanUp()
+
+	logger.Section("deploy resource with current state as invalid", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{
+			StdinReader: strings.NewReader(crdYaml + fmt.Sprintf(crYaml, "Invalid") + config),
+			AllowError:  true,
+		})
+
+		require.Contains(t, err.Error(), "Not in Failed or Running state")
+	})
```
cppforlife: instead of a separate section why dont we change wait rule to include message when it's failure? actually we should probably do same for success and check on the message as well.

sethiyash: Done, Made the changes to check for message also both in case of success and failure.

test/e2e/custom_wait_rules_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	config := `
+apiVersion: kapp.k14s.io/v1alpha1
+kind: Config
+
+waitRules:
+  - ytt:
+      funcContractV1:
+        resource.star: |
+          def is_done(resource):
```
jdef, line 28: is returning nothing the same (effectively) as returning `{}`? meaning done=false, successful=false, message='' ?

pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
praveenrewar: > inlining to where its used

@sethiyash I am not 100%sure, but I think what @cppforlife meant was to directly use `yaml.Marshal()` instead of `valuesYAML()` as that's the only thing it is doing.

test/e2e/custom_wait_rules_test.go
sethiyash: Hi @jdef 
Yes, Here returning `{}` is equivalent to returning `{"done": False, "successful": False, "message": ""}` in such case it will keep on waiting until our default wait timeout of 15 minutes is not reached.

pkg/kapp/resourcesmisc/wait_rule_contract_v1.go
cppforlife: > I believe it will be better to remain it in multiline otherwise every time we build it will show it as diff. Open for thoughts and suggestions.

i dont understand this comment @sethiyash

sethiyash: oh sorry, I thought you are saying to make this `valuesYAML()` method inline. Adopted the change to remove this method and to marshal it there directly. 

Thanks, @praveenrewar for the clarification.

PR #478, opened 2022-04-09
Support for matching observedGeneration for status.Condition level

### LINKED ISSUE #435: Support for condition-level `observedGeneration` wait rule

Hi - I'm trying to configure a resource wait rule including `observedGeneration` from the condition-level, but I don't think it's possible. I'd be interested if there is an alternative, but seems like this might be a feature request from looking into it more.

For a specific scenario, I'm using [cert-manager](https://cert-manager.io) and have a custom wait rule for `Certificate`. The wait rule works fine for new resources, but if an existing certificate gets updated (e.g. its `dnsNames` changes) it does not wait since `Ready` is already present and true. I tried the [`supportsObservedGeneration` flag](https://carvel.dev/kapp/docs/v0.45.0/config/#wait-rules), but it didn't work (i.e. ongoing "waiting for generation * to be observed"). Digging into the code I found kapp [only checks `status.observedGeneration`](https://github.com/vmware-tanzu/carvel-kapp/blob/e0bf1c9d485c5c59a23b4cfe75915782e7fb4b42/pkg/kapp/resourcesmisc/custom_waiting_resource.go#L52-L55), not `status.conditions[*].observedGeneration` (which is what [cert-manager provides](https://github.com/cert-manager/cert-manager/blob/b1180c59ad588e73ac25b0d70a86661cf7c180e1/pkg/api/util/conditions.go#L176), [not status-level](https://github.com/cert-manager/cert-manager/blob/ad4264b6ecd48702b4a3b83b06bd00bd32a19df5/deploy/crds/crd-certificates.yaml#L297-L359)).

It looks like [KEP-1623: Standardize Conditions](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1623-standardize-conditions#proposal) formalized `observedGeneration` as a [condition-level field](https://github.com/kubernetes/apimachinery/blob/14143357d417e6d4c3c3669265d13f41d6eb8adb/pkg/apis/meta/v1/types.go#L1483-L1488), so it seems like cert-manager is following standards, at least. That KEP was released in v1.19 (August 2020) which was after the original [kapp pull request](https://github.com/vmware-tanzu/carvel-kapp/pull/115) that introduced these custom wait rules, so I'd guess the condition-level behavior just hasn't been previously considered here. It also just seems much less common as I haven't found many other projects including the condition field in their schema ([contour does](https://github.com/vmware-tanzu/community-edition/blob/3f967483a08016aecba2901f7c4a4152387474a7/addons/packages/contour/1.18.2/bundle/config/upstream/contour.yaml#L477)), but I do see that `kubectl wait` officially respects it when present according to [tests](https://github.com/kubernetes/kubernetes/blob/4033355d292505eaed05655bd91f4ea9edc7ef27/staging/src/k8s.io/kubectl/pkg/cmd/wait/wait_test.go#L926-L1000).

<details>
<summary><code>kapp/Config</code> debugging</summary>

For reference, I was using the following config when trying to get this to work.

```yaml
kind: Config
waitRules:
- resourceMatchers:
  - apiVersionKindMatcher:
      apiVersion: cert-manager.io/v1
      kind: Certificate
  # hangs; needs to look at .status.conditions[*].observedGeneration instead
  # supportsObservedGeneration: true
  conditionMatchers:
  - type: Ready
    status: "True"
    success: true
```

</details>

Assuming I haven't misunderstood something, I think kapp should add support for `supportsObservedGeneration` at the condition matcher-level. I wouldn't expect operators to typically need it for both the wait rule and condition matcher, but it would technically be possible.

For the example scenario, I'd then use the following.

```yaml
kind: Config
waitRules:
- resourceMatchers:
  - apiVersionKindMatcher:
      apiVersion: cert-manager.io/v1
      kind: Certificate
  conditionMatchers:
  - type: Ready
    status: "True"
    success: true
    supportsObservedGeneration: true
```

Let me know if there are alternatives or this sounds reasonable - I could probably provide a small PR with the change.

Thanks!

---
Vote on this request

kumaritanushree:
Fixes: https://github.com/vmware-tanzu/carvel-kapp/issues/435
Solution: support for matching observedGeneration for status.Condition level

TODO

- [x] Manual testing
- [x] Should check for observedGeneration for each status.Condition if  SupportsObservedGeneration=true in waitRule.ConditionMatchers. (tested for cert manager as this condition is not applicable for some specific type resource)

**Reference(tested with CR cert-manager's certificate https://cert-manager.io/docs/usage/certificate/)**
_I have used kapp config while deploying certificate :_
```
apiVersion:` kapp.k14s.io/v1alpha1
kind: Config
waitRules:
- resourceMatchers:
  - apiVersionKindMatcher:
      apiVersion: cert-manager.io/v1
      kind: Certificate
  conditionMatchers:
  - type: Ready
    status: "False"
    success: true
    supportsObservedGeneration: true
  ```
    
In this scenario what kapp will do is, kapp will check if conditionMatchers.supportsObservedGeneration = true then it will match Metadata.Generation with Status.Condition.ObservedGeneration , also check for type =ready and status=false then kapp will say the resource got deployed successfully.

### DIFF
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -55,13 +55,15 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			"Waiting for generation %d to be observed", obj.Metadata.Generation)}
 	}
 
+	// True if a condition has not observed the latest generation
+	condObsGenMatchWait := false
 	// Check on failure conditions first
 	for _, condMatcher := range s.waitRule.ConditionMatchers {
 		for _, cond := range obj.Status.Conditions {
 			if cond.Type == condMatcher.Type && cond.Status == condMatcher.Status {
 				if condMatcher.SupportsObservedGeneration && obj.Metadata.Generation != cond.ObservedGeneration {
-					return DoneApplyState{Done: false, Message: fmt.Sprintf(
-						"Waiting for generation %d to be observed at status condition %s level", obj.Metadata.Generation, cond.Type)}
+					condObsGenMatchWait = true
+					continue
 				}
 				if condMatcher.Failure {
 					return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
@@ -77,8 +79,8 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 		for _, cond := range obj.Status.Conditions {
 			if cond.Type == condMatcher.Type && cond.Status == condMatcher.Status {
 				if condMatcher.SupportsObservedGeneration && obj.Metadata.Generation != cond.ObservedGeneration {
-					return DoneApplyState{Done: false, Message: fmt.Sprintf(
-						"Waiting for generation %d to be observed at status condition %s level", obj.Metadata.Generation, cond.Type)}
+					condObsGenMatchWait = true
+					continue
 				}
 				if condMatcher.Success {
 					return DoneApplyState{Done: true, Successful: true, Message: fmt.Sprintf(
@@ -89,5 +91,9 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 		}
 	}
 
+	if condObsGenMatchWait {
+		return DoneApplyState{Done: false, Message: fmt.Sprintf(
+			"Waiting for generation %d to be observed by status condition(s)", obj.Metadata.Generation)}
+	}
 	return DoneApplyState{Done: false, Message: "No failing or successful conditions found"}
 }
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -56,13 +56,13 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 	}
 
 	// True if a condition has not observed the latest generation
-	condObsGenMatchWait := false
+	hasConditionWaitingForGeneration := false
 	// Check on failure conditions first
 	for _, condMatcher := range s.waitRule.ConditionMatchers {
 		for _, cond := range obj.Status.Conditions {
 			if cond.Type == condMatcher.Type && cond.Status == condMatcher.Status {
 				if condMatcher.SupportsObservedGeneration && obj.Metadata.Generation != cond.ObservedGeneration {
-					condObsGenMatchWait = true
+					hasConditionWaitingForGeneration = true
 					continue
 				}
 				if condMatcher.Failure {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/kapp/resourcesmisc/custom_waiting_resource.go [hunk 1]
```diff
@@ -58,6 +59,10 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 	for _, condMatcher := range s.waitRule.ConditionMatchers {
 		for _, cond := range obj.Status.Conditions {
 			if cond.Type == condMatcher.Type && cond.Status == condMatcher.Status {
+				if condMatcher.SupportsObservedGeneration && obj.Metadata.Generation != cond.ObservedGeneration {
```
cppforlife, line 63: multiple conditions might be in play. some of them might be reporting old health (ie have old observedGeneration). i think we would want to ignore conditions that we "support observed generation for", but also it does not have current generation. if we just return then we would exit too quick (especially if there are other conditions that may be considered updated).

kumaritanushree: @cppforlife  updated the logic to check each condition instead of returning early.

rohitagg2020:
LGTM

pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -54,10 +55,16 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			"Waiting for generation %d to be observed", obj.Metadata.Generation)}
 	}
 
+	// True if a condition has not observed the latest generation
+	condObsGenMatchWait := false
```
praveenrewar:
```suggestion
	hasConditionWaitingForGeneration := false
```

pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -80,5 +91,9 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 		}
 	}
 
+	if condObsGenMatchWait {
+		return DoneApplyState{Done: false, Message: fmt.Sprintf(
+			"Waiting for generation %d to be observed by status condition(s)", obj.Metadata.Generation)}
```
praveenrewar, line 95: Just a suggestion, would love to know what others think.
```suggestion
			"Condition(s) waiting for generation %d to be observed", obj.Metadata.Generation)}
```

100mik (approved):
LGTM!

pkg/kapp/resourcesmisc/custom_waiting_resource.go
kumaritanushree: Updated the changes

pkg/kapp/resourcesmisc/custom_waiting_resource.go
kumaritanushree: As we all agreed that we want to go with older change only. Closing this without change.

praveenrewar (approved):
LGTM

PR #481, opened 2022-04-12
Fixing the issue when we wait on completed init container forever

rohitagg2020:
Fixing the issue when we wait on completed container forever.

### DIFF
--- a/pkg/kapp/logs/pod_container_log.go
+++ b/pkg/kapp/logs/pod_container_log.go
@@ -6,7 +6,6 @@ package logs
 import (
 	"bufio"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


pkg/kapp/logs/pod_container_log.go
```diff
@@ -144,10 +145,13 @@ func (l PodContainerLog) obtainStream(ui ui.UI, linePrefix string, cancelCh chan
 
 		if !isWaitingMsgPrintedOnce {
 			ui.BeginLinef("%s# waiting for '%s' logs to become available...\n", linePrefix, l.tag)
+			isWaitingMsgPrintedOnce = true
 			if !l.opts.Follow {
+				if err == nil {
+					err = errors.New(fmt.Sprintf("Container %s not in Ready state", l.container))
```
100mik: We stick to using `fmt.Errorf` I guess
(I think the linter is pointing out the same)

rohitagg2020: Adopted

PR #482, opened 2022-04-12
Migrated apps to have suffix when FQ env is unset

praveenrewar:
- Migrated app names to have suffix when KAPP_FQ_CONFIGMAP_NAMES env variable is unset, just like old versions of kapp would have
- Add constructor for recorded app
- Remove suffix when migration is enabled (this prevents double migration when someone is using fully qualified name even when migration is enabled)

`Current behaviour`
```bash
export KAPP_FQ_CONFIGMAP_NAMES=true

kapp deploy -a my-app -f app.yml --yes

$ kapp ls
Namespace  Name        Namespaces  Lcs   Lca
default    my-app      default     true  7s

$ KAPP_FQ_CONFIGMAP_NAMES=false kapp ls
Namespace  Name        Namespaces  Lcs   Lca
default    my-app      default     true  7s

KAPP_FQ_CONFIGMAP_NAMES=false inspect my-app
kapp: Error: App 'my-app' (namespace: default) does not exist: configmaps "my-app" not found
```

`Expected behaviour from this change`
```bash
export KAPP_FQ_CONFIGMAP_NAMES=true

kapp deploy -a my-app -f app.yml --yes

$ kapp ls
Namespace  Name        Namespaces  Lcs   Lca
default    my-app      default     true  7s

$ KAPP_FQ_CONFIGMAP_NAMES=false kapp ls
Namespace  Name                     Namespaces  Lcs   Lca
default    my-app.apps.k14s.io      default     true  7s
```

Doc [PR](https://github.com/vmware-tanzu/carvel/pull/414)

### DIFF
--- a/pkg/kapp/app/apps.go
+++ b/pkg/kapp/app/apps.go
@@ -54,7 +54,7 @@ func (a Apps) Find(name string) (App, error) {
 	}
 
 	return NewRecordedApp(name, a.nsName, time.Time{}, a.coreClient,
-		a.identifiedResources, a.appInDiffNsHintMsg, a.logger, nil), nil
+		a.identifiedResources, a.appInDiffNsHintMsg, a.logger), nil
 }
 
 func (a Apps) List(additionalLabels map[string]string) ([]App, error) {
@@ -82,8 +82,12 @@ func (a Apps) list(additionalLabels map[string]string, nsName string) ([]App, er
 	}
 
 	for _, app := range apps.Items {
-		result = append(result, NewRecordedApp(app.Name, app.Namespace, app.ObjectMeta.CreationTimestamp.Time, a.coreClient,
-			a.identifiedResources, a.appInDiffNsHintMsg, a.logger, &app))
+		recordedApp := NewRecordedApp(app.Name, app.Namespace, app.ObjectMeta.CreationTimestamp.Time, a.coreClient,
+			a.identifiedResources, a.appInDiffNsHintMsg, a.logger)
+
+		recordedApp.setMeta(app)
+
+		result = append(result, recordedApp)
 	}
 
 	return result, nil
--- a/pkg/kapp/app/apps.go
+++ b/pkg/kapp/app/apps.go
@@ -6,6 +6,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
@@ -89,7 +90,7 @@ func (a Apps) list(additionalLabels map[string]string, nsName string) ([]App, er
 		name := app.Name
 		isMigrated := false
 
-		if _, ok := app.Annotations[KappIsConfigmapMigratedAnnotationKey]; ok {
+		if _, ok := app.Annotations[KappIsConfigmapMigratedAnnotationKey]; ok && strings.ToLower(os.Getenv("KAPP_FQ_CONFIGMAP_NAMES")) == "true" {
 			name = strings.TrimSuffix(app.Name, AppSuffix)
 			isMigrated = true
 		}
--- a/pkg/kapp/app/apps.go
+++ b/pkg/kapp/app/apps.go
@@ -6,7 +6,6 @@ package app
 import (
 	"context"
 	"fmt"
-	"os"
 	"strings"
 	"time"
 
@@ -18,11 +17,8 @@ import (
 )
 
 const (
-	KappIsAppLabelKey                      = "kapp.k14s.io/is-app"
-	kappIsAppLabelValue                    = ""
-	KappIsConfigmapMigratedAnnotationKey   = "kapp.k14s.io/is-configmap-migrated"
-	KappIsConfigmapMigratedAnnotationValue = ""
-	AppSuffix                              = ".apps.k14s.io"
+	KappIsAppLabelKey   = "kapp.k14s.io/is-app"
+	kappIsAppLabelValue = ""
 )
 
 type Apps struct {
@@ -57,9 +53,8 @@ func (a Apps) Find(name string) (App, error) {
 		return nil, fmt.Errorf("Expected non-empty namespace")
 	}
 
-	return &RecordedApp{name, name + AppSuffix, a.nsName, false, time.Time{}, a.coreClient,
-		a.identifiedResources, a.appInDiffNsHintMsg, nil,
-		a.logger.NewPrefixed("RecordedApp")}, nil
+	return NewRecordedApp(name, a.nsName, time.Time{}, a.coreClient,
+		a.identifiedResources, a.appInDiffNsHintMsg, a.logger, nil), nil
 }
 
 func (a Apps) List(additionalLabels map[string]string) ([]App, error) {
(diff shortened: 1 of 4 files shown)


pkg/kapp/app/apps.go
```diff
@@ -89,7 +90,7 @@ func (a Apps) list(additionalLabels map[string]string, nsName string) ([]App, er
 		name := app.Name
 		isMigrated := false
 
-		if _, ok := app.Annotations[KappIsConfigmapMigratedAnnotationKey]; ok {
+		if _, ok := app.Annotations[KappIsConfigmapMigratedAnnotationKey]; ok && strings.ToLower(os.Getenv("KAPP_FQ_CONFIGMAP_NAMES")) == "true" {
```
100mik: can we not use isMigrationEnabled() here?

praveenrewar: It's private to recorded app. Should I use a constant for `KAPP_FQ_CONFIGMAP_NAMES`?

100mik: That sounds better!

cppforlife: it's probably time to make a constructor for RecorderApp: https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/apps.go#L92-L99. seems like a log of app guts are in Apps

praveenrewar: > seems like a log of app guts are in Apps

Yep, I have been thinking the same for a while.

> it's probably time to make a constructor for RecorderApp

Makes sense.

praveenrewar: Apps looks fine now, but I feel that the Recorded app constructor is a bit complex now (with the configmap and the usage of private methods)

pkg/kapp/app/recorded_app.go
```diff
@@ -40,11 +42,33 @@ type RecordedApp struct {
...
+	if _, ok := app.Annotations[KappIsConfigmapMigratedAnnotationKey]; ok && recordedApp.isMigrationEnabled() {
+		recordedApp.name = strings.TrimSuffix(recordedApp.name, AppSuffix)
+		recordedApp.isMigrated = true
+	}
+
+	recordedApp.setMeta(*app)
+
+	return recordedApp
+}
+
```
praveenrewar: Apps looks fine now, but I feel that the Recorded app constructor is a bit complex now (with the configmap and the usage of private methods)

rohitagg2020: As this code[From L55-L60] is specific to list, how about putting it back there and in here, we can just instantiate recordedApp and return it

sethiyash: +1 on what Rohit has said to make it separate from constructor and not to make constructor complex.

praveenrewar: The idea was to move things related to recorded app from apps to recorded app.

pkg/kapp/app/recorded_app.go
```diff
@@ -40,11 +42,27 @@ type RecordedApp struct {
...
 }
 
+func NewRecordedApp(name, nsName string, creationTimestamp time.Time, coreClient kubernetes.Interface,
+	identifiedResources ctlres.IdentifiedResources, appInDiffNsHintMsgFunc func(string) string, logger logger.Logger) *RecordedApp {
+
+	recordedApp := &RecordedApp{name, nsName, false, creationTimestamp, coreClient, identifiedResources, appInDiffNsHintMsgFunc,
+		nil, logger.NewPrefixed("RecordedApp")}
+
+	if recordedApp.isMigrationEnabled() {
+		// If migration is enabled always trim suffix, even if user added it manually (to avoid double migration)
```
praveenrewar, line 52: Currently if migration is enabled and a user manually adds suffix to a migrated app then kapp treats it as a non migrated app and migrates it again. 

We could either always trim suffix when migration is enabled which still allows users to use the fully qualified name but prevents double migration. 

Or we could treat this as a new app and return an error accordingly.

I have implemented the first scenario assuming that the suffix belongs to kapp and users wouldn't want to use it and to keep the implementation a bit simple since this is not an ideal scenario.

neil-hickey: Hmm, so lemme think:

`KAPP_FQ_CONFIGMAP_NAMES=true kapp deploy -a foo -f ...` ==> config name = `foo.apps.k14s.io`

now I run: `KAPP_FQ_CONFIGMAP_NAMES=true kapp deploy -a foo.apps.k14s.io -f ...`

- It sounds like you expect that the configmap stays named `foo.apps.k14s.io` and that this is treated as the same app.
- In my mind the user is deploying a different app that they have called `foo.apps.k14s.io` which is distinct from `foo` and so we should maintain that as the app name, i.e the configmap therefore becomes `foo.apps.k14s.io.apps.k14s.io` even though it looks weird.

I guess I see this `.apps.k14s.io` as an internal name - and users should not be using it even when `kapp deploy`'ing. If they do - that should be the name that is preserved.

In this case `KAPP_FQ_CONFIGMAP_NAMES=true kapp ls` will show `foo`? or `foo.apps.k14s.io`?

test/e2e/app_suffix_test.go
```diff
@@ -100,21 +100,20 @@ func TestAppSuffix_AppExists_MigrationEnabled(t *testing.T) {
 		NewMissingClusterResource(t, "configmap", newName, env.Namespace, kubectl)
 	})
 
-	logger.Section("name already contains app suffix", func() {
-		existingName := name + app.AppSuffix
+	logger.Section("name contains app suffix for migrated app", func() {
```
neil-hickey, line 103: This test was previously capturing what I saw as desired behaviour:

1. `KAPP_FQ_CONFIGMAP_NAMES=False kapp deploy -a foo.apps.k14s.io -f ...` ==> configmap = foo.apps.k14s.io
2. `KAPP_FQ_CONFIGMAP_NAMES=True kapp deploy -a foo.apps.k14s.io -f ...` ==> configmap = foo.apps.k14s.io.apps.k14s.io

Looks like this test now checks:
1. `KAPP_FQ_CONFIGMAP_NAMES=True kapp deploy -a foo -f ...` ==> configmap = foo.apps.k14s.io
2. `KAPP_FQ_CONFIGMAP_NAMES=True kapp deploy -a foo.apps.k14s.io -f ...` ==> configmap = foo.apps.k14s.io

---

So we are somewhat exposing the internal configmap suffix name to the user now, if you use this suffix - we are making an assumption you have migrated your app. Feels odd, that was the purpose of the annotation - to remove this value judgement, and assume the user given app_name is what they want their app to be named.

pkg/kapp/app/recorded_app.go
neil-hickey: I added some comments around this new behaviour - I think this PR is now two things, one which fixes the `kapp ls` bug and another which changes `KAPP_FQ_CONFIGMAP_NAMES` behaviour. If it's easier to split the concerns into different PRs maybe that would be useful

praveenrewar: > In my mind the user is deploying a different app that they have called foo.apps.k14s.io which is distinct from foo and so we should maintain that as the app name

I started with similar thoughts initially, but then I started thinking about how users would lead to using this and I changed my approach. 
- User first enables migration and deploys an app `KAPP_FQ_CONFIGMAP_NAMES=true kapp deploy -a foo -f ...`.
- They switch back to the old behaviour and notice that the app name now is `foo.apps.k14s.io`. So they start using that name.
- They again start using the new behaviour but still use the fq name `KAPP_FQ_CONFIGMAP_NAMES=true kapp deploy -a foo.apps.k14s.io -f ...`

Now we could either treat this as a new app (fq name `foo.apps.k14s.io.apps.k14s.io`) which would lead to an ownership error or continue as the same app (by first removing the suffix) which leads to a successful deployment.

Logically, we should go with the first approach. But I was thinking it more from the user perspective and to avoid confusion I went with the latter approach, but I am okay with changing it.

praveenrewar: > If it's easier to split the concerns into different PRs maybe that would be useful

I will definitely try to make a separate PR, I just used the same PR because I wanted to use the refactoring (constructor for Recorded app) I did in this PR.

test/e2e/app_suffix_test.go
praveenrewar: Yeah, it definitely is a bit weird. I will switch to the other approach.

pkg/kapp/app/recorded_app.go
neil-hickey: > Logically, we should go with the first approach. But I was thinking it more from the user perspective and to avoid confusion I went with the latter approach, but I am okay with changing it.

I trust your judgement, if this feels like a more natural flow and a better UX I am happy with the change 👍 . This weird edge case of a user having our very specific suffix feels slightly better suited to the problem set you are describing than the one I was. The flow you described feels like it's more likely to happen than the user already using our suffix.

praveenrewar:
Closing in favour of https://github.com/vmware-tanzu/carvel-kapp/pull/490

PR #483, opened 2022-04-12
update or noop fallback: allow a rebase rule to turn an update into a noop

joe-kimmel-vmw:
Dear kapp team,
I heard you-all are big on acceptance criteria, so here's an attempt:

kapp-controller needs an "upsert or walk away and do nothing" mechanism:
- the initial existence checks are too expensive in some backends (especially ones that rhyme with "crewgle crowd crengine") so we have to be able to pass `--existing-non-labeled-resources-check=false`
- because we're not checking for existence, sometimes the create is going to be an update of an existing object (so the operation we really want to perform is an upsert, which is basically what the create-strategy: fallback annotation already provides.)
- also under _certain conditions_ we want to ensure that:
  - the upsert doesn't change the underlying resource
  - the upsert doesn't change the owner
  - i.e. we want the upsert to turn into a noop, through conditionals checked in a rebaserule. 
  
  this PR accomplishes those goals by adding a new fallback-on-update-or-noop option to the create-strategy annotation.

  I don't think the code is particularly pretty and am happy to take suggestions on how else to organize it / pass information. Also happy if anyone sees a different path for accomplishing this goal.

### DIFF
--- a/pkg/kapp/clusterapply/add_or_update_change.go
+++ b/pkg/kapp/clusterapply/add_or_update_change.go
@@ -14,10 +14,10 @@ import (
 )
 
 const (
-	createStrategyAnnKey                                                = "kapp.k14s.io/create-strategy"
-	createStrategyPlainAnnValue            ClusterChangeApplyStrategyOp = ""
-	createStrategyFallbackOnUpdateAnnValue ClusterChangeApplyStrategyOp = "fallback-on-update"
-	createStrategyFallbackOnNoopAnnValue   ClusterChangeApplyStrategyOp = "fallback-on-noop"
+	createStrategyAnnKey                                                      = "kapp.k14s.io/create-strategy"
+	createStrategyPlainAnnValue                  ClusterChangeApplyStrategyOp = ""
+	createStrategyFallbackOnUpdateAnnValue       ClusterChangeApplyStrategyOp = "fallback-on-update"
+	createStrategyFallbackOnUpdateOrNoopAnnValue ClusterChangeApplyStrategyOp = "fallback-on-update-or-noop"
 
 	updateStrategyAnnKey                                                 = "kapp.k14s.io/update-strategy"
 	updateStrategyPlainAnnValue             ClusterChangeApplyStrategyOp = ""
@@ -57,8 +57,8 @@ func (c AddOrUpdateChange) ApplyStrategy() (ApplyStrategy, error) {
 		case createStrategyFallbackOnUpdateAnnValue:
 			return AddOrFallbackOnUpdateStrategy{newRes, c}, nil
 
-		case createStrategyFallbackOnNoopAnnValue:
-			return AddOrFallbackOnNoopStrategy{newRes, c}, nil
+		case createStrategyFallbackOnUpdateOrNoopAnnValue:
+			return AddOrFallbackOnUpdateOrNoopStrategy{newRes, c}, nil
 
 		default:
 			return nil, fmt.Errorf("Unknown create strategy: %s", strategy)
@@ -164,7 +164,7 @@ func (c AddOrUpdateChange) tryToResolveUpdateConflict(
 	return fmt.Errorf(errMsgPrefix+"(tried multiple times): %s", origErr)
 }
 
-func (c AddOrUpdateChange) tryToUpdateAfterCreateConflict() error {
+func (c AddOrUpdateChange) tryToUpdateAfterCreateConflict(allowNoopUpdates bool) error {
 	var lastUpdateErr error
 
 	for i := 0; i < 10; i++ {
@@ -186,7 +186,10 @@ func (c AddOrUpdateChange) tryToUpdateAfterCreateConflict() error {
 		}
 		switch recalcChanges[0].Op() {
 		case ctldiff.ChangeOpNoop:
-			return nil
+			if allowNoopUpdates {
+				return nil
+			}
+			return fmt.Errorf("Recalculated change was a noop but those are not allowed (try adding annotation kapp.k14s.io/create-strategy: fallback-on-update-or-noop)")
 		case ctldiff.ChangeOpUpdate:
 			updatedRes, err := c.identifiedResources.Update(recalcChanges[0].NewResource())
 			if err != nil {
@@ -290,7 +293,7 @@ func (c AddOrFallbackOnUpdateStrategy) Apply() error {
 	createdRes, err := c.aou.identifiedResources.Create(c.newRes)
 	if err != nil {
 		if errors.IsAlreadyExists(err) {
-			return c.aou.tryToUpdateAfterCreateConflict()
+			return c.aou.tryToUpdateAfterCreateConflict(false)
 		}
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -292,6 +302,28 @@ func (c AddOrFallbackOnUpdateStrategy) Apply() error {
...
+
+func (c AddOrFallbackOnUpdateOrNoopStrategy) Op() ClusterChangeApplyStrategyOp {
+	return createStrategyFallbackOnUpdateOrNoopAnnValue
+}
+
+func (c AddOrFallbackOnUpdateOrNoopStrategy) Apply() error {
+	createdRes, err := c.aou.identifiedResources.Create(c.newRes)
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			c.aou.opts.allowNoopUpdates = true
```
cppforlife: setting state on the object seems unnecessary -- we should just pass that into the function we are calling.

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,106 @@
...
+			"--existing-non-labeled-resources-check=false", "--dangerous-override-ownership-of-existing-resources", "-y", "--json"},
+			RunOpts{AllowError: false, StdinReader: strings.NewReader(objYaml)})
+		assert.NoError(t, err)
+
+		out := kapp.Run([]string{"inspect", "-a", name2})
+		assert.Contains(t, out, "precious-resource")
+		out = kapp.Run([]string{"inspect", "-a", name})
+		assert.NotContains(t, out, "precious-resource")
+	})
+
```
cppforlife: nit: extra newline

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,106 @@
...
+	objYaml := strings.Replace(`
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: precious-resource
+  namespace: __ns__
+  annotations:
+    kapp.k14s.io/create-strategy: "fallback-on-update-or-noop"
+data: {"refName":"pkg9.test.carvel.dev","releasedAt":null,"version":"0.0.0"}
```
cppforlife: "package like" data makes things seems that's somehow related. lets change to:

```
data:
  newKey: ""
```

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,105 @@
...
+		kapp.Run([]string{"delete", "-a", name})
+		kapp.Run([]string{"delete", "-a", name2})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy initial ns and package", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "-y"},
+			RunOpts{AllowError: false, StdinReader: strings.NewReader(nsYaml + objYaml)})
```
cppforlife: AllowError: false (and below) is unncessary.

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,105 @@
...
+          #@overlay/match missing_ok=True
+          annotations:
+            #@overlay/match missing_ok=True
+            kapp.k14s.io/noop: ""
+  resourceMatchers:
+  - apiVersionKindMatcher: {apiVersion: v1, kind: ConfigMap}
+`
+
+	name := "test-create-fallback-on-noop"
+	name2 := "test-2"
```
cppforlife: test-2 => test-create-fallback-on-noop2

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,105 @@
...
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+
+	objNs := env.Namespace + "-create-fallback-on-noop"
+	nsYaml := strings.Replace(`
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: __ns__
+`, "__ns__", objNs, -1)
```
cppforlife: not sure why we need ns... i think we can remove it.

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,105 @@
...
+	objYaml := strings.Replace(`
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: precious-resource
+  namespace: __ns__
```
cppforlife: we can avoid specifying namespace and kapp will fill it with env configured ns.

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,105 @@
...
+		kapp.Run([]string{"delete", "-a", name2})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy initial ns and package", func() {
+		_, err := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "-y"},
+			RunOpts{AllowError: false, StdinReader: strings.NewReader(nsYaml + objYaml)})
+		assert.NoError(t, err)
```
praveenrewar: Should we use `require` instead of `assert` to be consistent with other tests?

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,105 @@
...
+	})
+
+	logger.Section("assert theres a configmap that belongs to the app", func() {
```
praveenrewar: Maybe we should include the assertions and the deployment in the same section (just like `redeploy the second app without the rebase rule and it steals ownership of the configmap`)?

test/e2e/create_fallback_on_noop_test.go
```diff
@@ -0,0 +1,105 @@
+// Copyright 2020 VMware, Inc.
```
praveenrewar: Super nitty mode on 😁 
```suggestion
// Copyright 2022 VMware, Inc.
```

test/e2e/create_fallback_on_noop_test.go
joe-kimmel-vmw: the difference between require and assert is whether it keeps going --  a failed `require` halts the test immediately,  whereas a failed `assert` will continue (and then other assertions may also pass or fail).  so i don't think consistency is the right reason to prefer one over the other.     However in this case there's a require inside `kapp.RunWithOpts` that will fail if `AllowError` is false (which per DK's other comments, it is by default). 
 
   so on further inspection this check is really gratuitous regardless of whether require or assert is used.

test/e2e/create_fallback_on_noop_test.go
joe-kimmel-vmw: that's cool --  I cargo-culted it over from the test for the original fallback_on_... feature: https://github.com/vmware-tanzu/carvel-kapp/blob/develop/test/e2e/create_fallback_on_update_test.go#L31

I was wondering why it was important but didn't bother trying to remove it. :)

test/e2e/create_fallback_on_noop_test.go
joe-kimmel-vmw: i collapsed this one into the above section

praveenrewar (approved):
LGTM!

renuy:
Available as part of kapp v0.47.0

PR #485, opened 2022-04-14
Remove hardcoded namespace in  e2e tests

### LINKED ISSUE #484: e2e tests should not have hardcoded namespace `kapp-test`
**What steps did you take:**
```bash
export KAPP_E2E_NAMESPACE=test-ns
./hack/test-e2e.sh
```

**What happened:**
A few tests failed because they have hardcoded namespace `kapp-test` and hence require using a ns with the same name for running the e2e tests.

**What did you expect:**
Tests should have ran successfully.

If we currently use 
```bash
export KAPP_E2E_NAMESPACE=kapp-test
./hack/test-e2e.sh
```
the tests run successfully as the namespace from the environment variable matches with the hardcoded namespace
An [example](https://github.com/vmware-tanzu/carvel-kapp/blob/4efc4822852b4077de8a3e6d3887d67529ff30f6/test/e2e/annotations_test.go#L95) of hardcoded namespace. 
We need to update the tests to use`env.Namespace` instead of `kapp-test` so that setting any namespace as ` KAPP_E2E_NAMESPACE` would work.

**Environment:**

- kapp version (use `kapp --version`):

---
Vote on this request

EdwinWalela:
This seeks to close #484. I have replaced all instances of `kapp-test` in `e2e test` with `env.Namespace`


test/e2e/config_test.go
```diff
@@ -29,7 +29,7 @@ rebaseRules:
   resourceMatchers:
   - kindNamespaceNameMatcher:
       kind: ConfigMap
-      namespace: kapp-test
+      namespace: ` + env.Namespace + `
```
praveenrewar: What do you think about using `fmt.Sprintf()` to format it?

test/e2e/filter_test.go
```diff
@@ -62,7 +62,7 @@ data:
 			RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
 		expectedOutput1 := `
 Namespace  Name           Kind       Age  Op      Op st.  Wait to    Rs  Ri  
-kapp-test  redis-config   ConfigMap  -    create  -       reconcile  -   -  
+` + env.Namespace + `  redis-config   ConfigMap  -    create  -       reconcile  -   -
```
praveenrewar: Maybe use `fmt` or `strings` to format or replace namespace?

Also, outputs like these would need some special handling as the indentation would change based on the env.Namespace value. We could create separate PR to keep the changes minimal for now, but feel free to make the changes in the same PR itself.

test/e2e/template_test.go
```diff
@@ -99,7 +99,7 @@ data:
 ` + depYAML
 
 	expectedYAML1Diff := `
-@@ create configmap/config-ver-1 (v1) namespace: kapp-test @@
+@@ create configmap/config-ver-1 (v1) namespace: ` + env.Namespace + ` @@
```
praveenrewar: How about using `strings.Replace()` to replace the occurrences of test namespace?

test/e2e/config_test.go
EdwinWalela: That's actually a cleaner approach. `strings` would also work but `fmt.Sprintf()` would be more readable. Though I'm leaning more towards replace since it will be less repetitive compared to `fmt.Sprintf()`

test/e2e/template_test.go
EdwinWalela: Nice catch. Yes, `strings.Replace()` would be a better approach

test/e2e/filter_test.go
EdwinWalela: `strings` would be a good alternative. Though as you mentioned this might fail due to indentation. Currently, I don't have a solution to handle varying indentation.

test/e2e/config_test.go
praveenrewar: I guess we could use `fmt` when we need to add the namespace to one or two places, otherwise `strings.Replace` would be a good option.
Also, I think it would be good to use something like `<e2e-test-ns>` instead of `kapp-test` and then replace it with `env.Namespace` so that it is evident while reading the yaml that this value will be replaced.

praveenrewar: There are a couple of other places where you might wanna make this change :)

test/e2e/filter_test.go
praveenrewar: > Currently, I don't have a solution to handle varying indentation.

That's okay, we can have a separate PR to handle this. You can either leave it as it is for now (i.e. with hardcoded "kapp-test")

test/e2e/template_test.go
praveenrewar: Same as above
`
Also, I think it would be good to use something like <e2e-test-ns> instead of kapp-test and then replace it with env.Namespace so that it is evident while reading the yaml that this value will be replaced.
`

EdwinWalela:
@praveenrewar I opted to use the namespace placeholder `<e2e-test-ns>` and `strings` to replace instances of the placeholder with `env.Namespace`. It turns out it works well for varying indentation e.g in `filter_test.go`

test/e2e/filter_test.go
```diff
@@ -62,12 +62,13 @@ data:
 			RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
 		expectedOutput1 := `
 Namespace  Name           Kind       Age  Op      Op st.  Wait to    Rs  Ri
```
praveenrewar: Do you think something like this would be more readable?

```go
		expectedOutput1 := fmt.Sprintf(`
Namespace  Name           Kind       Age  Op      Op st.  Wait to    Rs  Ri  
%s  redis-config   ConfigMap  -    create  -       reconcile  -   -`, env.Namespace)
```

praveenrewar: Same for other similar places.

praveenrewar (commented):
I think we are almost done. Just one last comment (minor).

praveenrewar:
> It turns out it works well for varying indentation e.g in filter_test.go

Hmm, I don't think that would work, unless we use a namespace with same of characters as `kapp-test`. But that's okay, we can have a separate PR for it.

EdwinWalela:
```go
expectedOutput1 := fmt.Sprintf(`
Namespace  Name           Kind       Age  Op      Op st.  Wait to    Rs  Ri  
%s  redis-config   ConfigMap  -    create  -       reconcile  -   -`, env.Namespace)
```

Yes, this would be more readable. Also, this seems like it would cater to the varying indentation for different namespaces

praveenrewar:
> Also, this seems like it would cater to the varying indentation for different namespaces

The indentation varies based on the number of characters in each cell of the table.

EdwinWalela:
@praveenrewar I refactored to use `fmt.Sprintf` when defining the expected output and where instances of namespaces to be replaced was only one. Also did a rebase with the `develop` branch.

praveenrewar:
@EdwinWalela Thank you so much for working on the suggestions. 
You might wanna do a force push after rebasing to keep the commit history on the PR clean.

EdwinWalela:
@praveenrewar  Currently I have no new changes on my branch, how do I approach this?

praveenrewar:
I think you did a pull from remote after doing the rebase and then pushed the changes. I guess you will have to undo the rebase using `reset` and then rebase again, followed by force push.

praveenrewar:
Hey @EdwinWalela! Are you planning to open another PR for this? What exactly happened with the cla (did you get any reason as to why it was rejected)? You can try signing it again and if needed we can get some help from the folks who maintain the clabot.

PR #487, opened 2022-04-15
Use noop for previous versions of a resource

### LINKED ISSUE #461: Allow versioned resources to have failed previous versions.

When I use versioned job resources for data migrations, it would be useful to allow failed previous versions. In my development workflow these data migrations often fail and require developers to diagnose and deploy new versions. Ideally in this workflow:
- Developers don't need to manually modify (i.e. delete) k8s resources. Our deployments are all managed by a CI system, triggered when developers push code to git.
- Developers can keep the history of the failed previous versions.

Allow versioned resources to have failed previous versions. Perhaps through annotation configuration.

This is a followup from Slack:
https://kubernetes.slack.com/archives/CH8KCCKA5/p1647720204911829

---
Vote on this request

### LINKED ISSUE #491: Using --diff-changes when a failed version of a resource is already present in the cluster results in panic
**What steps did you take:**
`job-fail.yaml`
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: my-job
  annotations:
    kapp.k14s.io/versioned: ""
spec:
  template:
    metadata:
      name: my-job
    spec:
      restartPolicy: Never
      containers:
        - name: my-job
          image: alpine:3.15.0
          command: [ "sh", "-c", "exit 1" ]
  backoffLimit: 0
  ```
  ```bash
  $ kapp deploy -a test-diff -f job-fail.yaml
  ...
  Changes

Namespace  Name          Kind  Conds.  Age  Op      Op st.  Wait to    Rs  Ri
default    my-job-ver-1  Job   -       -    create  -       reconcile  -   -
...
kapp: Error: waiting on reconcile job/my-job-ver-1 (batch/v1) namespace: default:
  Finished unsuccessfully (Failed with reason BackoffLimitExceeded: Job has reached the specified backoff limit)
  
  
$  kapp deploy -a test-diff -f job-fail.yaml -c
@@ noop job/my-job-ver-1 (batch/v1) namespace: default @@
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1fe8e95]
  ```

**What happened:**
When deploying an app which already consists of a failed job, kapp panics when using the `-c` option.

**What did you expect:**
kapp shouldn’t panic and show an empty since there's no change made to the resource.

**Environment:**

- kapp version (use `kapp --version`): v0.46.0

---
Vote on this request

### PR DESCRIPTION
**Pull request by praveenrewar**
Use noop for previous versions of a resource and use keep for the latest existing version if there are no changes (Fixes #461)

`Scenarios`
Consider an app that contains a versioned `Job` and has some existing versions on the cluster.
```
Namespace           Name                   Kind       Owner    Rs    Ri                                   Age  
versioned-job-test  my-job-ver-1           Job        kapp     ok    Completed                            17s  
^                   my-job-ver-1--1-5zrjw  Pod        cluster  ok    -                                    17s  
^                   my-job-ver-2           Job        kapp     fail  Failed with reason                   8s  
                                                                     BackoffLimitExceeded: Job has          
                                                                     reached the specified backoff limit
```
Now if we make some changes to the Job and deploy it again, kapp should use `noop` for the previous versions, (i.e. `ver-1` and `ver-2`) and if we don't make any change to the Job, then kapp should use `keep` for `ver-2` and `noop` for `ver-1`.

If we remove the `versioned` annotation, then the behaviour should remain the same (delete all the existing versions and create a new non versioned job)

Usage of `versioned-keep-original` should remain the same.

---

Add configurableTextDiff to existing versions of a resource when op is ChangeOpKeep (Fixes #491 )
- Refactor the constructor NewChangePrecalculated to accept ChangeOp, ConfigurableTextDiff and OpsDiff

### DIFF
--- a/pkg/kapp/diff/change_set_with_versioned_rs.go
+++ b/pkg/kapp/diff/change_set_with_versioned_rs.go
@@ -42,21 +42,21 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
 
 	// First try to calculate changes will update references on all resources
 	// (which includes versioned and non-versioned resources)
-	_, _, err := d.addChanges(newRs, existingRsGrouped)
+	_, _, err := d.addAndKeepChanges(newRs, existingRsGrouped)
 	if err != nil {
 		return nil, err
 	}
 
 	// Since there might have been circular dependencies;
 	// second try catches ones that werent changed during first run
-	addChanges, alreadyAdded, err := d.addChanges(newRs, existingRsGrouped)
+	addChanges, alreadyAdded, err := d.addAndKeepChanges(newRs, existingRsGrouped)
 	if err != nil {
 		return nil, err
 	}
 
 	allChanges = append(allChanges, addChanges...)
 
-	keepAndDeleteChanges, err := d.keepAndDeleteChanges(existingRsGrouped, alreadyAdded)
+	keepAndDeleteChanges, err := d.noopAndDeleteChanges(existingRsGrouped, alreadyAdded)
 	if err != nil {
 		return nil, err
 	}
@@ -113,7 +113,7 @@ func (d ChangeSetWithVersionedRs) assignNewNames(
 	}
 }
 
-func (d ChangeSetWithVersionedRs) addChanges(
+func (d ChangeSetWithVersionedRs) addAndKeepChanges(
 	newRs versionedResources, existingRsGrouped map[string][]ctlres.Resource) (
 	[]Change, map[string]ctlres.Resource, error) {
 
@@ -139,6 +139,7 @@ func (d ChangeSetWithVersionedRs) addChanges(
 			case ChangeOpKeep:
 				// Use latest copy of resource to update affected resources
 				usedRes = existingRes
+				changes = append(changes, d.newKeepChange(existingRes))
 			default:
 				panic(fmt.Sprintf("Unexpected change op %s", updateChange.Op()))
 			}
@@ -182,7 +183,7 @@ func (d ChangeSetWithVersionedRs) newAddChangeFromUpdateChange(
 	return addChange
 }
 
-func (d ChangeSetWithVersionedRs) keepAndDeleteChanges(
+func (d ChangeSetWithVersionedRs) noopAndDeleteChanges(
 	existingRsGrouped map[string][]ctlres.Resource,
 	alreadyAdded map[string]ctlres.Resource) ([]Change, error) {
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/kapp/diff/change_set_with_versioned_rs.go [hunk 1]
```diff
@@ -42,21 +42,21 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
-	_, _, err := d.addChanges(newRs, existingRsGrouped)
+	_, _, _, err := d.addChanges(newRs, existingRsGrouped)
 	if err != nil {
 		return nil, err
 	}
 
 	// Since there might have been circular dependencies;
 	// second try catches ones that werent changed during first run
-	addChanges, alreadyAdded, err := d.addChanges(newRs, existingRsGrouped)
+	addChanges, alreadyAdded, alreadyAddedKeepOp, err := d.addChanges(newRs, existingRsGrouped)
```
praveenrewar: Maybe use the same map (`alreadyAdded`) with `Change` as a value instead of `Resource`?

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -175,12 +175,7 @@ func (d ChangeSetWithVersionedRs) newAddChangeFromUpdateChange(
 	newRes ctlres.Resource, updateChange Change) Change {
 
 	// Use update's diffs but create a change for new resource
-	addChange := NewChangePrecalculated(nil, newRes, newRes)
-	// TODO private field access
-	addChange.op = ChangeOpAdd
-	addChange.configurableTextDiff = updateChange.ConfigurableTextDiff()
-	addChange.opsDiff = updateChange.OpsDiff()
-	return addChange
+	return NewChangePrecalculated(nil, newRes, newRes, ChangeOpAdd, updateChange.ConfigurableTextDiff(), updateChange.OpsDiff())
```
100mik, line 178: Do you think a comment which elaborates on why this change was necessary will help?

100mik (approved):
I think the functions `existingVersionedResources` and `newVersionedResources` could be scoped as methods to a struct, but that could be a separate PR.
Left a thought. but no strong opinions on it

LGTM!

pkg/kapp/diff/change_set_with_versioned_rs.go
praveenrewar: I think the TODO `// TODO private field access` explains why this is changed.

I feel that comments should be added to explain why something is done (for future reference) or should be done (TODOs) instead of why something is changed (as that is only relevant for the PR) 😄  
But do let me know if the new change doesn't make sense and requires a comment.

praveenrewar:
> I think the functions existingVersionedResources and newVersionedResources could be scoped as methods to a struct, but that could be a separate PR.

I think these are sort of constructors for `versionedResources` and hence not scoped.

PR #490, opened 2022-04-19
Support migrated apps when migration is disabled

### LINKED ISSUE #146: add suffix/prefix to name of kapp created metadata configmaps to avoid configmap naming collisions

praveenrewar:
- Apps migrated to use the ConfigMap suffix should be supported (seen without suffix) even if migration is disabled
- Remove `isKappApp` check as we don't have that with existing rename and update as well. If we want to, we can add it in all places later on.

`Expected behaviour`
```bash
$ KAPP_FQ_CONFIGMAP_NAMES=true kapp deploy -a my-app -f app.yml --yes


$ KAPP_FQ_CONFIGMAP_NAMES=true kapp ls
Namespace  Name        Namespaces  Lcs   Lca
default    my-app      default     true  7s

$ kapp ls
Namespace  Name        Namespaces  Lcs   Lca
default    my-app      default     true  7s

$ kapp inspect -a my-app
...
Resources in app 'my-app'

Namespace  Name  Kind       Owner  Rs  Ri  Age
...
```

`Note`: This leads to an additional call to get configmap with fully qualified name when migration is disabled.

### DIFF
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -133,38 +133,36 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 		}.AsData(),
 	}
 
-	if a.isMigrationEnabled() {
-		migratedAnnotation := map[string]string{KappIsConfigmapMigratedAnnotationKey: KappIsConfigmapMigratedAnnotationValue}
+	migratedAnnotation := map[string]string{KappIsConfigmapMigratedAnnotationKey: KappIsConfigmapMigratedAnnotationValue}
 
-		configMapWithSuffix, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
-		if err == nil {
-			if a.isKappApp(configMapWithSuffix) {
-				return a.updateApp(configMapWithSuffix, labels)
-			}
-		} else if !errors.IsNotFound(err) {
-			// return if error is anything other than configmap not found
-			return fmt.Errorf("Getting app: %s", err)
-		}
-
-		configmapWithoutSuffix, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
-		if err == nil {
-			if a.isKappApp(configmapWithoutSuffix) {
-				return a.migrate(configmapWithoutSuffix, labels, a.fqName)
-			}
-		} else if !errors.IsNotFound(err) {
-			// return if error is anything other than configmap not found
-			return fmt.Errorf("Getting app: %s", err)
-		}
-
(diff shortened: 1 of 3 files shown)


pkg/kapp/app/recorded_app.go
```diff
@@ -133,38 +133,36 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
...
-		}
+	configMapWithSuffix, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.fqName, metav1.GetOptions{})
+	if err == nil {
+		return a.updateApp(configMapWithSuffix, labels)
+	}
 
+	if errors.IsNotFound(err) {
 		configmapWithoutSuffix, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), a.name, metav1.GetOptions{})
 		if err == nil {
-			if a.isKappApp(configmapWithoutSuffix) {
```
neil-hickey, line 151: Do we want to keep the isKappApp check? It looks like we dropped it

pkg/kapp/app/recorded_app.go
```diff
@@ -457,11 +435,6 @@ func (a *RecordedApp) labeledApp() (*LabeledApp, error) {
 	return &LabeledApp{sel, a.identifiedResources}, nil
 }
 
-func (a *RecordedApp) isKappApp(app *corev1.ConfigMap) bool {
```
neil-hickey, line 460: Same as above -  just wondering if the assumptions changed here - or why we dropped this

neil-hickey: Actually I think I answered my own question - we are assuming the suffix has meaning and is a kapp app by nature right?

neil-hickey (approved):
LGTM 👍

sethiyash (approved):
Looks Good.

kumaritanushree (approved):

pkg/kapp/app/recorded_app.go
praveenrewar: Yep, we are assuming that the suffix is internal and people will not use it.
Also, another reason to remove the `isKappApp` check was to make it consistent with current deploy and rename.
> Remove isKappApp check as we don't have that with existing rename and update as well. If we want to, we can add it in all places later on.

PR #497, opened 2022-04-26
display pod logs respecting the value provided in kapp.k14s.io/deploy-log annotation

### LINKED ISSUE #460: deploy logs always display logs for existing pods
**Issue**

I've noticed that for resources annotated with `kapp.k14s.io/deploy-logs`, kapp is printing logs for all of my new and existing pods. This is true both with the default value of `""` and with the specified valued `"for-new"`. According to the docs, both values should omit logs for existing pods and only show logs for new pods.

**Environment:**

- kapp: 0.46.0
- OS: alpine 14 (container)
- kubectl version: 1.21.6
- k8s cluster version: 1.21.6-gke.1503

---
Vote on this request

sethiyash:
Issue - #460 

`kapp.k14s.io/deploy-log`  with the default value of "" and with the specified valued "for-new" `kapp` should omit logs for existing pods and only show logs for new pods.

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -268,7 +268,8 @@ func (o *DeployOptions) newAndUsedGKs(newGKs []schema.GroupKind, app ctlapp.App)
 		Kind: "Pod",
 	}
 
-	if _, exists := gksByGK[podGK]; !exists {
+	// adding Pod in GKs to get existing Pod resources (#460)
+	if _, exists := gksByGK[podGK]; !exists && o.DeployFlags.Logs {
 		uniqGKs = append(uniqGKs, podGK)
 	}
 
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -144,6 +144,10 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	if o.DeployFlags.Logs {
+		usedGKs = append(usedGKs, schema.GroupKind{Kind: "Pod"})
+	}
+
 	existingResources, existingPodRs, err := o.existingResources(
 		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs)
 	if err != nil {
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -144,10 +144,6 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
-	if o.DeployFlags.Logs {
-		usedGKs = append(usedGKs, schema.GroupKind{Kind: "Pod"})
-	}
-
 	existingResources, existingPodRs, err := o.existingResources(
 		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs)
 	if err != nil {
@@ -266,6 +262,14 @@ func (o *DeployOptions) newAndUsedGKs(newGKs []schema.GroupKind, app ctlapp.App)
 			gksByGK[gk] = struct{}{}
 			uniqGKs = append(uniqGKs, gk)
 		}
+	}
+
+	podGK := schema.GroupKind{
+		Kind: "Pod",
+	}
+
+	if _, exists := gksByGK[podGK]; !exists {
+		uniqGKs = append(uniqGKs, podGK)
 	}
 
 	return uniqGKs, nil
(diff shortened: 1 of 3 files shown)


pkg/kapp/cmd/app/deploy.go
```diff
@@ -144,6 +144,10 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	if o.DeployFlags.Logs {
```
sethiyash: This flag is by default true as of now so here we will fetch Pods for every deploy operation. Other possible conditions for adding Pod in usedGKs - 
1. Parsing yaml and checking whether `kapp.k14s.io/deploy-log` annotation exists in the provided yaml. 
2. <s> Start checking earlier whether the PODs created for the given resource have the annotation or not as we check [here](https://github.com/vmware-tanzu/carvel-kapp/blob/bfa32357c416f669ea5bb20c891baf936e7af038/pkg/kapp/cmd/app/deploy.go#L470). </s>

100mik (dismissed):
Since GKs are just used for listing the resources, I do not see a major disadvantage of appending the GK for Pods when we are expecting to look for logs.

It will add one `List` call, and there are no subsequent calls if there are no pods.

100mik:
On a second thought, should we check how much of a difference this makes for an active cluster? See how much difference does it make? Not sure if it is already in place, but if the overhead is too high we could maybe disable podlogs for kapp-controller? (if it is not done already)

(Verified that kapp-controller does not set `--logs=false`)

sethiyash:
> On a second thought, should we check how much of a difference this makes for an active cluster? See how much difference does it make? Not sure if it is already in place, but if the overhead is too high we could maybe disable podlogs for kapp-controller? (if it is not done already)

Yeah sure, I can check this out.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -144,6 +144,10 @@ func (o *DeployOptions) Run() error {
...
+		usedGKs = append(usedGKs, schema.GroupKind{Kind: "Pod"})
```
cppforlife: should we dedup this if usedGKs already has Pod?

sethiyash: Ahh, my bad. We should not. Refactored the code to check for already existing Pod Group Kind in `newAndUsedGKs()` method.

cppforlife:
@sethiyash would be good to add an e2e test to catch this regression.

test/e2e/pod_log_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPodLogs(t *testing.T) {
```
cppforlife, line 14: to simplify the test, can't we just deploy Pod resources directly instead of using Deployment? alternatively we can use statefulset so that names of pods are deterministic.

sethiyash: Yeah sounds good. Started using `StatefulSet`.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -264,6 +264,15 @@ func (o *DeployOptions) newAndUsedGKs(newGKs []schema.GroupKind, app ctlapp.App)
 		}
 	}
 
+	podGK := schema.GroupKind{
+		Kind: "Pod",
+	}
+
+	// adding Pod in GKs to get existing Pod resources (#460)
+	if _, exists := gksByGK[podGK]; !exists && o.DeployFlags.Logs {
+		uniqGKs = append(uniqGKs, podGK)
```
100mik, line 273: I think we can do this inline 
```suggestion
		uniqGKs = append(uniqGKs, schema.GroupKind{Kind: "Pod"})
```
A variable seems a bit unnecessary

100mik (commented):
Just a thought, not blocking

pkg/kapp/cmd/app/deploy.go
sethiyash: But we are using `podGk` in this statement also before appending - 
> if _, exists := gksByGK[podGK]; !exists && o.DeployFlags.Logs {

100mik: Oh I see, my bad, LGTM then

sethiyash: That's why I thought have the podGk as variable and using it in both the places

100mik (approved):
LGTM!

PR #499, opened 2022-04-27
Add support for unblocking waiting changes in wait rule conditions

### LINKED ISSUE #427: Waiting for interdependent resources to become ready

I have two resources that are part of an app. Resource A will not be ready without resource B, and resource B will not become ready without resource A. I'd like to know for `kapp deploy` when both resources are ready.

If I define no wait rules or change rules then everything will come up, but I don't know when that happens via `kapp`.

If I define wait rules, then I know when both resources are up, but there's a race condition where one of the resource could be not ready and fail the `kapp deploy`.

If I define a change rule, then I can ensure that resource A is created before resource B. But because resource A will never become ready without resource B, I can't also specify a wait rule, as resource B will never be created as resource A is never ready.

More specific details below.

I'd like a way to order the upsert of resources via a change rule to wait for the resource to be "seen" without needing to wait for *all* of the wait rule conditions. Waiting for the resource to be created/updated on the api may be sufficient, or if a wait rule defines `supportsObservedGeneration: true` wait for the generations to match.

[Slack thread](https://kubernetes.slack.com/archives/CH8KCCKA5/p1643303143179500)

My specific use case involves trying to deploy a Knative `Service` (S) and a `ServiceBinding` (SB) resource. The S will crash loop until the SB is ready. The SB will not become ready until the S exists.

Prereqs:
1. Install [Knative Serving](https://knative.dev/docs/install/serving/install-serving-with-yaml/)
1. Install [Service Bindings](https://github.com/vmware-labs/service-bindings)
   ```
   kapp deploy -a service-bindings -f https://github.com/vmware-labs/service-bindings/releases/download/v0.6.0/service-bindings-0.6.0.yaml
   ```
1. Create a MySQL service
   ```
   kapp deploy -a mysql -f https://gist.github.com/scothis/f65bd3ec45bd42ca89b1a554442c7cdc/raw/467d84237f582c6615ae1f7543d30d3a5015f2c6/mysql.yaml
   ```

[Gist with config](https://gist.github.com/scothis/f65bd3ec45bd42ca89b1a554442c7cdc) I'm attempting to use. Has both wait rules and a change rule defined. It will fail as is.

   ```
   kapp deploy -a workload -f https://gist.github.com/scothis/f65bd3ec45bd42ca89b1a554442c7cdc/raw/467d84237f582c6615ae1f7543d30d3a5015f2c6/app.yaml
   ```

---
Vote on this request

praveenrewar:
Fixes: #427 

`unblockChanges` can be set true for a condition in wait rules of a resource, which will unblock all resources that are dependent on this resource.

`Example`
```
kind: Config
waitRules:
  conditionMatchers:
  - type: Progressing
    status: "True"
    success: true
    unblockChanges: true
  - type: Available
    status: "True"
    success: true
  resourceMatchers:
  - apiVersionKindMatcher: {apiVersion: apps/v1, kind: Deployment}
---
apiVersion: v1
kind: Service
metadata:
  name: simple-app
  annotations:
    kapp.k14s.io/change-rule: "upsert after upserting dep"
...
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: simple-app
  annotations:
    kapp.k14s.io/change-group: "dep"
...
```

In the above example, the service will be created after the deployment, based on the change group and change rules. But the `Progressing` condition on the deployment has the `unblockChanges` set to true (in the wait rules provided as part of config) and therefore the service will be created as soon as that condition is met.

### DIFF
--- a/pkg/kapp/config/config.go
+++ b/pkg/kapp/config/config.go
@@ -53,6 +53,7 @@ type WaitRuleConditionMatcher struct {
 	Failure                    bool
 	Success                    bool
 	SupportsObservedGeneration bool
+	SupportsUnblockingChanges  bool
 }
 
 type WaitRuleYtt struct {
--- a/pkg/kapp/config/config.go
+++ b/pkg/kapp/config/config.go
@@ -53,7 +53,7 @@ type WaitRuleConditionMatcher struct {
 	Failure                    bool
 	Success                    bool
 	SupportsObservedGeneration bool
-	SupportsUnblockingChanges  bool
+	UnblockChanges             bool
 }
 
 type WaitRuleYtt struct {
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -87,6 +87,8 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 		}
 	}
 
+	unblockingWaitingChangeMsg := ""
+
 	// If no failure conditions found, check on successful ones
 	for _, condMatcher := range s.waitRule.ConditionMatchers {
 		for _, cond := range obj.Status.Conditions {
@@ -96,6 +98,12 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 					continue
 				}
 				if condMatcher.Success {
+					if condMatcher.SupportsUnblockingChanges {
+						unblockingWaitingChangeMsg = fmt.Sprintf(
+							"Allowing blocked changes to proceed: Encountered successful condition %s == %s: %s",
+							cond.Type, condMatcher.Status, cond.Reason)
+						continue
+					}
 					return DoneApplyState{Done: true, Successful: true, Message: fmt.Sprintf(
 						"Encountered successful condition %s == %s: %s (message: %s)",
 						cond.Type, condMatcher.Status, cond.Reason, cond.Message)}
@@ -108,5 +116,10 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 		return DoneApplyState{Done: false, Message: fmt.Sprintf(
 			"Waiting for generation %d to be observed by status condition(s)", obj.Metadata.Generation)}
 	}
+
+	if unblockingWaitingChangeMsg != "" {
+		return DoneApplyState{Done: false, UnblockBlockedChanges: true, Message: unblockingWaitingChangeMsg}
+	}
+
 	return DoneApplyState{Done: false, Message: "No failing or successful conditions found"}
 }
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -65,8 +65,12 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 			return DoneApplyState{Done: true, Successful: false, Message: fmt.Sprintf(
 				"Error: Applying ytt wait rule: %s", err.Error())}
 		}
+		message := configObj.Message
+		if configObj.UnblockBlockedChanges {
+			message = fmt.Sprintf("Allowing blocked changes to proceed: %s", configObj.Message)
+		}
 		return DoneApplyState{Done: configObj.Done, Successful: configObj.Successful,
-			Message: configObj.Message}
+			UnblockBlockedChanges: configObj.UnblockBlockedChanges, Message: message}
 	}
 
 	hasConditionWaitingForGeneration := false
--- a/pkg/kapp/resourcesmisc/custom_waiting_resource.go
+++ b/pkg/kapp/resourcesmisc/custom_waiting_resource.go
@@ -66,11 +66,11 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 				"Error: Applying ytt wait rule: %s", err.Error())}
 		}
 		message := configObj.Message
-		if configObj.UnblockBlockedChanges {
+		if configObj.UnblockChanges {
 			message = fmt.Sprintf("Allowing blocked changes to proceed: %s", configObj.Message)
 		}
 		return DoneApplyState{Done: configObj.Done, Successful: configObj.Successful,
-			UnblockBlockedChanges: configObj.UnblockBlockedChanges, Message: message}
+			UnblockChanges: configObj.UnblockChanges, Message: message}
 	}
 
 	hasConditionWaitingForGeneration := false
@@ -91,7 +91,7 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 		}
 	}
 
-	unblockingWaitingChangeMsg := ""
+	unblockChangeMsg := ""
 
 	// If no failure conditions found, check on successful ones
 	for _, condMatcher := range s.waitRule.ConditionMatchers {
@@ -102,15 +102,15 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 					continue
 				}
 				if condMatcher.Success {
-					if condMatcher.SupportsUnblockingChanges {
-						unblockingWaitingChangeMsg = fmt.Sprintf(
-							"Allowing blocked changes to proceed: Encountered successful condition %s == %s: %s",
-							cond.Type, condMatcher.Status, cond.Reason)
-						continue
-					}
 					return DoneApplyState{Done: true, Successful: true, Message: fmt.Sprintf(
 						"Encountered successful condition %s == %s: %s (message: %s)",
 						cond.Type, condMatcher.Status, cond.Reason, cond.Message)}
+				}
+				if condMatcher.UnblockChanges {
+					unblockChangeMsg = fmt.Sprintf(
+						"Allowing blocked changes to proceed: Encountered condition %s == %s: %s",
+						cond.Type, condMatcher.Status, cond.Reason)
+					continue
 				}
 			}
 		}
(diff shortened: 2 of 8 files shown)


pkg/kapp/config/config.go
```diff
@@ -52,6 +52,7 @@ type WaitRuleConditionMatcher struct {
 	Failure                    bool
 	Success                    bool
 	SupportsObservedGeneration bool
+	SupportsUnblockingChanges  bool
```
praveenrewar: Open to suggestions on choosing a better name for this. (Other alternative I am inclined towards is `unblockBlockedChanges`)

100mik: Might be missing context here, when did we move away from "non-blocking"

praveenrewar: I closed that PR sometime back as it wasn't aligning with the ask on issue. [This](https://github.com/vmware-tanzu/carvel-kapp/issues/427#issuecomment-1092355922) comment might help in getting to know more about the need.

sethiyash (approved):
LGTM!

pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -96,6 +98,12 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 					continue
 				}
 				if condMatcher.Success {
+					if condMatcher.SupportsUnblockingChanges {
```
cppforlife: location of this impl seems to imply that it only works with custom waiting resource [0] which potentially becomes problematic if you want to "unblock" x but still want regular waiting logic. any thoughts?

[0] https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/clusterapply/converged_resource_factory.go#L47

praveenrewar: So in the e2e test example, we should not be forced to add the `Available` condition for the deployment as that's something we already have for deployments, right?
Hmm..., adding that condition did bother me but I didn't worry much about it thinking that it would mostly be used for custom resources. But this is definitely a valid concern, I will think more about it.

cppforlife: > So in the e2e test example, we should not be forced to add the Available condition for the deployment as that's something we already have for deployments, right?

as soon as you use custom waiting, code wont use "default" waiting rules (like for Deployment)

praveenrewar: Yep, and I am not able to think of a simple way to make this work for all resources (not just custom waiting resources) using `waitRules` because of this. 
Maybe we want to introduce some sort of `unblockingRules` that are only used for unblocking blocked changes when one of the mentioned conditions matches? (Although this sounds a bit redundant from user perspective, so maybe not the best way to go about it)

praveenrewar: We discussed to go forward with this approach for now and work on updating the logic for custom waiting resources so that they can be used along with default waiting rules available for other resources. I will create an issue for this and resolve this thread.

praveenrewar: Created a new [issue](https://github.com/vmware-tanzu/carvel-kapp/issues/512) to track this enhancement.

cppforlife: seems like adding this here is not enough if we would want to support same feature from ytt wait rules as well. (seems a bit awkward if we do not).

pkg/kapp/resourcesmisc/done_apply_state.go
```diff
@@ -4,9 +4,11 @@
 package resourcesmisc
 
 type DoneApplyState struct {
-	Done       bool   `json:"done"`
-	Successful bool   `json:"successful"`
-	Message    string `json:"message"`
+	Done       bool
```
cppforlife: json marks should be retained as they were added for ytt waiting.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
praveenrewar: Hmm, I didn't consider the ytt wait rules. I will make some changes.

pkg/kapp/resourcesmisc/done_apply_state.go
praveenrewar: I guess I messed up while rebasing. Good catch!!

pkg/kapp/resourcesmisc/custom_waiting_resource.go
praveenrewar: Done.

sethiyash (commented):
Changes looks good to me. 
Should be also start using `unblockBlockedChanges` field in ytt waitRules e2e testcases (custom_wait_rules_test.go)

praveenrewar:
> Should be also start using unblockBlockedChanges field in ytt waitRules e2e testcases (custom_wait_rules_test.go)

Yeah, I am still trying to figure out the best way to have an e2e test for this with ytt wait rules.

sethiyash (approved):
Looks Good.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
cppforlife: hmm, i just reread this section and i'm actually stuck now on why we are "unblocking" from the success block. if it's successful then it would be considered as done (and if it's done, it's unblocked). do we need to have non-success/non-failure condition that only allows to unblock?

cppforlife: (e.g. in ytt, i could say im not "done", but i will unblock anyway)

praveenrewar: This was a great callout. Thank you so much!!
I have made the changes accordingly.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
```diff
@@ -95,6 +97,14 @@ func (s CustomWaitingResource) IsDoneApplying() DoneApplyState {
 					hasConditionWaitingForGeneration = true
 					continue
 				}
+
+				if condMatcher.SupportsUnblockingChanges {
```
cppforlife: we should probably do this section after the below if success section. if it's successful, we just want to get out as quick as possible.

test/e2e/custom_wait_rules_test.go
```diff
@@ -119,3 +120,144 @@ status:
...
+Wait to: 3 reconcile, 0 delete, 0 noop
+
+<replaced>: ---- applying 1 changes [0/3 done] ----
+<replaced>: create customresourcedefinition/crontabs.stable.example.com (apiextensions.k8s.io/v1) cluster
+<replaced>: ---- waiting on 1 changes [0/3 done] ----
+<replaced>: ok: reconcile customresourcedefinition/crontabs.stable.example.com (apiextensions.k8s.io/v1) cluster
+<replaced>: ---- applying 1 changes [1/3 done] ----
+<replaced>: create crontab/my-new-cron-object-1 (stable.example.com/v1) namespace: kapp-test
+<replaced>: ---- waiting on 1 changes [1/3 done] ----
+<replaced>:  ^ Unblock blocked changes
```
cppforlife: hmm, may be we should have some consistency between ytt and non-ytt unblock messaging. e.g.

`Allowing blocked changes to proceed: <ytt user message>`
`Allowing blocked changes to proceed: Encountered condition %s == %s: reason`

there should also be a line describing resource above the `^`. but there isnt in this output.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
praveenrewar: Totally makes sense.

test/e2e/custom_wait_rules_test.go
praveenrewar: > hmm, may be we should have some consistency between ytt and non-ytt unblock messaging. e.g.

Good idea. Should we do this for other messages as well (success and failure)?

> there should also be a line describing resource above the ^. but there isnt in this output.

Apparently, the order of lines is messed up (and it's consistent, so I am guessing it's happening in one of the replace functions). I will take a look at it.

pkg/kapp/resourcesmisc/custom_waiting_resource.go
praveenrewar: Done

test/e2e/custom_wait_rules_test.go
praveenrewar: > hmm, may be we should have some consistency between ytt and non-ytt unblock messaging. e.g.

Done.

praveenrewar: > Apparently, the order of lines is messed up (and it's consistent, so I am guessing it's happening in one of the replace functions).

`sortActionsInResult` was causing this, not sure why I added it in the first place (maybe I just copy pasted it from another test). I have removed it and the order if fixed now.

test/e2e/order_test.go
```diff
@@ -210,6 +210,102 @@ Succeeded
...
+kind: Config
+waitRules:
+- supportsObservedGeneration: true
+  conditionMatchers:
+  - type: Progressing
+    status: "True"
+    supportsUnblockingChanges: true
+  - type: Available
+    status: "True"
+    success: true
```
cppforlife, line 237: since this field is called success/failure=true, should we change the name of the field from `supportsUnblockingChanges` to `unblockChanges` (supports seems like a wrong naming since it's not about supporting anything).

praveenrewar: Yes, makes sense. I will make the change.

praveenrewar: Done.

PR #501, opened 2022-04-30
Refactor migration logic for recorded app

praveenrewar:
Here's an attempt at making the migration logic a bit more readable and simple. 

I realised that I we were trying to avoid updating the `createOrUpdate` with the migration logic and in doing so it resulted in `createOrUpdate` being used only for create operations and not for updates, and this led me to make these changes.

- Created new private methods `find` and `create`.
- The above methods are then used in `CreateOrUpdate`, `RenamePrevApp`, `Exists`, `Rename`, `setMeta`, etc,. to make them more simple.
- Removed `createOrUpdate`
- Set `isMigrated` at relevant places.

### DIFF
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -130,6 +130,43 @@ func (a *RecordedApp) UpdateUsedGVsAndGKs(gvs []schema.GroupVersion, gks []schem
 func (a *RecordedApp) CreateOrUpdate(labels map[string]string) error {
 	defer a.logger.DebugFunc("CreateOrUpdate").Finish()
 
+	app, foundMigratedApp, err := a.find(a.fqName())
+	if err != nil {
+		return err
+	}
+
+	if foundMigratedApp {
+		a.isMigrated = true
+		return a.updateApp(app, labels)
+	}
+
+	app, foundNonMigratedApp, err := a.find(a.name)
+	if err != nil {
+		return err
+	}
+
+	if foundNonMigratedApp {
+		if a.isMigrationEnabled() {
+			return a.migrate(app, labels, a.fqName())
+		}
+		return a.updateApp(app, labels)
+	}
+
+	return a.create(labels)
+}
+
+func (a *RecordedApp) find(name string) (*corev1.ConfigMap, bool, error) {
+	cm, err := a.coreClient.CoreV1().ConfigMaps(a.nsName).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, false, nil
+		}
+		return nil, false, fmt.Errorf("Getting app: %w", err)
+	}
+	return cm, true, nil
+}
+
+func (a *RecordedApp) create(labels map[string]string) error {
 	configMap := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      a.name,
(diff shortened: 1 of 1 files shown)


pkg/kapp/app/recorded_app.go
```diff
@@ -130,6 +130,77 @@ func (a *RecordedApp) UpdateUsedGVsAndGKs(gvs []schema.GroupVersion, gks []schem
...
+	foundMigratedApp, err := a.findAndUpdate(a.fqName(), labels)
+	if err != nil {
+		return err
+	}
+
+	if foundMigratedApp {
+		a.isMigrated = true
+		return nil
+	}
+
```
100mik, line 142: nit: Do we need this new line?

pkg/kapp/app/recorded_app.go
```diff
@@ -130,6 +130,77 @@ func (a *RecordedApp) UpdateUsedGVsAndGKs(gvs []schema.GroupVersion, gks []schem
...
+	if foundNonMigratedApp {
+		return nil
+	}
+
+	return a.create(labels)
+}
+
+func (a *RecordedApp) findAndUpdate(name string, labels map[string]string) (bool, error) {
+	cm, exists, err := a.find(name)
+
```
100mik: nit: I do not think we need new lines after every call of `find()`
(I think this goes for all calls)

100mik (commented):
Just a few nits!
Will take a closer look on my local.
Love how this makes it legible <3

sethiyash (approved):
Looks really good to me.

pkg/kapp/app/recorded_app.go
praveenrewar: I added this new line because the rest of the code is for when the migrated app isn't found. But yeah I do see that there are a lot of new lines now.

pkg/kapp/app/recorded_app.go
praveenrewar: Usually we check if the `err` is non nil and return if that's the case, but since I am checking for `exists` first, I thought of using a new line so that it's not confusing. But I can definitely remove the new line if it doesn't make sense to have one.

pkg/kapp/app/recorded_app.go
100mik: Yep looks pretty spaced out, no strong opinions though, looks like it could be more "trimmed"

pkg/kapp/app/recorded_app.go
100mik: I think it would have made sense if we were doing both checks later, but I think it makes sense to have no new line if we are checking values returned in the previous statement. (similar o what we do for `changed` or `isSecretCreated` in kctrl)
Once again, no strong opinions about this, just thought I would put it down

pkg/kapp/app/recorded_app.go
praveenrewar: Removed the newline below the declaration. Same for `RenamePrevApp`. Let me know if this looks fine.

pkg/kapp/app/recorded_app.go
praveenrewar: Done.

100mik (approved):
Changes look good! 
Thanks for making this one. LGTM!

neil-hickey (approved):
LGTM. 

This new pattern of doXANDY() felt odd, but I think it does make it more readable. I was trying to find a way to keep the flow of `doX() if x { return y() }` instead of the `doXandY()` style this introduces, but I couldn't see a nice way. I don't want to block the refactor as I think it's a step in a more readable direction, but if it's valuable I can go through and see if what I am trying to describe provides the value I am hoping it does...

```
foundMigratedApp, err := a.findAndUpdate(a.fqName(), labels)
if err != nil {
	return err
}

if foundMigratedApp {
	a.isMigrated = true	
	return nil
}
```

would love if this could remain something like: (only because I am a fan on functions doing one thing only - the pattern of AND or even OR I find confusing)

```
if exists, err := a.find(a.fqName()); err != nil {
    return err
}

if exists {
  a.isMigrated = true	
  return a.updateApp(labels)
}

**Comment by praveenrewar** [comment]
> only because I am a fan on functions doing one thing only

Yeah, that sounds good. I will give this a shot. The only reason I didn't do it was because of the `RenamePrevApp` (it would become too lengthy)

**Comment by praveenrewar** [comment]
@neil-hickey Thanks a lot for your suggestions, I have made the relevant changes. Let me know your thoughts on it 😄

**Comment by neil-hickey** [comment]
> @neil-hickey Thanks a lot for your suggestions, I have made the relevant changes. Let me know your thoughts on it 😄

I prefer this pattern, it's more explicit what is going on. For example:

```
foundMigratedApp, err := a.findAndUpdate(a.fqName(), labels)
if err != nil {
	return err
}

if foundMigratedApp {
	a.isMigrated = true	
	return nil
}
```

vs (now)

```
app, foundMigratedApp, err := a.find(a.fqName())
if err != nil {
	return err
}

if foundMigratedApp {
	a.isMigrated = true
	return a.updateApp(app, labels)
}
```

You can see that they are effectively the same even with the old refactor of update into the findAndUpdate function. To me, now it's really clear what's happening if the app is found.. we update it, whereas before it was kinda hidden. Curious to see what others think!

**Comment by praveenrewar** [comment]
I myself like the new changes more than previous one.

File: pkg/kapp/app/recorded_app.go [hunk 3]
```diff
@@ -130,6 +130,43 @@ func (a *RecordedApp) UpdateUsedGVsAndGKs(gvs []schema.GroupVersion, gks []schem
...
+	}
+
+	if !foundNonMigratedApp {
+		return a.create(labels)
+	}
+
+	if a.isMigrationEnabled() {
+		return a.migrate(app, labels, a.fqName())
+	}
+	return a.updateApp(app, labels)
```
praveenrewar: Does this make more sense? The current way is sort of a standard way of doing things in Go, but in this context I felt that it would make more sense to call `create` at the end.
```suggestion
	if foundNonMigratedApp {
	        if a.isMigrationEnabled() {
		       return a.migrate(app, labels, a.fqName())
	        }
	        return a.updateApp(app, labels)
	}
	return a.create(labels)
```

neil-hickey: I like it !

100mik (approved):
Still looking good to me 🚀

PR #511, opened 2022-05-09
Add `--prev-app` to delete command

neil-hickey:
- This PR assumes the user wants to see the changeset diff on the cli
  also when providing a --prev-app. Hence the use of the Factory twice.
- The case of both existing is interesting, for now it ignores the
  --prev-app existing if --app exists
- Nothing inside recorded_app.go needs to change with this pattern

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


neil-hickey:
Some things to note:

1. I chose to use the `Factory` twice - allows most of the code to remain as is in `delete.go` and `recorded_app.go` doesn't need to change at all.
2. This may introduce other problems I am not fully thinking through right now, I wanted to check with @praveenrewar and others to see what you all thought.
3. I wanted the user to see the diff and hit `yes` etc even for a `--prev-app` app. So this factory request for each app made most sense in that context as it shows this diff and confirmation already

Interesting case:
App1 exists, App2 exists

`kapp delete -a App1 --prev-app App2`

What would you expect the behaviour to me? Given that App2 should have been migrated and not exist at the same time, I expect App1 to be deleted and App2 to be ignored.

Any UX opinions on this case?

pkg/kapp/cmd/app/delete.go
```diff
@@ -70,8 +72,28 @@ func (o *DeleteOptions) Run() error {
...
-		return nil
+		if o.MigrationFlags.prevAppName != "" {
+			o.AppFlags.Name = o.MigrationFlags.prevAppName
+
+			app, supportObjs, err = Factory(o.depsFactory, o.AppFlags, o.ResourceTypesFlags, o.logger)
+			if err != nil {
+				return err
+			}
+
+			prevExists, prevNotExistsMsg, err := app.Exists()
```
praveenrewar: (No strong opinions)
```suggestion
			prevAppExists, prevAppNotExistsMsg, err := app.Exists()
```

pkg/kapp/cmd/app/migration_flags.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2020 VMware, Inc.
```
praveenrewar: Super nitty mode on 
```suggestion
// Copyright 2022 VMware, Inc.
```

pkg/kapp/cmd/app/migration_flags.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/spf13/cobra"
+)
+
+type MigrationFlags struct {
+	prevAppName string
+}
+
+func (s *MigrationFlags) Set(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&s.prevAppName, "prev-app", "", "Name of old app incase of migration")
```
praveenrewar: *Suggestion*: should we _not_ use the term migration for renaming? When we use the term migration I honestly feel that we are talking about the app suffix stuff and I get a bit confused 😅 
Maybe we can have:
```suggestion
	cmd.Flags().StringVar(&s.prevAppName, "prev-app", "", "Set previous name of app")
```
(Also, I know it's too late now but I feel that `--prev-name` might have been a good choice, it's similar to `--new-name` and also we would have used `prev-name` when `name` wasn't found vs `prev-app` when `name`  isn't found now.)

praveenrewar: P.S. We might also want to rename the file and the struct if we go ahead with the the above suggestion.

praveenrewar:
> I chose to use the Factory twice - allows most of the code to remain as is in delete.go and recorded_app.go doesn't need to change at all.

I had something similar in my mind, so I am actually happy with this approach.

> App1 exists, App2 exists
`kapp delete -a App1 --prev-app App2`
What would you expect the behaviour to me?

> Given that App2 should have been migrated and not exist at the same time, I expect App1 to be deleted and App2 to be ignored.

I totally agree. We should just ignore App2 in this case as this scenario shouldn't have happened.

pkg/kapp/cmd/app/migration_flags.go
cppforlife: the reason why i was proposing --prev-app as the flag name is because it's more consistent with the naming of --app.

```
  -a, --app string            Set app name (or label selector) (format: name, label:key=val, !key)
```

pkg/kapp/cmd/app/migration_flags.go
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package app
+
+import (
+	"github.com/spf13/cobra"
+)
+
+type MigrationFlags struct {
+	prevAppName string
```
cppforlife: should make this field public since the struct is public on the command.

pkg/kapp/cmd/app/migration_flags.go
praveenrewar: Ah, I didn't think of it this way. 
Noice!

pkg/kapp/cmd/app/prev_app_flag.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2022 VMware, Inc.
```
neil-hickey, line 1: I'm not 100% on this name, I wanted to keep it separate from `AppFlags` because that's used more widely than where this feature is useful. But I struggled to come up with something better.

neil-hickey, line 1: I'm not 100% on this name, I wanted to keep it separate from `AppFlags` because that's used more widely than where this feature is useful. But I struggled to come up with something better.

praveenrewar: It looks good to me.

test/e2e/create_update_delete_test.go
```diff
@@ -143,8 +143,7 @@ data:
 	appName := "test-create-update-delete-prev-app"
 	prevAppName := "test-create-update-delete-prev-app-old"
 	cleanUp := func() {
-		kapp.Run([]string{"delete", "-a", appName})
-		kapp.Run([]string{"delete", "-a", prevAppName})
+		kapp.Run([]string{"delete", "-a", appName, "--prev-app", prevAppName})
```
praveenrewar: I may have missed some tests, but I think this will always delete `appName` as in all the tests we are using the `--prev-app` flag with deploy which will migrate the app to `appName`. 


Other than this, everything else looks good to me.

neil-hickey: Cool, I added some explicit tests and removed them from the cleanup.

praveenrewar (approved):
LGTM!

100mik (approved):
LGTM!

sethiyash (approved):
Looks Good.

renuy:
Available as part v0.48.0

PR #517, opened 2022-05-29
Strip kustomize-style name hash suffix

### LINKED ISSUE #81: Detect ConfigMap change from kustomize smartly
Kustomize will generate a ConfigMap suffix automatically, which makes the change like delete and create.

```
Namespace        Name               Kind        Conds.  Age  Op      Wait to    Rs  Ri
something-test   conf-d-7g4t7m57mg  ConfigMap   -       9m   delete  delete     ok  -
^                conf-d-dc5dg9m84f  ConfigMap   -       -    create  reconcile  -   -
```

Is it possible to smartly detect it as a resource change to highlight the diff?

criztovyl:

This PR tries to demonstrate a way to implement the feature request from #81.

Fixes #81

```release-note
Adds an `StripNameHashSuffix` option to config, allowing to strip kustomize-style ConfigMap/Secret hash suffixes.
```
##### Review Checklist:

- [x] Relevant docs in this repo added or updated (none found)
  a link to that PR (will do that after first feedback here)
  change

```docs

```


criztovyl:
Some thoughts on where to put the activation flag:

* the suffix-strip activation should got to config, like `templateRules`
* these are passed to `VersionedResource` (versioned resource type) from `ChangeSetWithVersionedRes` (change-set type)
* the change-set type needs to be adjusted  to pass the activation from config to the versioned resource
* the change-set type does not know about config, it only gets the templateRules directly.
  thus the places using the change-set will need be adjusted to pass the activation.
  as the demonstration tests is on change-set level I'll not care about this yet.

While looking how to change the change-set type, it came to my understanding that the versioned resource type seems to be used only in change-set type, for the following two things:
1. working with the name / resource itself (`BaseNameAndVersion()` etc.)
2. updating the name on affected resources (`UpdateAffected()`)

First, I'll need to adjust the places where it's used for 1.
There it seems the change-set type instantiates the versioned-resource type ad-hoc/in-line, with the templateRules `nil`ed-out.
It only instantiates with rules when actually calling the update for affected resources. This suggests the `templateRules` are only required when updating affected resources (i.e. 2.).

On the other hand the suffix-stripping flag is required in all the places the versioned-resource is currently instantiated ad-hoc.
This means I need to change all those places. I'll move forward with that for the moment.

I am considering to extract the ad-hoc instantiation into a method in the change-set type.
But generally I think it might be better to separate 1. from the above list into an own type, e.g. `VersionedResourceName`.
That being a more invasive change, is another reason to, for now, simply changing the ad-hoc instantiations.

criztovyl:
I suspect a simple flag will not be enough, might need to be a more complex type, as kustomize allows you to disable the name-suffix-hash, so you might need to exclude certain CMs/Secrets.

It also affects only CMs/Secrets, I think my current implementation will also strip non-suffixes from e.g. deployments like `name: my-deployment` (i.e. stripping `-deployment`).

criztovyl:
Will look into some e2e tests now and await initial feedback. :)

100mik:
Hey! Thanks for the PR <3
Going over this and taking a closer look 🚀

cppforlife:
> It also affects only CMs/Secrets, I think my current implementation will also strip non-suffixes from e.g. deployments like name: my-deployment (i.e. stripping -deployment).

this is my biggest concern as well. can we have some stricter way of determining what is "suffixed by kustomize" and what is not?

praveenrewar:
@criztovyl Thank you so much for the PR. 
I am pretty new to kustomize space, so just trying to understand a bit more about the suffixHash...

Does the suffix get added to only the ConfigMap and Secrets created using the kustomize generators? 
Is there any way to determine if a resource has a suffix added by kustomize (using some annotation, etc,.) which we can use to remove the suffix. 
Or is there a way to disable the suffix (I noticed this option provided [here](https://github.com/vmware-tanzu/carvel-kapp/issues/81#issuecomment-1092014577), but not sure if this will work for all the resources)

criztovyl:
Thanks for the feedback :)
I am working on this over the weekends, I'll answer your questions then.

ChristianCiach:
I was one of the first persons to upvote the issue https://github.com/vmware-tanzu/carvel-kapp/issues/81 many months ago and I have been thinking about this issue a lot since then.

I am sorry for being so negative, but I think this issue is close to impossible to solve. 

> Or is there a way to disable the suffix

Yes. You've already found `disableNameSuffixHash: true`. This is what we are using right now, but this has to be configured for each and every configMap/secret generator separately and is easily forgotten and also adds a lot of "noise" to the `kustomization.yaml` files. 

> Is there any way to determine if a resource has a suffix added by kustomize (using some annotation, etc,.) which we can use to remove the suffix.

No, there are no annotations or anything else to identify names of configMaps or secrets that have been rewritten by kustomize.

This is one of the main reason why I think this issue is next to impossible to solve with perfect accuracy. Kapp would not only have to remove the hash-suffixes from the configMaps and secrets, but also from all resources that are *referencing* these configMaps and secrets. Again, there a no annotations or anything to identify these references.

To make matters worse and which is probably the final nail in the coffin here: Similar to kapp, kustomize has a set of hardcoded [template rules](https://carvel.dev/kapp/docs/v0.48.0/config/#templaterules) which are called [name reference transformers](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/README.md#name-reference-transformer) in kustomize. Similar to kapp, users can provide additional transformers via `kustomizeconfig.yaml` files to enable kustomize to also rewrite configMap/secret references inside Custom Resources (CRs) that kustomize does not support out of the box. For example, we use this `kustomizeconfig.yaml` to let kustomize transform the name of a `Secret` reference inside a CR called [ServersTransport](https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#kind-serverstransport):

```yaml
nameReference:
  - kind: 'Secret'
    fieldSpecs:
      - kind: 'ServersTransport'
        path: 'spec/rootCAsSecrets'
```

If we want kapp to *strip* this hash-suffix from this CR, kapp needs to take this config into account to *know* about this reference:

```yaml
kind: Config
templateRules:
  - resourceMatchers:
      - apiVersionKindMatcher: {apiVersion: v1, kind: Secret}
    affectedResources:
      objectReferences:
        - path: [spec, rootCAsSecrets, {allIndexes: true}]
      resourceMatchers:
  - apiVersionKindMatcher: {apiVersion: traefik.containo.us/v1alpha1, kind: ServersTransport}
```

I don't know how I feel about that. These `templateRules` primarily exist so that kapp can **add** its own "version"-suffix to all resources that reference the given resource (a "Secret" in this example). But now we need these `templateRules` so that kapp can **remove** these suffixes that have been added by another tool. This feels... icky.

criztovyl:
Oh, true, I did not think about that consequence yet, thank you!

Actually my concern is just cosmetic, I dont necessarily want true versioning.
I am happy if the diff shows two hash-suffixed CMs/Secrets as an update, not as delete + add.

For that I can ignore the referneces, they will show as updated anyway.
So the stripping of the suffix should not go into final cluster state (as it would do with my current code), it should only be used to identifying potentially related resources.

I'll look into that.

criztovyl:
From my understanding the resources are diffed by grouping the old and new one using their name and then diffing these groups-of-two. Thus the challenge should be to "just" get the old and new hash-suffixed CMs/Secrets into the same group (i.e. update) instead of two separate one (delete+add).

The pieces are already there, used for actually versioned resources.
 
Spoken differently I want to build a `VersionedResource` that provides a `UniqVersionedKey` that strips the suffix, but where `SetBaseName` and `UpdateAffected` are no-op (return directly without updating resource name nor references).

I think that implementing this into the `VersionedResource` type is technically possible but will push it to it's edge, I intend to refactor that properly.

praveenrewar:
Thank you so much for sharing all the insights @ChristianCiach. 

> Kapp would not only have to remove the hash-suffixes from the configMaps and secrets, but also from all resources that are referencing these configMaps and secrets

This is definitely a valid concern and I can't think of an easy way around it (maybe [ytt](https://carvel.dev/ytt) would be helpful here). 

@criztovyl Thank you again for giving this a shot. I will try to take a closer look at all the details again in the morning.

cppforlife:
> So the stripping of the suffix should not go into final cluster state (as it would do with my current code), it should only be used to identifying potentially related resources.

as an interested observer (of this PR), this comment clears up some of my misunderstanding.

praveenrewar:
> This is what we are using right now, but this has to be configured for each and every configMap/secret generator separately and is easily forgotten and also adds a lot of "noise" to the kustomization.yaml files.

I have been thinking about this and I tried to use ytt overlays to add the versioned annotation and the disableNameSuffixHash option to all the secret generators and configmap generators. But I noticed that `kustomize build` command only accepts paths to directories and hence it's not possible to pipe the ytt output to kustomize. We would have to update the kustomization.yaml first and then use kustomize, so I guess it's not a viable option either.

pkg/kapp/diff/versioned_resource.go
```diff
@@ -34,6 +40,12 @@ func (d VersionedResource) BaseNameAndVersion() (string, string) {
 	if len(pieces) > 1 {
 		return strings.Join(pieces[0:len(pieces)-1], nameSuffixSep), pieces[len(pieces)-1]
 	}
+	if d.StripNameHashSuffix() {
```
praveenrewar: Since the hash is generated for ConfigMaps and Secrets only, should we limit this to only these 2 kinds?

Also, will it be possible to generate a hash with the same logic that kustomize uses and use that to make sure that we are removing a hash generated by kustomize (it seems they are using sha256 to generate a hash from the resource content). Or to keep it more simple maybe just check the length of the suffix and check if it's alphanumeric?
(The idea is to minimise the side effects as much as possible)

ChristianCiach:
> So the stripping of the suffix should not go into final cluster state (as it would do with my current code), it should only be used to identifying potentially related resources.

I see. So you want this change to be purely cosmetic, just to show an `update` operation on deploy instead of a `delete` and `create`. 

Yes, this sounds a lot less invasive, since there is no need to rename any references in this case. 

Just one more thing to keep in mind: It is unlikely, but when stripping (for example) `dashboard-config-7bmh4kf599` to `dashboard-config`, it could conflict with a configMap that is *actually* named `dashboard-config`. 

I still think this is nice to have, but keeping in mind that this is a purely cosmetic change, I still think it may not be worth it, depending on code complexity and future maintenance overhead.

pkg/kapp/diff/versioned_resource.go
criztovyl: Yes, I intend to add a (customizeable) filter for CMs and Secrets. :)

My intention is to "hide" stripping the suffix behind a flag (in Config YAML), targeted at kustomize users. I inted that the customizeable filter above allows exluding resources by resourceMatchers. Hiding it behind a flag means users will need to opt-in, where they then should be aware of the potential side effects and how to fix the unwanted ones via config. That would be my take on the latter. But the code and demo (e2e) tests I intend to write should be better at explaining that :)

criztovyl:
> I still think it may not be worth it, depending on code complexity and future maintenance overhead.

Yes. I am doing this also for the fun of it -- should it get to complex I have no problem to accept the implementation does not fit :)

pkg/kapp/diff/versioned_resource.go
ChristianCiach: @praveenrewar While it is true that kustomize only creates hash-suffixes for ConfigMaps and Secrets when only using the [built-in plugins](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin), kustomize can be [extended with custom plugins](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) that may also use the built-in [hash-suffix-transformer](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/hashtransformer/HashTransformer.go).

In practice I've never seen anyone actually doing that, but I wanted to add this information for the sake of completeness.

praveenrewar: > Yes, I intend to add a (customizeable) filter for CMs and Secrets. :)

Thanks you :)

> Hiding it behind a flag means users will need to opt-in, where they then should be aware of the potential side effects and how to fix the unwanted ones via config

True. Just mentioned thinking that it would be great to have less side effects to begin with :)

praveenrewar: > While it is true that kustomize only creates hash-suffixes for ConfigMaps and Secrets when only using the [built-in plugins](https://github.com/kubernetes-sigs/kustomize/tree/master/plugin), kustomize can be [extended with custom plugins](https://kubectl.docs.kubernetes.io/guides/extending_kustomize/) that may also use the built-in [hash-suffix-transformer](https://github.com/kubernetes-sigs/kustomize/blob/master/plugin/builtin/hashtransformer/HashTransformer.go).

Ah, I see. So I guess we won't be able to have that check then 😞

criztovyl:
Hi o/

I have implemented that the diff ignores the nameHashSuffix. The new `TestChangeSet_StripKustomizeSuffix` unit test should demonstrate what I want(ed):
https://github.com/vmware-tanzu/carvel-kapp/blob/7973475f576cf399e7f7f4b67398360767b130b7/pkg/kapp/diff/change_set_with_versioned_rs_test.go#L187-L196

for me the following points are still open:

- [x] refactor versioned resource type
      I would like to split it into two types, one for name-matching related stuff and one for name-and-reference-updating stuff.
      cf https://github.com/vmware-tanzu/carvel-kapp/pull/517#issuecomment-1140425656)
- [x] customizable filter for which resources the suffix should not be trimmed or should be trimmed additonally
      I would like to implement this using resourceMatchers in (default) config.
- [x] add e2e test for my use-case

Locally all unit tests succeed, e2e also look good.

GitHub tells me the Action workflows need approval from a maintainer, maybe those could also be activated?
Then I don't run into funny works-on-my-machine stuff :)

For the VersioneResource refactoring I am considering extracting that into a dedicated PR and (re)basing the diff changes on that later. What you you think?

pkg/kapp/diff/change_set_with_versioned_rs_test.go
```diff
@@ -157,6 +153,96 @@ metadata:
...
+  foo: bar
+`))
+
+	stripNameHashSuffix := true
+	changeSetWithVerRes := NewChangeSetWithVersionedRs([]ctlres.Resource{existingRs}, []ctlres.Resource{newRs}, nil, ChangeSetOpts{}, ChangeFactory{}, stripNameHashSuffix)
+
+	changes, err := changeSetWithVerRes.Calculate()
+	require.NoError(t, err)
+
+	//require.Equal(t, ChangeOpUpdate, changes[0].Op(), "Expect to get updated") // TODO: check why this is add, not update
```
praveenrewar: > // TODO: check why this is add, not update

It's `add` because a new resource is _created_ on the cluster for each new version of a versioned resource. The diff just shows the difference between the previous version and the current version.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,19 +77,24 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
 	return allChanges, nil
 }
 
+func (d ChangeSetWithVersionedRs) versionedResourceName(res ctlres.Resource) VersionedResource {
```
praveenrewar: versionedResourceName  -> versionedResource ?

praveenrewar:
> customizable filter for which resources the suffix should not be trimmed or should be trimmed additonally
I would like to implement this using resourceMatchers in (default) config.

Looking forward to seeing this :)

> For the VersioneResource refactoring I am considering extracting that into a dedicated PR and (re)basing the diff changes on that later

Sure, we can have a separate PR. (The changes are still reviewable in the same PR so I am actually comfortable with both the options)

pkg/kapp/diff/change_set_with_versioned_rs.go
criztovyl: seems i forgot a TODO comment here. ^^

this is my "splitting the overloaded VersionedResource type"  refactoring-quickfix.  the "name" suffix should qualify that it can only be use for name-related stuff, but not for reference-update stuff; for the latter the templateRules are missing.

later this will return an instance of the type dedicated to the name-related logic, properly renamed.

criztovyl:
finally took the time to start thinking about e2e test and directly uncovered a bug in the new code.

while the diff is correctly shown, the old CM is not deleted.

```
christoph@chschulz09:~/Documents/dev/carvel-kapp/test/e2e/res/kustomize/overlays/versioned2                                                                   
[criztovyl/strip-name-hash-suffix %= 173c57b]$ ../../../../../../kapp deploy -a versioned -f kapp.yml -c                                                                                                                                                                                                                     
Target cluster 'https://127.0.0.1:37067' (nodes: minikube)                                                                                                    
                                                                                                                                                              
@@ create configmap/config-map-798k5k7g9f (v1) namespace: default @@           
  ...                                                                                                                                                         
  1,  1   data:                                                                                                                                               
  2     -   foo: foo                                                                                                                                                                                                                                                                                                         
      2 +   foo: bar                                                                                                                                                                                                                                                                                                         
  3,  3   kind: ConfigMap                                                                                                                                     
  4,  4   metadata:                                                                                                                                           
  ...                                                                                                                                                         
  7,  7       kapp.k14s.io/app: "1656313846367027698"                                                                                                         
  8     -     kapp.k14s.io/association: v1.dacacaa0f2d3ae85e5dc5c636cc2c35e                                                                                                                                                                                                                                                  
      8 +     kapp.k14s.io/association: v1.ba177a572163dd4bcbee99aee2c7a5a6                                                                                                                                                                                                                                                  
  9,  9     managedFields:                                                                                                                                    
 10, 10     - apiVersion: v1                                                                                                                                  
                                                                                                                                                              
Changes                                                                                                                                                       
                                                                                                                                                              
Namespace  Name                   Kind       Age  Op      Op st.  Wait to    Rs  Ri                                                                           
default    config-map-798k5k7g9f  ConfigMap  -    create  -       reconcile  -   -                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                             
Op:      1 create, 0 delete, 0 update, 0 noop, 0 exists                                                                                                                                                                                                                                                                      
Wait to: 1 reconcile, 0 delete, 0 noop                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                             
Continue? [yN]: y                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                             
9:12:19AM: ---- applying 1 changes [0/1 done] ----                                                                                                            
9:12:20AM: create configmap/config-map-798k5k7g9f (v1) namespace: default                                                                                     
9:12:20AM: ---- waiting on 1 changes [0/1 done] ----                                                                                                          
9:12:20AM: ok: reconcile configmap/config-map-798k5k7g9f (v1) namespace: default                                                                              
9:12:20AM: ---- applying complete [1/1 done] ----                                                                                                             
9:12:20AM: ---- waiting complete [1/1 done] ----                                                                                                              
                                                                               
Succeeded

christoph@chschulz09:~/Documents/dev/carvel-kapp/test/e2e/res/kustomize
[criztovyl/strip-name-hash-suffix= c0fd89c]$ kubectl get cm
NAME                     DATA   AGE
config-map-798k5k7g9f    1      7m39s
config-map-7tgk8db28b    1      9m3s
[...]
```

I'll have a look at that.
In the output above kapp tells `1 create, 0 delete [...]` I'll start with checking how kapp marks objects as deleted. :)

praveenrewar:
> while the diff is correctly shown, the old CM is not deleted.

Generally, for versioned resources, kapp keeps upto 5 versions (customisable using `kapp.k14s.io/num-versions=int`) and `noop` operation is used for the previous versions. I will have to take a look at the changes again to be sure, but I am guessing that something similar is happening here.

Also, I am wondering why the association label is getting changed.

P.S. There are some tests that are failing on the latest version of k8s (`test-gh`), so you might wanna rebase with develop once to get the latest changes which have a fix for them.

criztovyl:
> I am wondering why the association label is getting changed.

I checked this and found that the association label is created on the *input* name of the CM.
This means if you have an actual versioned resource, the name in input always is the same, the `-ver-{n}` is added later, with the association label already added.
But with the name-suffix-hash the name in input changes, thus also the association label does.

> Generally, for versioned resources, kapp keeps upto 5 versions (customisable using `kapp.k14s.io/num-versions=int`) and `noop` operation is used for the previous versions. I will have to take a look at the changes again to be sure, but I am guessing that something similar is happening here.

I have checked this with `num-versions=1` and a third change -- then the first CM is deleted.
But I suspect this only works with 1 version to keep; with more the only information kapp has for determining the order of name-suffix-strip CMs is the age.

praveenrewar:
> But with the name-suffix-hash the name in input changes, thus also the association label does.

I see. That makes sense.

> But I suspect this only works with 1 version to keep; with more the only information kapp has for determining the order of name-suffix-strip CMs is the age.

That's true. With the current changes I think it would be randomly deleting any of the changes. But do you need to keep more than 1 versions for your use case?

criztovyl:
> But do you need to keep more than 1 versions for your use case?

Actually, no, I thought about dropping any other version for suffix-stripped resources :)

criztovyl:
kapp now behaves like I want it; the diff is shown and the old CM is deleted. \o/

https://gist.github.com/criztovyl/85f594d455a2a413d6e20b0766303033

e2e seem to be running fine locally, there were some failures regarding wait rules, but recreating the `kapp-test` NS fixed those.

I have not addressed the changing association label yet, but I'll put that at the end of the backlog ^^

Next I'll check how to add some end-to-end tests, or maybe first some more unit tests. If I have that, I'll look into the `VersionedResource` refactoring.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -101,28 +107,29 @@ func (d ChangeSetWithVersionedRs) assignNewNames(
...
-			newVerRes.SetBaseName(1)
+			newVerRes.AssignNewVersion()
 		}
 	}
 }
 
 func (d ChangeSetWithVersionedRs) addAndKeepChanges(
 	newRs versionedResources, existingRsGrouped map[string][]ctlres.Resource) (
-	[]Change, map[string]ctlres.Resource, error) {
+	[]Change, map[string]Change, error) {
```
criztovyl: see comment at `numOfResourcesToKeep` why this changed

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -238,7 +248,14 @@ func (ChangeSetWithVersionedRs) numOfResourcesToKeep(res ctlres.Resource) (int,
 		if numToKeep < 1 {
 			return 0, fmt.Errorf("Expected annotation '%s' value to be a >= 1", versionedResNumVersAnnKey)
 		}
+	} else if versionedRes.IsVersioned() && !versionedRes.IsTracked() {
+		if change.Op() == ChangeOpKeep {
+			numToKeep = 1
+		} else {
+			numToKeep = 0
+		}
```
criztovyl: it's tricky to explain why this is necessary, i might need to think about adding a comment to the code why this necessary.

but for now I'll add it here:

Generally, for name-hash-suffix CMs it's useless to keep old CMs, as we cannot properly order them there is no "last-X-versions" semantic. Thus, we always want to delete (all) existing/old ones. But we're facing a problem here when the CM does not change -- then there is no change that creates the CM but this method would still lead to the not-actually-old existing CM to get deleted.

Using the change to determine if the existing CM is old or, well, existing, I added to pass down the change that "produces" this CM -- if it's a "keep" change, it's an unchanged CM we want to keep, otherwise an old one we want to delete.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -225,9 +234,10 @@ func (d ChangeSetWithVersionedRs) newNoopChange(existingRes ctlres.Resource) Cha
 	return NewChangePrecalculated(existingRes, nil, nil, ChangeOpNoop, nil, OpsDiff{})
 }
 
-func (ChangeSetWithVersionedRs) numOfResourcesToKeep(res ctlres.Resource) (int, error) {
-	// TODO get rid of arbitrary cut off
-	numToKeep := 5
+func (d ChangeSetWithVersionedRs) numOfResourcesToKeep(change Change) (int, error) {
+	res := change.NewOrExistingResource()
```
criztovyl: I am not a 100% sure about the side-effects of this, but this should only play a role for "actually" versioned resources, as only those should get `keep` Changes that have no "new" resource.

That actually might need further investigation as the `numOfResourcesToKeep` will now work on `existingRes` for `keep` changes, not `newRes` anymore ... :thinking:

pkg/kapp/diff/versioned_resource.go
```diff
@@ -56,7 +80,55 @@ func (d VersionedResource) UniqVersionedKey() ctlres.UniqueResourceKey {
 	return ctlres.NewUniqueResourceKeyWithCustomName(d.res, baseName)
 }
 
+func (d VersionedResource) trackVersions() bool {
```
criztovyl: not quite happy with the name, will address that in refactoring.

pkg/kapp/diff/change_set_with_versioned_rs.go
praveenrewar: > it's tricky to explain why this is necessary

I actually was thinking of something similar when the we were discussing about this the last time, but I do think that there might be a better way of doing this... I will give it some more thought and let you know if I come up with something.

pkg/kapp/diff/versioned_resource.go
```diff
@@ -56,7 +80,55 @@ func (d VersionedResource) UniqVersionedKey() ctlres.UniqueResourceKey {
...
+	_, hasVersionedAnn := d.res.Annotations()[versionedResAnnKey]
+	return hasVersionedAnn
+}
+
+func (d VersionedResource) IsVersioned() bool {
```
praveenrewar: It seems a bit weird to have a method called `IsVersioned` on a `VersionedResource`. I think you are planning to refactor to have a new type for these resources but for now does it make sense to use `hasVersionedAnn()` `||` something like `stripEnabled(res, d.stripNameHasSuffix)` wherever needed?

pkg/kapp/diff/versioned_resource.go
praveenrewar: I think just `hasVersionedAnn()` makes sense. But ideally, resources that have this annotation should be created as `VersionedResources{}`.

pkg/kapp/diff/versioned_resource.go
```diff
@@ -56,7 +80,55 @@ func (d VersionedResource) UniqVersionedKey() ctlres.UniqueResourceKey {
...
+	if d.StripNameHashSuffix() {
+		return true
+	}
+	return d.trackVersions()
+}
+
+func (d VersionedResource) IsTracked() bool {
```
praveenrewar: Same as above. Does it makes to use `hasVersionedAnn()` `&&` something like `stripEnabled(res, d.stripNameHasSuffix)` wherever needed?

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -276,18 +292,10 @@ func newVersionedResources(rs []ctlres.Resource) versionedResources {
...
-		// Expect that versioned resources should not be transient
-		// (Annotations may have been copied from versioned resources
-		// onto transient resources for non-versioning related purposes).
-		_, hasVersionedAnn := res.Annotations()[versionedResAnnKey]
-
-		versionedRs := VersionedResource{res: res}
-		_, version := versionedRs.BaseNameAndVersion()
-
-		if hasVersionedAnn && !res.Transient() && version != "" {
+		if d.versionedResourceName(res).IsExistingVersioned() {
```
praveenrewar: Something like this might help in reducing the noise?
```suggestion
		if (hasVersionedAnn && !res.Transient() && version != "") || d.stripEnabled(res) {
```

praveenrewar (commented):
Hey @criztovyl !!
I think you are planning to do some refactoring later on to have a new type (similar to VersionedResources) for the kustomize resources with strip enabled, but for now do you think it makes sense to keep all (or most of) that logic in ChangeSetWithVersionedRs? This will help in reducing some noise in the current PR and also the refactoring will be easy. 
I have added a couple of comments on similar lines.

(Totally fine if you are planning to do the refactoring as part of the same PR)

criztovyl:
Finally took the time to write the e2e test, based on them I can now confidently make your suggested changes and/or do the refactorings. :)

criztovyl:
I have changed the e2e test as it failed on CI because the diff looks different. I dropped checking the diff now and only check there is one "update" diff and not two "delete and create" diffs.

I also made configurable which resources are covered by the name-hash-suffix-stripping, thus you can now exclude CMs or other resources which do not have a suffix (e.g. because it was disabled in kustomize or does not come from and configMapGenerator).

I plan next to restructure the commits so I have a base with failing tests on top of which I can maybe try different approaches to actually stripping the resources. I'll also look through the review comments, did not do that yet. :)

praveenrewar:
> I plan next to restructure the commits so I have a base with failing tests on top of which I can maybe try different approaches to actually stripping the resources. I'll also look through the review comments, did not do that yet. :)

Awesome, do let me know when you would want me to take another look at the PR :)

criztovyl:
I hope the tests are fixed now, I will focus on properly refactoring my changes now.

criztovyl:
@praveenrewar Could you review how I have implemented the configuration for my feature?

The config is used in the following place, added with commit https://github.com/vmware-tanzu/carvel-kapp/pull/517/commits/9807418550ed0827f69118fb131da5d7fe8973cb:

https://github.com/vmware-tanzu/carvel-kapp/blob/9807418550ed0827f69118fb131da5d7fe8973cb/pkg/kapp/diff/versioned_resource.go#L27-L29

praveenrewar:
> Could you review how I have implemented the configuration for my feature?

I really like that we have resource matchers now in the StripNameHashSuffixConfig. 
What is your opinion on removing the `enabled` boolean now that we have the resource matchers present? I think that the mere presence or absence of a resource matcher can be used to decide if the suffix needs to be trimmed for a resource or not (So we won't have to add a disabled config in the default config).

criztovyl:
> What is your opinion on removing the `enabled` boolean now that we have the resource matchers present? I think that the mere presence or absence of a resource matcher can be used to decide if the suffix needs to be trimmed for a resource or not (So we won't have to add a disabled config in the default config).

The idea behind the boolean flag is that it allows for kustomize users to simply set that flag, without needing to configure resource matchers.

praveenrewar:
> The idea behind the boolean flag is that it allows for kustomize users to simply set that flag, without needing to configure resource matchers.

I see, I just feel that having able to add the resource matchers would give more control to users. For example, right now a user might get confused if they would want to enable the stripping just for Secrets or just for ConfigMaps and not for both.

criztovyl:
True, I'll check converting my flag to `EnableKustomizeMatchers` and based on that add the kustomize resource matchers :)

praveenrewar:
> I'll check converting my flag to EnableKustomizeMatchers and based on that add the kustomize resource matchers :)

Hehe, I still feel that the boolean is a bit redundant, like just removing a particular resource matcher would be equivalent to keeping the boolean value false. But if this is more convenient for you then we can keep it :)

criztovyl:
I started my refactoring by making the `Versioned` of the `versionedResources` type actual `[]VersionedResource`.

For most of the code that simply means changing the signatures and occasionally throwing in a `.res` or replacing ad-hoc VRes instantiations. But it was also necessary to make some more intrusive changes:
* for `diff.GroupResources` I turned it into an `GroupVersionedResources`, it does not seem to be used anywhere else.
* for `VersionedResource#UpdateAffected` we're running into a problem: this method needs to be called for both versioned and non-versioned resoures, but the lists in `versionedResources` now have different types. As a workaround I simply convert the list ad-hoc.

<s>But both places strictly do not need the "Versioned" part of the VersionedResource, as such I am considering to introduce a new interface, a `ResourceHolder`, which turns `.res`-field-access into an actual `Res()` method:

```go
// resource_holder.go (new)

type ResourceHolder interface {
    Res() ctlres.Resource
}

type ResourceHolderImpl struct {
    res ctlres.Resource
}

func (d ResourceHolderImpl) Res() ctlres.Resource {
    return d.res
}

// versioned_resource.go

func (d VersionedResource) Res() ctlres.Resource {
    return d.res
}
```

What do you think about such an `ResourceHolder` interface?</s>
Edit: go does not allow easy conversion between slices of "similar" types.

criztovyl:
Ah, forgot to look into the `Enabled`/`EnableKustomizeResourceMatchers`.

praveenrewar:
Hey @criztovyl ! Sorry I didn't get a chance to look at the changes properly, will get it to it soon :)

criztovyl:
ok, no worries :)

pkg/kapp/config/conf.go
```diff
@@ -178,3 +178,21 @@ func (c Conf) ChangeRuleBindings() []ChangeRuleBinding {
 	}
 	return result
 }
+
+type StripNameHashSuffixConfigs []StripNameHashSuffixConfig
+
+func (c StripNameHashSuffixConfigs) AggregateToCtlRes() (enabled bool, resourceMatchers [][]ctlres.ResourceMatcher) {
```
praveenrewar: We typically don't use named result parameters and naked returns unless we have a strong reason to do so.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -254,44 +270,87 @@ func (d ChangeSetWithVersionedRs) newChange(existingRes, newRes ctlres.Resource)
...
-func existingVersionedResources(rs []ctlres.Resource) versionedResources {
+func (d ChangeSetWithVersionedRs) existingVersionedResources() versionedResources {
 	var result versionedResources
-	for _, res := range rs {
-		// Expect that versioned resources should not be transient
-		// (Annotations may have been copied from versioned resources
-		// onto transient resources for non-versioning related purposes).
-		_, hasVersionedAnn := res.Annotations()[versionedResAnnKey]
+	for _, res := range d.existingRs {
+		result.appendRes(d.VersionedFromExistingResource(res))
```
praveenrewar: I think we can keep the logic here itself instead of a separate method. Also, using a public method inside a private method _generally_ is not a good idea. It usually leads to some issues (not in every case though).

praveenrewar (commented):
> For most of the code that simply means changing the signatures and occasionally throwing in a .res or replacing ad-hoc VRes instantiations.

I am still thinking about this. (I do like the idea, but `.res` is something that is making me think more about it.)

criztovyl:
> the mere presence or absence of a resource matcher can be used to decide if the suffix needs to be trimmed for a resource or not

I have now implemented it that way.

Completely neglected docs until now, will look into that next. :)

criztovyl:
While looking into the docs I realized the name of this feature, which was not very good in the beginning, after the iterations this PR went through, went bad...

The central part of this PR now is the alternative implementation of `VersionResource` (now an interface) for resources with hash suffixes, `HashSuffixResource`.

As such, what do you think about renaming everything to that? Maybe you have a better name in mind?

|||
|-|--|
|`HashSuffixResource`|`HashSuffixResource` (same)|
|`stripnamehashsuffix{,_test}.go`|`hashsuffix_resource{,_test}.go`|
|`conf.StripHashNameSuffixConfig`|`conf.HashSuffixDiffConfig` (`conf.HashSuffixDiffRules`?)|
|`ctldiff.stripNameHashSuffixConfig`|`ctldiff.hashSuffixDiffConfig`(`ctldiff.hashSuffixDiffRules`?)|
|`ctlconf.StripHashNameSuffixConfig(s)`|`ctlconf.HashSuffixDiffConfig(s)` (`ctlconf.HashSuffixDiffRules(List)`?)|

praveenrewar:
> As such, what do you think about renaming everything to that?

Renaming definitely makes sense to me, but I will have to take a closer look at the changes (planning to do that tomorrow) to be able to comment on all the names.

pkg/kapp/config/default.go
```diff
@@ -643,6 +643,8 @@ changeRuleBindings:
               - anyMatcher: {matchers: *rbacMatchers}
               - anyMatcher: {matchers: *podRelatedMatchers}
       - hasNamespaceMatcher: {}
+
+StripNameHashSuffix: {}
```
praveenrewar: `StripNameHashSuffix` -> `nameHashSuffix` OR `stripNameHashSuffix`
Also, I think that we should skip adding an empty config to default config.

pkg/kapp/config/config.go
```diff
@@ -38,6 +38,8 @@ type Config struct {
 	// TODO validations
 	ChangeGroupBindings []ChangeGroupBinding
 	ChangeRuleBindings  []ChangeRuleBinding
+
+	StripNameHashSuffixConfig StripNameHashSuffixConfig
```
praveenrewar: Should probably be a slice?

pkg/kapp/config/conf.go
```diff
@@ -178,3 +178,21 @@ func (c Conf) ChangeRuleBindings() []ChangeRuleBinding {
 	}
 	return result
 }
+
+type StripNameHashSuffixConfigs []StripNameHashSuffixConfig
+
+func (c StripNameHashSuffixConfigs) AggregateToCtlRes() [][]ctlres.ResourceMatcher {
```
praveenrewar: What do you think about adding a `ResourceMatcher` method to `config.go`? Something like [this](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/config/config.go#L312) (this also uses the AnyMatcher which I hope would reduce some additional complexity in `stripnamehassuffix.go`)

I am assuming that we would want users to be able to use something like below to be able to strip the suffix for configmaps and secrets.
```
stripNameHashSuffix:
- resourceMatchers:
  - apiVersionKindMatcher: {apiVersion: v1, kind: ConfigMap}
  - apiVersionKindMatcher: {apiVersion: v1, kind: Secret}
```

pkg/kapp/config/conf.go
```diff
@@ -178,3 +178,21 @@ func (c Conf) ChangeRuleBindings() []ChangeRuleBinding {
...
+	var resourceMatchers [][]ctlres.ResourceMatcher
+	for _, conf := range c {
+		resourceMatchers = append(resourceMatchers, ResourceMatchers(conf.ResourceMatchers).AsResourceMatchers())
+	}
+	return resourceMatchers
+}
+
+func (c Conf) StripNameHashSuffixConfigs() StripNameHashSuffixConfigs {
```
praveenrewar: nit: We can just use []StripNameHashSuffixConfig as return type and directly return the configs.

pkg/kapp/diff/stripnamehashsuffix.go
```diff
@@ -0,0 +1,83 @@
...
+
+	var allMatchers []ctlres.ResourceMatcher
+	for _, matchers := range resourceMatchers {
+		// configurations that do not specify any matchers (like default
+		// config) have nil, which would lead AndMatcher to never match.
+		// as all the configs AndMatchers are ANDed, too, this would prevent
+		// matchers from any configuration to match, which would strip the
+		// whole suffix-strip configuration of any usefulness; as such we
+		// exclude nil matchers. :)
+		if matchers != nil {
```
praveenrewar: I think we can skip this. I do not think that anyone would want to add an empty config to the kapp config

pkg/kapp/diff/stripnamehashsuffix_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	ctlres "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/resources"
+)
+
+func TestStripNameHashSuffix_TestConfig_IncludeExcludeAccrossConfigs(t *testing.T) {
+	requireStripNameHashSuffixMatches(t, [][]ctlres.ResourceMatcher{
```
praveenrewar: Do you think something like table tests would be more intuitive over here?

pkg/kapp/diff/stripnamehashsuffix.go
```diff
@@ -0,0 +1,83 @@
...
+			allMatchers = append(allMatchers, ctlres.AndMatcher{
+				Matchers: matchers,
+			})
+		}
+	}
+	// N.B. if no config specifies any matchers (the default), then
+	// allMatchers will stay uninitialized/nil and the AndMatcher will never
+	// match, effectively disabling suffix strip.
+	result.ResourceMatcher = ctlres.AndMatcher{Matchers: allMatchers}
```
praveenrewar: I am still trying to understand why this is required. Would you be able to share a couple of scenarios that you are trying to cover, maybe we can find an alternative to it.

praveenrewar (commented):
Apologies for not being able to review this completely yet. I am still going through it, but thought of hitting the submit button once.

pkg/kapp/config/conf.go
criztovyl: yes, the dedicated type is only required for the `AggregateToCtlRes()`.
(Unless `func (cs []StripNameHashSuffixConfig) ...` is valid, I assumed it does not work, but did not check, though.)

pkg/kapp/diff/stripnamehashsuffix_test.go
criztovyl: I'll have a look how to do table tests with go test and if they are more expressive. :)

pkg/kapp/diff/stripnamehashsuffix.go
criztovyl: I was trying to allow include / exclude semantics.

Rethinking that it probably is better to have explicit fields for include and exclude matchers:
```yaml
stripNameHashSuffix: # hashSuffixDiffConfig
  include:
  - apiVersionKindMatcher: {apiVersion: v1, kind: ConfigMap}
  - apiVersionKindMatcher: {apiVersion: v1, kind: Secret}
  exclude:
  - kindNamespaceNameMatcher: {kind: ConfigMap, name: foo},
```
for the sake of discussion ([RE](https://github.com/vmware-tanzu/carvel-kapp/pull/517/files#r991539968)) I intentionally chose to not make the config a slice here:
in the end the include/exclude lists will each be merged into a single include/exclude list; there I assume that a list of include+exclude objects does not really add anything besides a layer of merging.
It might make sense in context of consistency, though, as many of the other config fields (template rules, etc) are also lists.

pkg/kapp/config/conf.go
criztovyl: I'll consider that.

pkg/kapp/config/config.go
criztovyl: See the reply regarding why the AndMatcher is required below.

pkg/kapp/diff/stripnamehashsuffix.go
criztovyl: > I do not think that anyone would want to add an empty config to the kapp config

yes, but currently (subject to change) there is any empty config in the default kapp config.

criztovyl:
#571 introduced a conflict on `func (d ChangeSetWithVersionedRs) groupResources([]ctlres.Resource)`:
it moved it to `func newGroupedVersionedResources(...)` and uses it via `diff.existingResourcesMap`.
But I changed the method to use (and return) `[]VersionedResource`, instead of instantiating `VersionedResource` ad-hoc.

To resolve that conflict I think I need to implement the suggestion from https://github.com/vmware-tanzu/carvel-kapp/pull/571#discussion_r989791601, extracting `existingVersionedResources` and `newVersionedResources` to a new struct.

I am thinking about something like below, what do you think?

```go
// cmd/app/deploy.go
// func (o *DeployOptions) calculateAndPresentChanges(...)
//...
diffResources := ctldiff.NewDiffResourceFactory(existingResource, newResources,
    conf.TemplateRules(), conf.StripNameHashSuffixConfigs()).NewDiffResources()

err := ctldiff.NewRenewableResource(diffResources).Prepare()
// ...

changes, err := ctldiff.NewChangeSetWithVersionedRs(diffResources,
    opts ChangeSetOps, changeFactory ChangeFactory).Calculate()
// ...
```

with:

```go
package diff

type DiffResources struct { // or any other name, only was the first one that came to my mind ^^
    ExisitingResources, NewResources versionedResources
    ExistingResourcesGrouped map[string][]VersionedResource
    // NB: While RenewableResources wants a map[string]ctlres.Resource
    // it should also be able to use [-1]VersionedResource.Res().
}

type DiffResourcesFactory struct { // or VersionedResourcesFactory?
    // attributes extracted from ChangeSetWithVersionedRs, only used by the extracted methods (only).
    existingRs, newRs []ctlres.Resource
    rules []ctlconf.TemplateRules
    stripNameHashSuffixConfig stripNameHashSuffixConfig
    // extracted attributes end
}

func NewDiffResourcesFactory(exisitingRs, newRs []ctlres.Resource,
    rules []ctlconf.TemplateRule,
    stripNameHashSuffixConfigs []ctlconf.StripNameHashSuffixConfig) DiffResourcesFactory {

    return DiffResourcesFactory{existingRs, newRs,
        rules, stripNameHashSuffixConfigFromConf(stripNameHashSuffixConfigs)}
}

func (d DiffResourcesFactory) NewDiffResources() DiffResources {
    result := DiffResources{}
    result.ExistingResources = d.existingVersionedRs()
    result.ExistingResourcesGrouped = d.existingVersionedRsGrouped()
    result.NewResources = d.newVersionedRs()
    return result
}

// methods extracted from ChangeSetWithVersionedRs

func (d DiffResourcesFactory) existingVersionedRs() versionedResources {
    // existingVersionedResources
}

func (d DiffResourcesFactory) existingVersionedRsGrouped() map[string][]VersionedResource {
    // groupResources / newGroupedVersionedResources
}

func (d DiffResourcesFactory) newVersionedRs() versionedResources {
    // newVersionedResources
}
```

pkg/kapp/diff/stripnamehashsuffix.go
praveenrewar: We already have a notMatcher present. Do you think that will be useful instead of `exclude`? 
So, this how the matchers currently work. 
By default, any matcher is used. So you can just have something like this to match ConfigMaps and Secrets.
```yaml
stripNameHashSuffix:
- resourceMatchers:
  - apiVersionKindMatcher: {apiVersion: v1, kind: ConfigMap}
  - apiVersionKindMatcher: {apiVersion: v1, kind: Secret}
```
If you want to exclude something, then you can use an andMatcher (with anyMatcher and notMatcher nested inside), something like [this](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/config/default.go#L224-L235). So I think we can rely on the existing capabilities matchers.

criztovyl: The challenge then is aggregating such nested "and: (any , not)" matchers across configurations.

praveenrewar: > across configurations

Ideally we would want to have one Config only (although we could add as many as we want). Do you have any specific scenario where we would need multiple Configs?

criztovyl: I thought this over, no use-case springs to mind; Ii wanted to include it for completeness and the include/exclude was the latest pragamtic way I found to implement it.

I would then explicitly implement it so that only the suffix-config from last kapp-config is used :)

PR #523, opened 2022-06-03
Add status code check for http file sources

### LINKED ISSUE #521: Intermitent failure when executing kapp deploy command
**What steps did you take:**
Occasionally, executing kapp deploy command fails with the following error

**What happened:**
```
      if [ "$(kubectl api-resources --api-group=kappctrl.k14s.io -o name | wc -l)" -eq 0 ] ; then \
      	kapp deploy \
      		--yes \
      		--app=kc \
      		--namespace=kube-system \
      		--file=https://github.com/vmware-tanzu/carvel-kapp-controller/releases/latest/download/release.yml ; \
      fi
      Target cluster 'https://<omitted>:<port>' (nodes: <omitted>, 3+)
        kapp: Error: error converting YAML to JSON: yaml: line 8: mapping values are not allowed in this context
```      
      
Kapp version: `kapp version 0.47.0`

praveenrewar:
- Use http.Client instead of http to make get calls
- Add status code check to error out for non OK http status codes
   (OK = 2xx)
- Add unit tests for http file sources

Fixes #521 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

Current behaviour
```bash
$ kapp deploy -a test -f https://github.com/fake/path/
Target cluster 'https://192.168.64.93:8443' (nodes: minikube)

kapp: Error: error converting YAML to JSON: yaml: line 163: mapping values are not allowed in this context
```

Expected behaviour:
```bash
$ /kapp deploy -a test -f https://github.com/fake/path/                                      
Target cluster 'https://192.168.64.93:8443' (nodes: minikube)

kapp: Error: Requesting URL 'https://github.com/fake/path/': 404 Not Found
```

### DIFF
--- a/pkg/kapp/resources/file_sources.go
+++ b/pkg/kapp/resources/file_sources.go
@@ -44,24 +44,29 @@ func (s LocalFileSource) Description() string        { return fmt.Sprintf("file 
 func (s LocalFileSource) Bytes() ([]byte, error)     { return ioutil.ReadFile(s.path) }
 
 type HTTPFileSource struct {
-	url string
+	url    string
+	Client *http.Client
 }
 
 var _ FileSource = HTTPFileSource{}
 
-func NewHTTPFileSource(path string) HTTPFileSource { return HTTPFileSource{path} }
+func NewHTTPFileSource(path string) HTTPFileSource { return HTTPFileSource{path, &http.Client{}} }
 
 func (s HTTPFileSource) Description() string {
 	return fmt.Sprintf("HTTP URL '%s'", s.url)
 }
 
 func (s HTTPFileSource) Bytes() ([]byte, error) {
-	resp, err := http.Get(s.url)
+	resp, err := s.Client.Get(s.url)
 	if err != nil {
 		return nil, fmt.Errorf("Requesting URL '%s': %s", s.url, err)
 	}
 
 	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		return nil, fmt.Errorf("Requesting URL '%s': %s", s.url, resp.Status)
+	}
 
 	result, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/kapp/resources/file_sources.go
```diff
@@ -44,25 +44,30 @@ func (s LocalFileSource) Description() string        { return fmt.Sprintf("file
...
 func (s HTTPFileSource) Bytes() ([]byte, error) {
-	resp, err := http.Get(s.url)
+	resp, err := s.Client.Get(s.url)
 	if err != nil {
 		return nil, fmt.Errorf("Requesting URL '%s': %s", s.url, err)
 	}
 
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
```
cppforlife: i think loosing it up would be nice. e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status 200-299

praveenrewar: The other status codes, for example 201 Created are not very relevant so I just included 200, but I don't have any strong opinions, I will make the change to include 2xx.

praveenrewar: Done.

100mik (commented):
LGTM!
I think using status OK would be more appropriate for the test, but since we are going ahead with a range, I guess it makes sense to have something that is in the middle 🚀

100mik (approved):

kumaritanushree:
LGTM

pkg/kapp/resources/file_sources_test.go
```diff
@@ -0,0 +1,80 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package resources
```
cppforlife: should be under _test suffix

praveenrewar: Thanks! Done.

PR #527, opened 2022-06-07
Enhance tty experience

### LINKED ISSUE #239: kapp inspect --raw should not generate header/footer informational message to stdout
**What steps did you take:**

Run:

```
kapp inspect -a example --raw > resources.yaml
```

**What happened:**

This generates all the resources making up the application deployment, but the header/footer information messages are also generated to stdout, meaning they end up in the yaml file with the resources. This makes it impossible to then do further processing on the YAML file without first editing the file to remove the messages.

Eg., file contains:

```
Target cluster 'https://vcluster.lab-package-testing-w01-s013-vc.svc.cluster.local' (nodes: kind-control-plane)

---
apiVersion: apps/v1
...

Succeeded
```

**What did you expect:**

The header/footer should either not be generated at all, or should be generated to stderr instead.

This would then allow processing of resulting YAML file, including being able to pipe it directly into other commands.

Eg.

```
kapp inspect -a example --raw | kbld inspect -f -
```

Nope.

**Environment:**

```
$ kapp --version
kapp version 0.37.0
```

---
Vote on this request

cc @jorgemoralespou

praveenrewar:
Currently default value of `--tty` is true and therefore even when using `non tty` outputs like files, CIs, users have to manually set the value of `--tty` flag to false.
- Set default value of `--tty` to false.
- Set default value of `--tty` to true for `deploy` and `delete` command. This is done to ensure that even while using CIs, `kapp deploy` will show all the logs. `--tty=false` can still be used if less verbose output is desired.

Fixes #239

```release-note
Default value of tty flag is set to false for commands other than deploy and delete
```

The flag is now added to each of the commands separately instead of a persistent flag on the root command. Therefore it's no longer a global flag.

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -47,6 +47,7 @@ func NewDeleteCmd(o *DeleteOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 		RunE:  func(_ *cobra.Command, _ []string) error { return o.Run() },
 		Annotations: map[string]string{
 			cmdcore.AppHelpGroup.Key: cmdcore.AppHelpGroup.Value,
+			TTYByDefaultKey:          "",
 		},
 	}
 	o.AppFlags.Set(cmd, flagsFactory)
--- a/pkg/kapp/cmd/kapp.go
+++ b/pkg/kapp/cmd/kapp.go
@@ -138,6 +138,18 @@ func NewKappCmd(o *KappOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Comman
 		return nil
 	})
 
+	configureTTYFlag := func(cmd *cobra.Command) {
+		var forceTTY bool
+		// Default tty to true for deploy and delete commands: https://github.com/vmware-tanzu/carvel-kapp/issues/28
+		_, defaultTTY := cmd.Annotations[cmdapp.TTYByDefaultKey]
+
+		cmd.Flags().BoolVar(&forceTTY, "tty", defaultTTY, "Force TTY-like output")
+		cobrautil.VisitCommands(cmd, cobrautil.WrapRunEForCmd(func(*cobra.Command, []string) error {
+			o.ui.EnableTTY(forceTTY)
+			return nil
+		}))
+	}
+
 	cobrautil.VisitCommands(cmd, cobrautil.ReconfigureLeafCmds(cobrautil.DisallowExtraArgs))
 
 	// Completion command have to be added after the cobrautil.DisallowExtraArgs.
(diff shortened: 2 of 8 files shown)


pkg/kapp/cmd/kapp.go
```diff
@@ -138,6 +138,17 @@ func NewKappCmd(o *KappOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Comman
 		return nil
 	})
 
+	configureTTY := func(cmd *cobra.Command) {
+		var forceTTY bool
+		// Default tty to true for deploy and delete commands: https://github.com/vmware-tanzu/carvel-kapp/issues/28
+		_, defaultTTY := cmd.Annotations["cli.carvel.dev/tty-by-default"]
+
+		cmd.Flags().BoolVar(&forceTTY, "tty", defaultTTY, "Force TTY-like output")
+		cmd.PreRun = func(cmd *cobra.Command, args []string) {
```
cppforlife: instead of PreRun, lets try using cobrautil.WrapRunEForCmd (https://github.com/vmware-tanzu/carvel-kapp/blob/331c4ef932b203a4acee83b98e239a0b05d11a4d/vendor/github.com/cppforlife/cobrautil/misc.go#L33). otherwise we would have to consider what if PreRun is already set, etc...

pkg/kapp/cmd/app/delete.go
```diff
@@ -46,7 +46,8 @@ func NewDeleteCmd(o *DeleteOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 		Short: "Delete app",
 		RunE:  func(_ *cobra.Command, _ []string) error { return o.Run() },
 		Annotations: map[string]string{
-			cmdcore.AppHelpGroup.Key: cmdcore.AppHelpGroup.Value,
+			cmdcore.AppHelpGroup.Key:        cmdcore.AppHelpGroup.Value,
+			"cli.carvel.dev/tty-by-default": "",
```
cppforlife: can we pull this into a constant somewhere. seems like going to be easy to "typo" it and not realize it.

pkg/kapp/cmd/kapp.go
praveenrewar: > otherwise we would have to consider what if PreRun is already set, etc...

Totally makes sense. I will make the change.

pkg/kapp/cmd/app/delete.go
praveenrewar: Yep, I thought of moving it to cmdcore, but it felt weird to have it there and I kept it as it is. I will move it to cmdapp as it is being used there the most and it won't cause any cyclic dependency as well.

pkg/kapp/cmd/kapp.go
praveenrewar: Done

pkg/kapp/cmd/app/delete.go
praveenrewar: Done.

PR #532, opened 2022-06-18
Apply as many changes as possible with error summary at end

### LINKED ISSUE #426: Don't exit at first error but continues and summarizes all the found ones at the end of execution
We want to use kapp to reconcile our e2e tests. We're trying to deploy multiple CRs that don't depend on each other, they can be deployed in parallel. There's no dependency between CRs, so if any of those fails it should not stop the other ones from being reconciled. Currently the kapp command is exiting as soon as it finds an error, with a message like this one: 
```
kapp: Error: waiting on reconcile customresource/resource-name namespace: default:
  Finished unsuccessfully (Encountered failure condition Succeeded == False: Error (message: The job did not complete successfully))
```
So we're not getting feedback about the success/failure of all the other CRs that were trying to get reconciled.

We would like a new CLI argument (e.g.: --ignore-errors or --no-abort) or a new configuration field to indicate to the kapp CLI to wait for all the possible results before exiting, indicating which ones failed in which steps of the reconciliation process. 

Output would look like: "Reconciled 10/20. Finished unsuccessfully: 5. Finished successfully: 5. Not deployed: 10."

The configuration file we're using to replicate this issue looks something like this: 
```
kind: Config

waitRules:
    conditionMatchers:
      - type: Succeeded
        status: "True"
        success: true
      - type: Succeeded
        status: "False"
        failure: true
    resourceMatchers:
      - andMatcher:
          matchers:
            - apiVersionKindMatcher: {apiVersion: my-custom-resource-definition.com/v1beta1, kind: CustomResource}
            - hasAnnotationMatcher: { keys: [kapp-matcher-success] }
    conditionMatchers:
      - type: Succeeded
        status: "False"
        success: true
      - type: Succeeded
        status: "True"
        failure: true
    resourceMatchers:
      - andMatcher:
          matchers:
            - apiVersionKindMatcher: {apiVersion: my-custom-resource-definition.com/v1beta1, kind: CustomResource}
            - hasAnnotationMatcher: { keys: [kapp-matcher-failure] }
```

---
Vote on this request

praveenrewar:
Currently kapp errors out as soon as it detects an error while apply a resource or waiting for a resource, but we would want to apply as many changes as possible and return all the errors at end. If a change doesn't depend on a failed change then it should get applied.

- Changes that depend on failing changes shouldn't be applied
- `exit-early-on-apply-error` flag can be used to exit as soon as an error is encountered while applying changes (default true)
- `exit-early-on-wait-error` flag can be used to exit as soon as an error is encountered while waiting for changes (default true)

Sample gist for testing: https://gist.github.com/praveenrewar/2746318c5a273d13ed26497f34755638

TODO:
- [x] Add e2e test
- [x] Do more manual testing

Fixes #426 

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/clusterapply/applying_changes.go
+++ b/pkg/kapp/clusterapply/applying_changes.go
@@ -23,10 +23,11 @@ type ApplyingChanges struct {
 	applied              map[*ctldgraph.Change]struct{}
 	clusterChangeFactory ClusterChangeFactory
 	ui                   UI
+	exitOnError          bool
 }
 
-func NewApplyingChanges(numTotal int, opts ApplyingChangesOpts, clusterChangeFactory ClusterChangeFactory, ui UI) *ApplyingChanges {
-	return &ApplyingChanges{numTotal, opts, map[*ctldgraph.Change]struct{}{}, clusterChangeFactory, ui}
+func NewApplyingChanges(numTotal int, opts ApplyingChangesOpts, clusterChangeFactory ClusterChangeFactory, ui UI, exitOnError bool) *ApplyingChanges {
+	return &ApplyingChanges{numTotal, opts, map[*ctldgraph.Change]struct{}{}, clusterChangeFactory, ui, exitOnError}
 }
 
 type applyResult struct {
@@ -37,14 +38,16 @@ type applyResult struct {
 	Err           error
 }
 
-func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange, error) {
+func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange, []string, error) {
 	startTime := time.Now()
+
+	var unsuccessfulChangeDesc []string
 
 	for {
 		nonAppliedChanges := c.nonAppliedChanges(allChanges)
 		if len(nonAppliedChanges) == 0 {
 			// Do not print applying message if no changes
-			return nil, nil
+			return nil, unsuccessfulChangeDesc, nil
 		}
 
 		c.ui.NotifySection("applying %d changes %s", len(nonAppliedChanges), c.stats())
@@ -88,10 +91,14 @@ func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange
 
 			if result.Err != nil {
 				lastErr = result.Err
-				if result.Retryable {
-					continue
+				if !result.Retryable {
+					if c.exitOnError {
+						return nil, nil, result.Err
+					}
+					unsuccessfulChangeDesc = append(unsuccessfulChangeDesc, result.Err.Error())
+					c.markApplied(result.Change)
 				}
-				return nil, result.Err
+				continue
 			}
 
 			c.markApplied(result.Change)
@@ -99,11 +106,11 @@ func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange
 		}
 
 		if len(appliedChanges) > 0 {
-			return appliedChanges, nil
+			return appliedChanges, unsuccessfulChangeDesc, nil
 		}
 
 		if time.Now().Sub(startTime) > c.opts.Timeout {
-			return nil, fmt.Errorf("Timed out waiting after %s: Last error: %w", c.opts.Timeout, lastErr)
+			return nil, unsuccessfulChangeDesc, fmt.Errorf("Timed out waiting after %s: Last error: %s", c.opts.Timeout, lastErr)
 		}
 
 		time.Sleep(c.opts.CheckInterval)
--- a/pkg/kapp/clusterapply/cluster_change_set.go
+++ b/pkg/kapp/clusterapply/cluster_change_set.go
@@ -5,7 +5,9 @@ package clusterapply
 
 import (
 	"fmt"
+	"strings"
 
+	uierrs "github.com/cppforlife/go-cli-ui/errors"
 	ctlconf "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/config"
 	ctldiff "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/diff"
 	ctldgraph "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/diffgraph"
@@ -15,6 +17,9 @@ import (
 type ClusterChangeSetOpts struct {
 	ApplyingChangesOpts
 	WaitingChangesOpts
+
+	ExitEarlyOnApplyError bool
+	ExitEarlyOnWaitError  bool
 }
 
 type ClusterChangeSet struct {
@@ -95,18 +100,30 @@ func (c ClusterChangeSet) Apply(changesGraph *ctldgraph.ChangeGraph) error {
 
 	blockedChanges := ctldgraph.NewBlockedChanges(changesGraph)
 	applyingChanges := NewApplyingChanges(
-		expectedNumChanges, c.opts.ApplyingChangesOpts, c.clusterChangeFactory, c.ui)
-	waitingChanges := NewWaitingChanges(expectedNumChanges, c.opts.WaitingChangesOpts, c.ui)
+		expectedNumChanges, c.opts.ApplyingChangesOpts, c.clusterChangeFactory, c.ui, c.opts.ExitEarlyOnApplyError)
+	waitingChanges := NewWaitingChanges(expectedNumChanges, c.opts.WaitingChangesOpts, c.ui, c.opts.ExitEarlyOnWaitError)
+
+	var unsuccessfulChanges []string
 
 	for {
-		appliedChanges, err := applyingChanges.Apply(blockedChanges.Unblocked())
+		appliedChanges, unsuccessfulChangeDesc, err := applyingChanges.Apply(blockedChanges.Unblocked())
 		if err != nil {
 			return err
 		}
 
+		unsuccessfulChanges = append(unsuccessfulChanges, unsuccessfulChangeDesc...)
+
 		waitingChanges.Track(appliedChanges)
 
 		if waitingChanges.IsEmpty() {
+			if len(unsuccessfulChanges) == 1 {
+				return fmt.Errorf("%s", unsuccessfulChanges[0])
+			}
+
+			if len(unsuccessfulChanges) > 0 {
+				return uierrs.NewSemiStructuredError(fmt.Errorf("[%s]", strings.Join(unsuccessfulChanges, ", ")))
+			}
+
 			err := applyingChanges.Complete()
 			if err != nil {
 				c.ui.Notify([]string{fmt.Sprintf("Blocked changes:\n%s\n", blockedChanges.WhyBlocked(blockedChanges.Blocked()))})
@@ -116,10 +133,12 @@ func (c ClusterChangeSet) Apply(changesGraph *ctldgraph.ChangeGraph) error {
 			return waitingChanges.Complete()
 		}
 
-		doneChanges, err := waitingChanges.WaitForAny()
+		doneChanges, unsuccessfulChangeDesc, err := waitingChanges.WaitForAny()
 		if err != nil {
 			return err
 		}
+
+		unsuccessfulChanges = append(unsuccessfulChanges, unsuccessfulChangeDesc...)
 
 		for _, change := range doneChanges {
 			blockedChanges.Unblock(change.Graph)
--- a/pkg/kapp/cmd/app/apply_flags.go
+++ b/pkg/kapp/cmd/app/apply_flags.go
@@ -50,6 +50,8 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 	cmd.Flags().StringVar(&s.AddOrUpdateChangeOpts.DefaultUpdateStrategy, prefix+"apply-default-update-strategy",
 		defaults.AddOrUpdateChangeOpts.DefaultUpdateStrategy, "Change default update strategy")
 
+	cmd.Flags().BoolVar(&s.ExitEarlyOnApplyError, prefix+"exit-early-on-apply-error", true, "Exit quickly on apply failure")
+
 	cmd.Flags().BoolVar(&s.Wait, prefix+"wait", defaults.Wait, "Set to wait for changes to be applied")
 	cmd.Flags().BoolVar(&s.WaitIgnored, prefix+"wait-ignored", defaults.WaitIgnored, "Set to wait for ignored changes to be applied")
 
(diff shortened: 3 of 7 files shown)


pkg/kapp/clusterapply/cluster_change_set.go
```diff
@@ -98,15 +100,22 @@ func (c ClusterChangeSet) Apply(changesGraph *ctldgraph.ChangeGraph) error {
...
 	waitingChanges := NewWaitingChanges(expectedNumChanges, c.opts.WaitingChangesOpts, c.ui)
 
+	var unsuccessfulChanges []string
+
 	for {
-		appliedChanges, err := applyingChanges.Apply(blockedChanges.Unblocked())
+		appliedChanges, unsuccessfulChangeDesc, err := applyingChanges.Apply(blockedChanges.Unblocked())
 		if err != nil {
 			return err
 		}
```
praveenrewar, line 112: Should the timeout errors also include the unsuccessfulChanges?

100mik: I think they should!

100mik (commented):
Some thoughts:
- Should there be a flag which enables disables this behaviour? What should be it's default value?
- How does this look with multiple failures? I think we might need better formatting than `[error1, error2, error3]` maybe more like:
```
- error1
- error2
- error3
```

praveenrewar:
> Should there be a flag which enables disables this behaviour? What should be it's default value?

Hmm, maybe, but I am not sure why someone would want to turn it off.

> How does this look with multiple failures? 

It's just like you have mentioned 😁 please do take a look at the gist that I have shared, it contains an example manifest and the output that we would get while deploying it.

100mik:
> Hmm, maybe, but I am not sure why someone would want to turn it off.

One case I can think of is pipelines. They would probably want to error out the moment something goes south rather than wait for other resources to be created.
_Some_ packages might be interested in this I guess, but I do not see a super strong case for it. (Not sure if kapp-controller will choose to error out by default because the installation fails none the less)
>It's just like you have mentioned 😁 please do take a look at the gist that I have shared, it contains an example manifest and the output that we would get while deploying it.

Oooh noice! Missed that. I noticed there was a ']' added to a test and misunderstood 🤔

praveenrewar:
> They would probably want to error out the moment something goes south rather than wait for other resources to be created.

I don't know how that would be useful, but maybe it would be useful in some scenarios that I am not able to think at the moment, so I am keeping this discussion open for now, we can discuss more later.

> I noticed there was a ']' added to a test and misunderstood

Yeah, I am still trying to figure out how to best use `NewSemiStructuredError`. Right now it formats properly when there are more errors (>1), but if there is just one error then it keeps the brackets. I will look more into it.

pkg/kapp/clusterapply/applying_changes.go
```diff
@@ -85,25 +87,25 @@ func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange
 			result := <-applyCh
 
 			c.ui.Notify(result.DescMsgs)
-
 			if result.Err != nil {
 				lastErr = result.Err
-				if result.Retryable {
-					continue
+				if !result.Retryable {
+					unsuccessfulChangeDesc = append(unsuccessfulChangeDesc, result.Err.Error())
```
rohitagg2020, line 98: If a resource fails, then all the dependent resource in it in the change graph, what will happen to them ? Also, if a resource depends on two resources, one of them passes and other fails, then what is going to happen ?

test/e2e/custom_wait_rules_test.go
```diff
@@ -116,6 +116,7 @@ status:
 
 		require.Contains(t, res, "Current state as Failed")
 
-		require.Contains(t, err.Error(), "kapp: Error: waiting on reconcile crontab/my-new-cron-object-1")
+		require.Contains(t, err.Error(), "kapp: Error: Finished unsuccessfully")
+		require.Contains(t, err.Error(), "crontab/my-new-cron-object-1 (stable.example.com/v1) namespace: kapp-test (Current state as Failed)")
```
rohitagg2020: For timeout, does it make sense to show as timed-out error instead of `(Current state as Failed)`.

pkg/kapp/clusterapply/applying_changes.go
praveenrewar: If a resource is blocked on another resource which has failed, then it won't get applied.

test/e2e/custom_wait_rules_test.go
praveenrewar: Sure. 
(except this is not a timed-out error)

pkg/kapp/clusterapply/cluster_change_set.go
praveenrewar: I am still thinking if we should do this and what would be the best way to present the errors.

cppforlife:
- we should probably have a flag that fails on first failure (current behaviour) so that folks can still "quickly" exit if that's what they want.
- how does this feature play with timeout errors? should we be erroring on timeout errors immediately?

praveenrewar:
> we should probably have a flag that fails on first failure (current behaviour) so that folks can still "quickly" exit if that's what they want.

Yeah, that's something @100mik had suggested as well. I had kept this discussion open till now, I will add a flag to disable the behaviour.

> how does this feature play with timeout errors? 

Right now, we error out for `--wait-timeout` but not for `--wait-resource-timeout`.

> should we be erroring on timeout errors immediately?

I was thinking that individual resource timeouts can be treated as resource errors in this case, as one resource getting timed out doesn’t mean that others (which don't depend on it) will also get timed out.

test/e2e/formatted_error_test.go
```diff
@@ -50,16 +50,13 @@ spec:
 
 	logger.Section("deploy with errors", func() {
 		expectedErr := strings.TrimSpace(`
-kapp: Error: Applying create job/successful-job (batch/v1) namespace: default:
-  Creating resource job/successful-job (batch/v1) namespace: default:
-    API server says:
-      Job.batch "successful-job" is invalid: 
+kapp: Error: Finished unsuccessfully: [create job/successful-job (batch/v1) namespace: default: Creating resource job/successful-job (batch/v1) namespace: default: API server says: Job.batch "successful-job" is invalid:
```
cppforlife: do we not get nicely formatted messages anymore?

praveenrewar: Apparently not when there is only 1 error in the array. I will take a closer look at `SemiStructuredError`.

pkg/kapp/clusterapply/cluster_change_set.go
```diff
@@ -95,18 +99,30 @@ func (c ClusterChangeSet) Apply(changesGraph *ctldgraph.ChangeGraph) error {
...
 
+		unsuccessfulChanges = append(unsuccessfulChanges, unsuccessfulChangeDesc...)
+
 		waitingChanges.Track(appliedChanges)
 
 		if waitingChanges.IsEmpty() {
+			if len(unsuccessfulChanges) == 1 {
```
praveenrewar, line 119: `NewSemiStructuredError` doesn't work really well when there is only one error inside `[]` because there might be errors containing something like `map[string]string` and there is no good way to differentiate between an error and such field.
I should probably add this as a comment.

100mik:
The changes themselves LGTM 🤔 

Re: Error structuring
I think we are:
- Retaining how we used to show errors for one error by using the `len(unsuccessfulChanges) == 1` check
- However multiple errors won't have the formatting due to capitalisation?

kumaritanushree:
Suggestion: I think it would be good if you can add little more description like:  how the new output will look like? You can add some example output.

praveenrewar:
> Retaining how we used to show errors for one error by using the len(unsuccessfulChanges) == 1 check

That's correct.

> However multiple errors won't have the formatting due to capitalisation?

That doesn't happen in NewSemistructuredError.

praveenrewar:
> Suggestion: I think it would be good if you can add little more description like: how the new output will look like? You can add some example output.

I think the test cases and the gist that I have provided should be a good indication of the output, but do let me know if I should add more examples.

100mik:
So I am guessing multiple errors will be unformatted 🤔 
Also, I m not entirely sure if the way we format a single message will help in case of multiple messages.
But I think we should explore if we have any options better than having it in one line.

That said I do not think it is a major blocker but more of a nice to have thing imo 🚀

sethiyash (commented):
LGTM. Just thinking about the output. For the example given in the [gist](https://gist.github.com/praveenrewar/2746318c5a273d13ed26497f34755638) wouldn't it be better to show `2 Resource Succeded` before showing errors or final summary at the end on number of resources succeeded and failed.

praveenrewar:
> wouldn't it be better to show 2 Resource Succeded before showing errors or final summary at the end on number of resources succeeded and failed.

Hmm, definitely a good thought. The reason I didn't include more details around successful changes at the end is because we already include enough information in the logs (`ok: reconcile...`). I also was trying to keep the ui changes to the minimum, but I am open to creating a separate PR if we want to enhance the experience a bit further later on.

100mik:
I think I am satisfied with how the UI changes look as well, I cannot think of how we can make a list of errors look better right away. But these cosmetic improvements can always go as a separate PR if we have better ideas.

100mik:
Also, @praveenrewar looks like this branch needs a rebase. I will take one final look at this once it is done!

pkg/kapp/cmd/app/apply_flags.go
```diff
@@ -63,6 +63,8 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 		5, "Maximum number of concurrent wait operations")
 
 	cmd.Flags().BoolVar(&s.ExitStatus, prefix+"apply-exit-status", false, "Return specific exit status based on number of changes")
+
+	cmd.Flags().BoolVar(&s.ExitOnError, prefix+"exit-on-apply-wait-error", false, "Exit quickly on apply or wait failure")
```
100mik: Did we not wanna set the default to true to start off with? (as per the discussion in the community meeting, where more users would like to know immediately If something is not working out)

nit: "Exit quickly on" => "Exit immediately on"/"Exit on"
I guess quickly is a bit more open ended than immediately?

nit:(to self) I am alright with flag name but gonna give it some thought 🤔

pkg/kapp/clusterapply/applying_changes.go
```diff
@@ -88,22 +91,26 @@ func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange
 
 			if result.Err != nil {
 				lastErr = result.Err
-				if result.Retryable {
-					continue
+				if !result.Retryable {
+					if c.exitOnError {
+						return nil, nil, result.Err
+					}
+					unsuccessfulChangeDesc = append(unsuccessfulChangeDesc, result.Err.Error())
```
100mik, line 98: A thought in  line with the default behaviour, if we do decide to apply as many changes as we can by default.
Should we print a message like `failed: reconcile deployment/kapp-controller (apps/v1) namespace: kc` the moment a resource fails?

Not convinced about this myself. It does not benefit cases where kapp is being run as a part of a script.
However I am guessing I would not like to wait till everything has finished applying know that a resource has failed

100mik (commented):
A few nitty thoughts!
Over all the implementation looks good 🚀

pkg/kapp/cmd/app/apply_flags.go
praveenrewar: > Did we not wanna set the default to true to start off with?

Yeah, that's something I am still thinking about. I am okay with setting default to true for now.

> nit: "Exit quickly on" => "Exit immediately on"/"Exit on"

Hehe, I started with `Exit on...` at first but realised that we are always going to exit in case of a failure, but yeah `Exit immediately...` makes more sense.

> nit:(to self) I am alright with flag name but gonna give it some thought 🤔

I am still having second thoughts about the name, so open to some suggestions.

pkg/kapp/clusterapply/applying_changes.go
praveenrewar: > Should we print a message like ok: reconcile deployment/kapp-controller (apps/v1) namespace: kc the moment a resource fails?

I didn't understand this part, do you mean we should print `ok: reconcile...` for a resource when it fails?

100mik: No! My bad, updated the comment to reflect what I meant ^

100mik: Not sure if we should do this, but I think it might be a nice to have if the default behaviour is to apply maximally

pkg/kapp/cmd/app/apply_flags.go
100mik: > I am still having second thoughts about the name, so open to some suggestions.

I cannot think of anything that is as easy to understand as this one.

> Yeah, that's something I am still thinking about. I am okay with setting default to true for now.

I think that might be the way to go judging from the discussion we had. Also prevents this from being a breaking change, even though it is not a mean one.

pkg/kapp/clusterapply/applying_changes.go
praveenrewar: Ah, I see.

So, we are doing something like this
```bash
3:38:16PM: fail: reconcile job/my-job-fail (batch/v1) namespace: default
3:38:16PM:  ^ Failed with reason BackoffLimitExceeded: Job has reached the specified backoff limit
```
(^ Taken from the gist example)

100mik: Oops! I missed that. Resolving this 🚀

pkg/kapp/cmd/app/apply_flags.go
100mik: Bumping this ^ 
I think this is the major change blocking this?

praveenrewar: Yep, will have a better answer once I have done some more testing with this enabled.
(Although I am thinking that we should go ahead with the existing behaviour i.e exit on error, so that we can get some feedback about how intuitive the output is before we go on and make this as a default behaviour)

pkg/kapp/cmd/app/apply_flags.go
```diff
@@ -50,6 +50,8 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 	cmd.Flags().StringVar(&s.AddOrUpdateChangeOpts.DefaultUpdateStrategy, prefix+"apply-default-update-strategy",
 		defaults.AddOrUpdateChangeOpts.DefaultUpdateStrategy, "Change default update strategy")
 
+	cmd.Flags().BoolVar(&s.ExitOnApplyError, prefix+"exit-on-apply-error", false, "Exit quickly on apply failure")
```
rohitagg2020: nit: Rename the flag to `exit-early-on-apply-error`

pkg/kapp/cmd/app/apply_flags.go
```diff
@@ -63,6 +65,8 @@ func (s *ApplyFlags) SetWithDefaults(prefix string, defaults ApplyFlags, cmd *co
 		5, "Maximum number of concurrent wait operations")
 
 	cmd.Flags().BoolVar(&s.ExitStatus, prefix+"apply-exit-status", false, "Return specific exit status based on number of changes")
+
+	cmd.Flags().BoolVar(&s.ExitOnWaitError, prefix+"exit-on-wait-error", true, "Exit quickly on wait failure")
```
rohitagg2020: nit: Rename the flag to `exit-early-on-wait-error`

100mik (approved):
LGTM!

PR #533, opened 2022-06-19
Add resource namespaces saved during deploy to fallbackAllowedNamespaces

### LINKED ISSUE #479: kapp should add namespaces used by resources to fallback namespaces
currently fallback namespace for finding existing resources is only single namespace that is specified via --namespace flag. fallback namespaces are needed by kapp cannot find deployed resources easily (across all namespace). this becomes important if configuration is deployed across multiple namespaces but kapp does not have perms to list all namespaces (which is desirable if you have a lot of namespaces on the cluster). 

currently only --namespace is added as fallback namespace. we should populate fallback namespaces with namespaces used by resources at the time of the deploy, and since we track them in state configmap used a union of them with new namespaces on the next deploy.

to reproduce: kapp deploy a set of resources across 3 namespaces but make sure that service-account/user given to kapp does not have ability to list namespaces on the cluster.

---
Vote on this request

praveenrewar:
Add resource namespaces saved during deploy to fallbackAllowedNamespaces.
These resource namespaces are stored as app change meta during deploy

Fixes #479 

```release-note
NONE
```

 I am still thinking if there's a better place to make this change. Another option that I thought of was to call the [FactoryClients](https://github.com/vmware-tanzu/carvel-kapp/blob/dab63af7fa338e8ba9c9a222d09e1b781aa7f5a6/pkg/kapp/cmd/app/factory.go#L21) separately and a little late (after [Exists](https://github.com/vmware-tanzu/carvel-kapp/blob/dab63af7fa338e8ba9c9a222d09e1b781aa7f5a6/pkg/kapp/cmd/app/delete.go#L70) function is called) and populate the fallbackAllowedNamespaces with the resource namespaces there itself, but that would require a lot of repetition as [App.Find](https://github.com/vmware-tanzu/carvel-kapp/blob/dab63af7fa338e8ba9c9a222d09e1b781aa7f5a6/pkg/kapp/cmd/app/factory.go#L73) requires a lot of things from `FactoryClients`.
 
##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -152,8 +152,13 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	meta, err := app.Meta()
+	if err != nil {
+		return err
+	}
+
 	existingResources, existingPodRs, err := o.existingResources(
-		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs)
+		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...))
 	if err != nil {
 		return err
 	}
@@ -202,7 +207,7 @@ func (o *DeployOptions) Run() error {
 	if o.DeployFlags.Logs {
 		cancelLogsCh := make(chan struct{})
 		defer func() { close(cancelLogsCh) }()
-		go o.showLogs(supportObjs.CoreClient, supportObjs.IdentifiedResources, existingPodRs, labelSelector, cancelLogsCh)
+		go o.showLogs(supportObjs.CoreClient, supportObjs.IdentifiedResources, existingPodRs, labelSelector, cancelLogsCh, append(meta.LastChange.Namespaces, nsNames...))
 	}
 
 	defer func() {
@@ -356,7 +361,7 @@ func (o *DeployOptions) newResourcesFromFiles() ([]ctlres.Resource, error) {
 
 func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 	labeledResources *ctlres.LabeledResources, resourceFilter ctlres.ResourceFilter,
-	apps ctlapp.Apps, usedGKs []schema.GroupKind) ([]ctlres.Resource, []ctlres.Resource, error) {
+	apps ctlapp.Apps, usedGKs []schema.GroupKind, resourceNamespaces []string) ([]ctlres.Resource, []ctlres.Resource, error) {
 
 	labelErrorResolutionFunc := func(key string, val string) string {
 		items, _ := apps.List(nil)
@@ -380,7 +385,8 @@ func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 
 		//Scope resource searching to UsedGKs
 		IdentifiedResourcesListOpts: ctlres.IdentifiedResourcesListOpts{
-			GKsScope: usedGKs,
+			GKsScope:           usedGKs,
+			ResourceNamespaces: resourceNamespaces,
 		},
 	}
 
@@ -484,7 +490,7 @@ const (
 
 func (o *DeployOptions) showLogs(
 	coreClient kubernetes.Interface, identifiedResources ctlres.IdentifiedResources,
-	existingPodRs []ctlres.Resource, labelSelector labels.Selector, cancelCh chan struct{}) {
+	existingPodRs []ctlres.Resource, labelSelector labels.Selector, cancelCh chan struct{}, resourceNamespaces []string) {
 
 	existingPodsByUID := map[string]struct{}{}
 
@@ -518,7 +524,7 @@ func (o *DeployOptions) showLogs(
 
 	podWatcher := ctlres.FilteringPodWatcher{
 		podMatcherFunc,
-		identifiedResources.PodResources(labelSelector),
+		identifiedResources.PodResources(labelSelector, resourceNamespaces),
 	}
 
 	contFilterFunc := func(pod corev1.Pod) []string {
--- a/pkg/kapp/resources/identified_resources_pods.go
+++ b/pkg/kapp/resources/identified_resources_pods.go
@@ -5,6 +5,7 @@ package resources
 
 import (
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -35,28 +36,34 @@ func (w UniquePodWatcher) Watch(podsToWatchCh chan corev1.Pod, cancelCh chan str
 	go func() {
 		// Watch Pods in all namespaces first and fallback to the
 		// fallbackAllowedNamespaces if lack of permission
-		namespace := ""
-		index := 0
-		for {
+		namespaces := []string{""}
+		namespaces = append(namespaces, w.fallbackAllowedNamespaces...)
+		var forbiddenNamespaces []string
+
+		for _, namespace := range namespaces {
 			podWatcher := NewPodWatcher(
 				w.coreClient.CoreV1().Pods(namespace),
 				metav1.ListOptions{LabelSelector: w.labelSelector.String()},
 			)
-
 			err := podWatcher.Watch(nonUniquePodsToWatchCh, cancelCh)
-			if err != nil {
-				if !errors.IsForbidden(err) {
-					fmt.Printf("Pod watching error: %s\n", err) // TODO
+			if err == nil {
+				if namespace == "" {
 					break
 				}
+				continue
 			}
-
-			if len(w.fallbackAllowedNamespaces) > index {
-				namespace = w.fallbackAllowedNamespaces[index]
-				index++
+			if !errors.IsForbidden(err) {
+				fmt.Printf("Pod watching error: %s\n", err) // TODO
+				break
+			}
+			if namespace != "" {
+				forbiddenNamespaces = append(forbiddenNamespaces, fmt.Sprintf(`"%s"`, namespace))
 			}
 		}
 
+		if len(forbiddenNamespaces) > 0 {
+			fmt.Printf(`Pod watching error: pods is forbidden: User cannot list resource "pods" in API group "" in the namespace(s) %s`, strings.Join(forbiddenNamespaces, ", "))
+		}
 		close(nonUniquePodsToWatchCh)
 	}()
 
--- a/pkg/kapp/resources/identified_resources_pods.go
+++ b/pkg/kapp/resources/identified_resources_pods.go
@@ -5,6 +5,7 @@ package resources
 
 import (
 	"fmt"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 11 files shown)


pkg/kapp/resources/resources.go
```diff
@@ -96,6 +96,21 @@ func (c *ResourcesImpl) All(resTypes []ResourceType, opts AllOpts) ([]Resource,
...
+	// TODO: Move to a separate function
+	if len(opts.ResourceNamespaces) > 0 {
+		exists := map[string]bool{}
+		for _, ns := range c.opts.FallbackAllowedNamespaces {
+			exists[ns] = true
+		}
+		// Populate FallbackAllowedNamespace with resource namespaces stored during deploy
+		for _, ns := range opts.ResourceNamespaces {
+			if !exists[ns] && ns != "(cluster)" {
+				c.opts.FallbackAllowedNamespaces = append(c.opts.FallbackAllowedNamespaces, ns)
```
praveenrewar: I have added the resource namespaces to the existing fallback namespaces, but I was thinking that we can actually replace it with the resource namespaces because if there is a resource in the fallback Namespace, it would already be present in the resource namespace. 
(Although it doesn't hurt to have one extra namespace)

rohitagg2020:
How are we updating the `LastChange.Namespaces`. Do we update it after every redeploy? If yes, then it will be interesting to see how things will behave when we use filter

pkg/kapp/cmd/app/deploy.go
```diff
@@ -202,7 +207,7 @@ func (o *DeployOptions) Run() error {
 	if o.DeployFlags.Logs {
 		cancelLogsCh := make(chan struct{})
 		defer func() { close(cancelLogsCh) }()
-		go o.showLogs(supportObjs.CoreClient, supportObjs.IdentifiedResources, existingPodRs, labelSelector, cancelLogsCh)
+		go o.showLogs(supportObjs.CoreClient, supportObjs.IdentifiedResources, existingPodRs, labelSelector, cancelLogsCh, append(meta.LastChange.Namespaces, nsNames...))
```
100mik, line 210: nit: split this into two lines for legibility?

pkg/kapp/resources/identified_resources_pods.go
```diff
@@ -35,31 +37,33 @@ func (w UniquePodWatcher) Watch(podsToWatchCh chan corev1.Pod, cancelCh chan str
...
+						fmt.Printf("Pod watching error: %s\n", err) // TODO
 						break
 					}
+					forbiddenNamespaces = append(forbiddenNamespaces, fmt.Sprintf(`"%s"`, namespace))
 				}
-			} else {
-				fmt.Printf("Pod watching error: %s\n", err) // TODO
+			}
+
+			if index >= len(w.fallbackAllowedNamespaces) {
```
100mik: Why is this the only breaking condition on success, do we not wanna stop watching the moment we can watch all namespaces?

100mik: Looks like we only stop if
1. There is an error which is not a forbidden error
2. We have watched *all* the fallback allowed namespaces and tried watching all namespaces as well.

pkg/kapp/resources/identified_resources_pods.go
```diff
@@ -35,31 +37,33 @@ func (w UniquePodWatcher) Watch(podsToWatchCh chan corev1.Pod, cancelCh chan str
 		// Watch Pods in all namespaces first and fallback to the
 		// fallbackAllowedNamespaces if lack of permission
 		namespace := ""
+		var index int
+		var forbiddenNamespaces []string
 		for {
```
100mik: I am thinking we could do something like :

```golang
namespaces := []string{}
namespaces = append(namespaces, ...fallbackAllowedNamespaces)
for _, ns := namespaces {
// .... truncated
// And then have a break condition like this right on the err returned by Watch()
    if err ==nil && ns == "" {
        break
    }
```

pkg/kapp/resources/resources.go
```diff
@@ -96,6 +96,9 @@ func (c *ResourcesImpl) All(resTypes []ResourceType, opts AllOpts) ([]Resource,
 		opts.ListOpts = &metav1.ListOptions{}
 	}
 
+	// Populate FallbackAllowedNamespace with resource namespaces stored during deploy
+	c.opts.FallbackAllowedNamespaces = uniqAndValidNamespaces(append(c.opts.FallbackAllowedNamespaces, opts.ResourceNamespaces...))
```
100mik, line 100: Does this not mean that `--dangerous-scope-to-fallback-allowed-namespaces` does not do anything unless only one ns is being used? Do we want it to handle multiple namespaces?

I do not recall when exactly it comes into play.

100mik (commented):
Initial thoughts

I will take a closer look at how this affects how we list resources in `allowedNamespaces` today

pkg/kapp/resources/identified_resources_pods.go
rohitagg2020: +1

pkg/kapp/resources/identified_resources_pods.go
```diff
@@ -35,31 +37,33 @@ func (w UniquePodWatcher) Watch(podsToWatchCh chan corev1.Pod, cancelCh chan str
...
+			}
+
+			if index >= len(w.fallbackAllowedNamespaces) {
 				break
 			}
+			namespace = w.fallbackAllowedNamespaces[index]
+			index++
+		}
+		if len(forbiddenNamespaces) > 0 {
+			fmt.Printf(`Pod watching error: pods is forbidden: User cannot list resource "pods" in API group "" in the namespace(s) %s`, strings.Join(forbiddenNamespaces, ", "))
```
rohitagg2020, line 65: nit: `pods is forbidden:` looks odd

100mik: +1

pkg/kapp/resources/identified_resources_pods.go
praveenrewar: Makes sense. Done.

pkg/kapp/resources/identified_resources_pods.go
praveenrewar: > do we not wanna stop watching the moment we can watch all namespaces?

Yes, good catch. Made the changes.

pkg/kapp/resources/identified_resources_pods.go
praveenrewar:
```
pods is forbidden: User "system:serviceaccount:kapp-test:scoped-sa" cannot list resource "pods" in API group "" in the namespace "kapp-test"
```
This is the error returned by the server, I have just copy pasted the format and added multiple namespaces instead of one. Open to suggestions.

pkg/kapp/resources/resources.go
praveenrewar: I am not sure what you mean here 😓

pkg/kapp/cmd/app/deploy.go
praveenrewar: Not a fan of splitting things into multiple lines, but if it makes more sense then sure :)

100mik: No strong feelings about this one, so fine by me

pkg/kapp/resources/identified_resources_pods.go
100mik: I think it is trying to say "pods" as resource rather than as a plural of pod

Fine by me

pkg/kapp/resources/resources.go
100mik: Food for thought:
When do we use whatever is happening in line 103. Is the dangerous flag only to be used when all resources belong to one namespace and we wanna scope to that?

Or do we wanna change things to accommodate multiple namespaces

praveenrewar: Made some changes to use scopeToFallbackAllowedNamespaces even when the number of namespaces is more than 1 and scope is limited to those namespaces only. 
Next step: I will try to add some e2e tests around this.

pkg/kapp/resources/resources.go
```diff
@@ -169,6 +169,24 @@ func (c *ResourcesImpl) All(resTypes []ResourceType, opts AllOpts) ([]Resource,
 				}
 			}
 
+			// In case we successfully get the list for limited scope, we will
+			// try to run namespace scoped lists for rest of the allowed namespaced
+			if resType.Namespaced() && nsScopeLimited {
+				for i := 1; i < len(c.opts.FallbackAllowedNamespaces); i++ {
```
100mik: Lets be cool kids and use `range` here?

pkg/kapp/resources/resources.go
```diff
@@ -100,7 +100,7 @@ func (c *ResourcesImpl) All(resTypes []ResourceType, opts AllOpts) ([]Resource,
 	c.opts.FallbackAllowedNamespaces = uniqAndValidNamespaces(append(c.opts.FallbackAllowedNamespaces, opts.ResourceNamespaces...))
 
 	nsScope := "" // all namespaces by default
-	nsScopeLimited := c.opts.ScopeToFallbackAllowedNamespaces && len(c.opts.FallbackAllowedNamespaces) == 1
+	nsScopeLimited := c.opts.ScopeToFallbackAllowedNamespaces && len(c.opts.FallbackAllowedNamespaces) >= 1
```
100mik: So we went with `>=1`, but wouldn't the value `o.NamespaceFlag.Name` will always be included. So the second condition would always be true?

pkg/kapp/resources/resources.go
```diff
@@ -169,6 +169,24 @@ func (c *ResourcesImpl) All(resTypes []ResourceType, opts AllOpts) ([]Resource,
 				}
 			}
 
+			// In case we successfully get the list for limited scope, we will
+			// try to run namespace scoped lists for rest of the allowed namespaced
+			if resType.Namespaced() && nsScopeLimited {
```
100mik: On a second thought, is this not exactly what `c.allForNamespaces` does? Should we refactor the code above as well to deduce as much as we can?

100mik: The case above seems to be when we do not have permission to list in all ns
And here it is because flag specifies this behaviour.

But the end result should be what `c.allForNamespaces` does in both cases I believe

pkg/kapp/resources/resources.go
100mik: And even though the condition [here](https://github.com/vmware-tanzu/carvel-kapp/pull/533/commits/3ce429f925682f2c8effe264be6fe2205ae0fc28#diff-4bf6c9b48383fc3b53344634e9527785c014d0aa647140a49c63d76bc1b22918L107-L110) is never met.

I think it should now be
```
if c.opts.ScopeToFallbackAllowedNamespaces && len(c.FallbackAllowedNamespaces) == 1
```
(Case where it makes a single list call)

100mik (commented):
Some more thoughts
Do you wanna pair on this one for a bit, it might help to go over this section I believe

pkg/kapp/resources/resources.go
praveenrewar: range would start from 0 again :(
We are not kids anymore xD xD

100mik: Oh I see 🤣 
We are dodging `""`, alrighty makes sense

praveenrewar: Well, not `""`, but the first namespace that is there in fallbackAllowedNamespaces.

100mik:
Code looks good to me, gonna take a closer look at the new test!

100mik (approved):

PR #548, opened 2022-07-03
Make TestYttWaitRulesWithUnblockChanges more certain

praveenrewar:
- Make TestYttWaitRulesWithUnblockChanges more certain

Fixes #

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/test/e2e/custom_wait_rules_test.go
+++ b/test/e2e/custom_wait_rules_test.go
@@ -127,28 +127,7 @@ func TestYttWaitRules_WithUnblockChanges(t *testing.T) {
 	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
 	kubectl := Kubectl{t, env.Namespace, logger}
 
-	yaml := `
-apiVersion: kapp.k14s.io/v1alpha1
-kind: Config
-
-waitRules:
-  - ytt:
-      funcContractV1:
-        resource.star: |
-          def is_done(resource):
-              state = resource.status.currentState
-              if state == "Progressing":
-                return {"done": False, "unblockChanges": True, "message": "Unblock blocked changes"}
-              elif state == "Running":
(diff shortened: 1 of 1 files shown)


test/e2e/custom_wait_rules_test.go
```diff
@@ -223,39 +225,36 @@ metadata:
...
 
 		expectedOutput := strings.TrimSpace(replaceSpaces(`Changes
 
-Namespace  Name                         Kind                      Age  Op      Op st.  Wait to    Rs  Ri  $
-(cluster)  crontabs.stable.example.com  CustomResourceDefinition  -    create  -       reconcile  -   -  $
-kapp-test  my-new-cron-object-1         CronTab                   -    create  -       reconcile  -   -  $
-^          test-cm                      ConfigMap                 -    create  -       reconcile  -   -  $
+Namespace  Name                  Kind       Age  Op      Op st.  Wait to    Rs  Ri  $
+kapp-test  my-new-cron-object-1  CronTab    -    create  -       reconcile  -   -  $
+^          test-cm               ConfigMap  -    create  -       reconcile  -   -  $
```
cppforlife, line 240: why did we need to split into 2 kapp deploys? will users using kapp have to split things as well when encountering similar cases?

praveenrewar: Sometimes while deploying the crontab CRD, we get "Condition Established is not set" before the CRD is successfully created. This messes up the expected out. Since the main concern for this test case is the CR and the ConfigMap (CR has the wait rules defined and the ConfigMap is unblocked when it's in the Progressing state), I moved the CRD to a separate app to have a deterministic expected output.

cppforlife: i see.

PR #558, opened 2022-07-20
updated errors wrapping to use %w instead of %s 

### LINKED ISSUE #498: kapp public API should not hide error types
**What steps did you take:**

As a codebase which uses the public kapp go API, we call the public `RecordedApp.UsedGVs()` which in turn calls the private `RecordedApp.meta()` which in turn uses the k8s API to fetch the config map:

https://github.com/vmware-tanzu/carvel-kapp/blob/3486b0b61553e4cd1ebfc2fed3f95d81c1f65d10/pkg/kapp/app/recorded_app.go#L281-L287

So the code neatly checks if the error is a NotFound error, but rather than wrapping the actual error (which would not lose information), it returns a new error which loses this information, which is returned up the chain to the response to the public `UsedGVs()` call.

**What happened:**

In our code, (before digging into the kapp source) I tried to do:

```golang
if errors.IsNotFound(err) { ... }
```
which will happily work with wrapped errors. But because the returned error is not a wrapped NotFound, it obviously does not detect it.

**What did you expect:**

To be able to use `errors.IsNotFound()` on the returned error. Instead, we need to do some string matching:

```golang
if strings.Contains(err.Error(), fmt.Sprintf("configmaps %q not found", appName)) { ... }
```

Guessing this code was just written before the go `errors` api was released, so it may be similar throughout, but should be simple enough to change to use `%w` instead of `%s` in those error strings.

kumaritanushree:
updated errors wrapping to use %w instead of %s so that errors will not loose info like error type.

Fixes #498 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -75,7 +75,6 @@ func (a *RecordedApp) LabelSelector() (labels.Selector, error) {
 func (a *RecordedApp) UsedGVs() ([]schema.GroupVersion, error) {
 	meta, err := a.meta()
 	if err != nil {
-		fmt.Printf("Error-1: %s\n", err.Error())
 		return nil, err
 	}
 
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -75,6 +75,7 @@ func (a *RecordedApp) LabelSelector() (labels.Selector, error) {
 func (a *RecordedApp) UsedGVs() ([]schema.GroupVersion, error) {
 	meta, err := a.meta()
 	if err != nil {
+		fmt.Printf("Error-1: %s\n", err.Error())
 		return nil, err
 	}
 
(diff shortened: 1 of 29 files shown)


pkg/kapp/app/recorded_app.go
```diff
@@ -75,6 +75,7 @@ func (a *RecordedApp) LabelSelector() (labels.Selector, error) {
 func (a *RecordedApp) UsedGVs() ([]schema.GroupVersion, error) {
 	meta, err := a.meta()
 	if err != nil {
+		fmt.Printf("Error-1: %s\n", err.Error())
```
praveenrewar: nit: Remove Print used for testing.

kumaritanushree: removed

sethiyash (approved):
Looks Good.

PR #560, opened 2022-07-26
Add `--appMetadataFile` deploy flag

DennisDenuto:
TODO: 
- [x] remove any hardcoding left over
- [x] fix up commits

related to: https://github.com/vmware-tanzu/carvel-kapp-controller/pull/799

Yes, it introduces a new flag to write the app metadata config map to disk

```release-note
- Adds deploy flag `--appMetadataFile` to write an app metadata file to disk
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -5,6 +5,7 @@ package app
 
 import (
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 
@@ -225,8 +226,24 @@ func (o *DeployOptions) Run() error {
 		}
 
 		// Remove unused GVs and GKs
-		return app.UpdateUsedGVsAndGKs(failingAPIServicesPolicy.GVs(newResources, nil),
+		err = app.UpdateUsedGVsAndGKs(failingAPIServicesPolicy.GVs(newResources, nil),
 			NewUsedGKsScope(newResources).GKs())
+		if err != nil {
+			return err
+		}
+
+		if o.DeployFlags.AppMetadataFile != "" {
+			meta, err := app.Meta()
+			if err != nil {
+				return err
+			}
+			err = os.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)
+			if err != nil {
+				return err
+			}
+		}
+
+		return err
 	})
 	if err != nil {
 		return err
--- a/pkg/kapp/cmd/app/deploy_flags.go
+++ b/pkg/kapp/cmd/app/deploy_flags.go
@@ -21,8 +21,9 @@ type DeployFlags struct {
 
 	AppChangesMaxToKeep int
 
-	Logs    bool
-	LogsAll bool
+	Logs            bool
+	LogsAll         bool
+	AppMetadataFile string
 }
 
 func (s *DeployFlags) Set(cmd *cobra.Command) {
@@ -48,4 +49,5 @@ func (s *DeployFlags) Set(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&s.Logs, "logs", true, fmt.Sprintf("Show logs from Pods annotated as '%s'", deployLogsAnnKey))
 	cmd.Flags().BoolVar(&s.LogsAll, "logs-all", false, "Show logs from all Pods")
+	cmd.Flags().StringVar(&s.AppMetadataFile, "app-metadata-file-output", "", "Set filename to write app metadata")
 }
--- a/test/e2e/app_change_test.go
+++ b/test/e2e/app_change_test.go
@@ -5,12 +5,15 @@ package e2e
 
 import (
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 	"time"
 
 	uitest "github.com/cppforlife/go-cli-ui/ui/test"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
 )
 
 func TestAppChange(t *testing.T) {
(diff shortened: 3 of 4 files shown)


pkg/kapp/cmd/app/deploy_flags.go
```diff
@@ -48,4 +49,5 @@ func (s *DeployFlags) Set(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&s.Logs, "logs", true, fmt.Sprintf("Show logs from Pods annotated as '%s'", deployLogsAnnKey))
 	cmd.Flags().BoolVar(&s.LogsAll, "logs-all", false, "Show logs from all Pods")
+	cmd.Flags().StringVar(&s.AppMetadataFile, "appMetadataFile", "", "write app metatadata to file")
```
praveenrewar: `app-metadata-file`?

praveenrewar (commented):
> a new flag to write the app metadata config map to disk

Nice!! I am curious to know the use case for this :)

*Update*: Never mind, just noticed https://github.com/vmware-tanzu/carvel-kapp-controller/pull/799

pkg/kapp/cmd/app/deploy_flags.go
DennisDenuto: yeah good suggestion! i'll change it now

pkg/kapp/cmd/app/deploy.go
```diff
@@ -225,8 +227,25 @@ func (o *DeployOptions) Run() error {
...
+		if err != nil {
+			return err
+		}
+
+		if o.DeployFlags.AppMetadataFile != "" {
+			meta, err := app.Meta()
+			if err != nil {
+				return err
+			}
+			err = ioutil.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)
```
praveenrewar: Maybe we should directly use `os.WriteFile()`? (Since that's just what `ioutil.WriteFile()` is doing)
Similar in e2e test.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -225,8 +227,25 @@ func (o *DeployOptions) Run() error {
...
+			if err != nil {
+				return err
+			}
```
praveenrewar, line 243: Should we just return the err at the end instead of nil? (No strong opinions here as handling the err more closer to it’s  occurrence seems more readable.)

pkg/kapp/cmd/app/deploy.go
```diff
@@ -225,8 +227,25 @@ func (o *DeployOptions) Run() error {
...
+		}
+
+		return nil
+
```
praveenrewar: nit: remove newline

test/e2e/app_change_test.go
```diff
@@ -89,3 +93,112 @@ spec:
...
+
+	name := "test-app-change"
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	firstDeploy, err := ioutil.TempFile(os.TempDir(), "output1")
```
praveenrewar: Same as above, we could maybe use `os.CreateTemp` instead?

pkg/kapp/cmd/app/deploy_flags.go
```diff
@@ -48,4 +49,5 @@ func (s *DeployFlags) Set(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&s.Logs, "logs", true, fmt.Sprintf("Show logs from Pods annotated as '%s'", deployLogsAnnKey))
 	cmd.Flags().BoolVar(&s.LogsAll, "logs-all", false, "Show logs from all Pods")
+	cmd.Flags().StringVar(&s.AppMetadataFile, "app-metadata-file", "", "write app metatadata to file")
```
praveenrewar: nit:
```suggestion
	cmd.Flags().StringVar(&s.AppMetadataFile, "app-metadata-file", "", "Write app metadata to file")
```

Also, maybe we would want to suggest that it accepts a filename as input? Something like "Set filename to write app metadata"?

test/e2e/app_change_test.go
```diff
@@ -89,3 +93,112 @@ spec:
 		require.Equal(t, 2, len(resp2.Tables[0].Rows), "Expected to have 2 app-changes")
 	})
 }
+
+func TestAppKindChangeWithMetadataOutput(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+
+	yaml1 := `
```
100mik, line 101: Not a blocker per say, but in interest of keeping the tests minimal 🤔 

We could have a ConfigMap and a Secret instead of a Deployment and a Service right? 
I think it would be less verbose and probably make the test faster

test/e2e/app_change_test.go
```diff
@@ -89,3 +93,86 @@ spec:
...
+
+	logger.Section("deploy app", func() {
+		kapp.RunWithOpts([]string{"deploy", "--app-metadata-file", firstDeploy.Name(), "-f", "-", "-a", name}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
+	})
+
+	logger.Section("deploy with changes", func() {
+		kapp.RunWithOpts([]string{"deploy", "--app-metadata-file", secondDeploy.Name(), "-f", "-", "-a", name}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml2)})
+	})
+
+	configMapFirstDeploy, err := ioutil.ReadFile(firstDeploy.Name())
```
praveenrewar: (I think I forgot to add a comment about this one and at line 158)
*Suggestion*: Use `os.ReadFile()` instead

pkg/kapp/cmd/app/deploy.go
DennisDenuto: sgtm! returned an err instead of nil at the end

praveenrewar: hehe, I actually meant that we can remove the err check and just return err at the end instead of nil. 
```go
			err = os.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)
		}
		return err
```
But I am okay with the current way as well because err should be handled closer to it’s  origin.

praveenrewar (approved):
LGTM!

100mik (approved):
Looks good!

test/e2e/app_change_test.go
```diff
@@ -5,6 +5,9 @@ package e2e
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v2"
```
cppforlife: nit: external imports in separate section

pkg/kapp/cmd/app/deploy_flags.go
```diff
@@ -48,4 +49,5 @@ func (s *DeployFlags) Set(cmd *cobra.Command) {
 
 	cmd.Flags().BoolVar(&s.Logs, "logs", true, fmt.Sprintf("Show logs from Pods annotated as '%s'", deployLogsAnnKey))
 	cmd.Flags().BoolVar(&s.LogsAll, "logs-all", false, "Show logs from all Pods")
+	cmd.Flags().StringVar(&s.AppMetadataFile, "app-metadata-file", "", "Set filename to write app metadata")
```
cppforlife: @praveenrewar @100mik is the plan here would also be to read from such file in future? if not, we should probably follow carvel convention of having `-output` suffix.

praveenrewar: > is the plan here would also be to read from such file in future?

I can't think of any reason we would want to read from the file instead of the ConfigMap. So yeah, we should probably add the `-output` suffix.

DennisDenuto: sgtm; added `-output` suffix to the flag name

sethiyash (approved):
LGTM!

PR #565, opened 2022-08-01
Add flag to use default label scoping rules

### LINKED ISSUE #563: Add flag to disable default label scoping rules
### Discussed in https://github.com/vmware-tanzu/carvel-kapp/discussions/212

<div type='discussions-op-text'>

<sup>Originally posted by **cppforlife** April 17, 2021</sup>
Currently it's possible to disable label scoping feature per resource via annotation. We've seen that when users are migrating and they have existing Deployments it results in an annoying error for changing immutable field (spec.selector). We should figure out how to disable this feature.

we cannot simply change the kapp's default since we already have existing deploys out there (this would be a breaking change). one idea is to disable it for new deployments; however, that may not be great either if users are expecting this behavior for their new deployments as well (e.g. in CI you deploy test deployment every time).

what can we do to improve experience?</div>

praveenrewar:
Add flag to use default label scoping rules.

Fixes #563 

```release-note
Added --default-label-scoping-rules flag which can be used to enable or disable the default label scoping rules added during deploy.
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy_flags.go
+++ b/pkg/kapp/cmd/app/deploy_flags.go
@@ -21,6 +21,8 @@ type DeployFlags struct {
 
 	AppChangesMaxToKeep int
 
+	DefaultLabelScopingRules bool
+
 	Logs            bool
 	LogsAll         bool
 	AppMetadataFile string
@@ -44,6 +46,9 @@ func (s *DeployFlags) Set(cmd *cobra.Command) {
 		100, "Concurrency to check for existing non-labeled resources")
 	cmd.Flags().BoolVar(&s.OverrideOwnershipOfExistingResources, "dangerous-override-ownership-of-existing-resources",
 		false, "Steal existing resources from another app")
+
+	cmd.Flags().BoolVar(&s.DefaultLabelScopingRules, "default-label-scoping-rules",
+		true, "Use default label scoping rules")
 
 	cmd.Flags().IntVar(&s.AppChangesMaxToKeep, "app-changes-max-to-keep", ctlapp.AppChangesMaxToKeepDefault, "Maximum number of app changes to keep")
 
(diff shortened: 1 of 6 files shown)


pkg/kapp/config/conf.go
```diff
@@ -122,11 +122,14 @@ func (c Conf) WaitRules() []WaitRule {
 	return rules
 }
 
-func (c Conf) LabelScopingMods() func(kvs map[string]string) []ctlres.StringMapAppendMod {
+func (c Conf) LabelScopingMods(disableDefault bool) func(kvs map[string]string) []ctlres.StringMapAppendMod {
 	return func(kvs map[string]string) []ctlres.StringMapAppendMod {
 		var mods []ctlres.StringMapAppendMod
 		for _, config := range c.configs {
 			for _, rule := range config.LabelScopingRules {
+				if disableDefault && rule.IsDefault {
```
rohitagg2020: nit: Change name to `disableDefaultRules` instead of `disableDefault`

praveenrewar: Sure 😃

pkg/kapp/cmd/app/deploy_flags.go
```diff
@@ -45,6 +47,9 @@ func (s *DeployFlags) Set(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&s.OverrideOwnershipOfExistingResources, "dangerous-override-ownership-of-existing-resources",
 		false, "Steal existing resources from another app")
 
+	cmd.Flags().BoolVar(&s.DisableDefaultLabelScopingRules, "disable-default-label-scoping-rules",
```
cppforlife: not sure if we have other similar flags for tweaking config... may be default-label-scoping-rules=true (as a default)

praveenrewar: > not sure if we have other similar flags for tweaking config

Yeah, I don't think we have any other similar flag.

> may be default-label-scoping-rules=true (as a default)

Definitely make more sense. I will make the change.

PR #571, opened 2022-08-04
changes to update resources based on predetermined interval

### LINKED ISSUE #224: kapp deploy triggers a new versioned asset based on a predetermined interval
We currently use kapp versioning capability for several Custom Resources beyond just ConfigMap and Secrets, this is very easily done with [`templateRules`](https://github.com/voor/cluster-api-gitops/blob/default/management-cluster/workload/manifests/kapp-config.yaml#L4-L21) and allows for some super powerful capabilities like greatly reducing the complexity of how we roll out new AMIs with Cluster API.  

One of the things the Cluster API AWS team is looking to do specifically in Cluster API is turn on some more deterministic lookups of AMIs, which would help us with no longer needing to populate the AMI ID field at all in our `AWSMachineTemplate` (see example [here](https://github.com/voor/cluster-api-gitops/blob/default/management-cluster/workload/manifests/overlays/cluster-upgrade-overlay.yaml#L18) that means effectively this file *never mutates anymore*, but we still _want it_ to mutate periodically to invoke a roll-out of the machines.

If there was some kapp annotation that looked like a cron job or had a similar behavior, like:
```
#@overlay/match by=overlay.subset({"kind":"AWSMachineTemplate"}),expects="0+"
---
metadata:
  #@overlay/match missing_ok=True
  annotations:
    kapp.k14s.io/versioned-cron-schedule: "0 0 * * 0"
    kapp.k14s.io/versioned: ""
    kapp.k14s.io/num-versions: "10"
```
with a result something like this (note the `kapp.k14s.io/versioned-cron-schedule-last-run`):
```
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
kind: AWSMachineTemplate
metadata:
  annotations:
    kapp.k14s.io/delete-strategy: orphan
    kapp.k14s.io/identity: ...
    kapp.k14s.io/num-versions: "10"
    kapp.k14s.io/original: '...'
    kapp.k14s.io/original-diff-md5: 58e0494c51d30eb3494f7c9198986bb9
    kapp.k14s.io/versioned: ""
    kapp.k14s.io/versioned-cron-schedule: "0 0 * * 0"
    kapp.k14s.io/versioned-cron-schedule-last-run: "2021-06-09T14:24:30Z"
  creationTimestamp: "2021-06-09T14:24:30Z"
  name: my-custom-resource-ver-9
```

That would just trigger a new versioning on a set interval, the timing would be dependent on kapp deploy being executed again. 
 Basically if it has been +1 week since this kapp deploy was last run based on the "last run" in the annotation then `my-custom-resource-ver-10` would be created.

* This is designed to be run through kapp-controller, so the periodic running of kapp is implicitly provided.
* The other application not mentioned explicitly in this issue is cert-manager refreshing certificates without updating the associated secret, by triggering a new Certificate periodically *before* cert-manager refreshes a certificate you can pre-empt certificate refreshes.

---
Vote on this request

kumaritanushree:
Kapp will update resources based on predetermined interval set. To set the interval user needs to add annotation kapp.k14s.io/renew-duration to the resource

Fixes #224 

```release-note
New annotation has been introduced called kapp.k14s.io/renew-duration to set the duration at which user want to renew their resources.
```

##### Review Checklist:

  a link to that PR
  change

[Proposal](https://hackmd.io/Es98h5LuQCq_VtTZTKAo_g?view)


pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -25,6 +25,8 @@ const (
 	updateStrategyFallbackOnReplaceAnnValue ClusterChangeApplyStrategyOp = "fallback-on-replace"
 	updateStrategyAlwaysReplaceAnnValue     ClusterChangeApplyStrategyOp = "always-replace"
 	updateStrategySkipAnnValue              ClusterChangeApplyStrategyOp = "skip"
+
+	maxDuration = "kapp.k14s.io/max-duration"
```
praveenrewar: `maxDuration` -> `maxDurationAnnKey`

pkg/kapp/cmd/app/deploy.go
```diff
@@ -29,7 +30,10 @@ import (
 )
 
 const (
-	TTYByDefaultKey = "cli.carvel.dev/tty-by-default"
+	TTYByDefaultKey    = "cli.carvel.dev/tty-by-default"
+	versionedResAnnKey = "kapp.k14s.io/versioned"
+	maxDuration        = "kapp.k14s.io/max-duration"
+	lastRenewTime      = "kapp.k14s.io/last-renewed-time"
```
praveenrewar: We already have another declaration for maxDuration, should we keep it at one place only and reuse it?
`lastRenewTime` -> `lastRenewTimeAnnKey`

test/e2e/priodic_update_test.go
```diff
@@ -0,0 +1,97 @@
...
+      6 +     kapp.k14s.io/last-renewed-time: "2006-01-02T15:04:05Z07:00"
+  6,  7       kapp.k14s.io/max-duration: 4s
+  7,  8     creationTimestamp: "2006-01-02T15:04:05Z07:00"
+`
+
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		out = replaceString(out)
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		require.Contains(t, out, expectedOutput, "output does not match")
+	})
```
praveenrewar: I tried adding a deploy after a sleep of 10s, but that didn't trigger any update 🤔 The resources should be recreated again after an interval of 4s, right?

test/e2e/priodic_update_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPriodicUpdate(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+	kubectl := Kubectl{t, env.Namespace, logger}
+	//fieldsExcludedInMatch := []string{"kapp.k14s.io/app", "creationTimestamp:", "resourceVersion:", "uid:", "selfLink:", "kapp.k14s.io/association"}
```
praveenrewar: nit: remove commented code.

test/e2e/priodic_update_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		out = replaceString(out)
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		require.Contains(t, out, expectedOutput, "output does not match")
+	})
+
+}
+
+// clearKeys will remove all matching fields in fieldsExcludedInMatch from out
```
praveenrewar: nit: comment seems to be copied from another function

test/e2e/priodic_update_test.go
```diff
@@ -0,0 +1,97 @@
...
+func replaceString(out string) string {
```
praveenrewar: nit: function name could be a little bit more descriptive, like what exactly is being replaced?

praveenrewar (commented):
I will take another look at the implementation in some time.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -448,6 +457,72 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
 	return clusterChangeSet, clusterChangesGraph, (len(clusterChanges) == 0), changesSummary, err
 }
 
+func CheckForPriodicAnnotation(existingResources, newResources []ctlres.Resource) error {
+
```
sethiyash: nit: you can remove the empty line here and everywhere else we usually don't have an empty line after function

pkg/kapp/cmd/app/deploy.go
```diff
@@ -448,6 +457,72 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
...
+			if found && time.Now().After(rs.CreatedAt().Add(duration)) {
+				err = addNonceMod.Apply(res)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+
```
sethiyash: nit: remove an empty line

test/e2e/priodic_update_test.go
praveenrewar: Resolving this as it happened because of wrong test name used.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -448,6 +457,72 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
...
+	for _, res := range newResources {
+		rs, found := existRes[res.Name()+res.Kind()]
+		if found {
+			val, found := rs.Annotations()[maxDuration]
+			if !found {
+				continue
+			}
+
+			duration, err = convertToTimeDuration(val)
```
praveenrewar: Maybe directly use `time.ParseDuration()`? I don't see this function doing anything additional to that.

pkg/kapp/cmd/app/deploy.go
praveenrewar: > We already have another declaration for maxDuration, should we keep it at one place only and reuse it?

Same goes for `versionedResAnnKey`

pkg/kapp/cmd/app/deploy.go
```diff
@@ -400,6 +404,11 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
 		changeFactory := ctldiff.NewChangeFactory(conf.RebaseMods(), conf.DiffAgainstLastAppliedFieldExclusionMods())
 		changeSetFactory := ctldiff.NewChangeSetFactory(o.DiffFlags.ChangeSetOpts, changeFactory)
 
+		err := CheckForPriodicAnnotation(existingResources, newResources)
```
praveenrewar: Couple of points here:
- The function name suggests that we are checking for `PriodicAnnotation`, but we are also making some updates to the resources based on the presence of that annotation, so the name should signify that update as well.
- It seems a bit weird to have this logic present here. It seems more like a _preparation_ we would want to do before calculating the changes, so maybe it should be present in the diff package?

pkg/kapp/cmd/app/deploy.go
```diff
@@ -448,6 +457,72 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
...
+			lastRenewTime: fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339)),
+		},
+	}
+	existRes := existingPriodicResource(existingResources)
+
+	for _, res := range newResources {
+		rs, found := existRes[res.Name()+res.Kind()]
+		if found {
+			val, found := rs.Annotations()[maxDuration]
+			if !found {
```
praveenrewar: What happens if the the max-duration annotation is added to an already existing resource for which the max-duration has passed since it was created? 
What happens if the annotation is removed for an existing resource with the annotation present?

kumaritanushree: - If we will add max-duration to an existing resource and condition (creationTimestamp + max-duration < current time) has passed then for versioned resource, new resource will get created while for non-versioned resource, resource will get updated with updated strategy always-replace.
-  If we will remove max-duration from an existing resource then normal update will happen. 

Added e2e test case for the same.

pkg/kapp/config/default.go
```diff
@@ -130,6 +130,12 @@ rebaseRules:
   resourceMatchers:
   - apiVersionKindMatcher: {apiVersion: v1, kind: Pod}
 
+- path: [metadata, annotations, "kapp.k14s.io/last-renewed-time"]
+  type: copy
+  sources: [new, existing]
```
rohitagg2020: Do we see this annotation being set by the user? If yes, in what scenarios. If no, then we can always do `sources: [existing]`

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,71 @@
...
+	)
+
+	addNonceMod := ctlres.StringMapAppendMod{
+		ResourceMatcher: ctlres.AllMatcher{},
+		Path:            ctlres.NewPathFromStrings([]string{"metadata", "annotations"}),
+		KVs: map[string]string{
+			lastRenewTimeAnnKey: fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339)),
+		},
+	}
+	existRes := existingPeriodicResource(existingResources)
```
rohitagg2020: nit: rename to existResMap

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,71 @@
...
+			if !found {
+				continue
+			}
+
+			duration, err = time.ParseDuration(val)
+			if err != nil {
+				return err
+			}
+
+			if found && time.Now().After(rs.CreatedAt().Add(duration)) {
```
rohitagg2020: We can remove `found` from here. Also may be we can rewrite it as:
```
for _, res := range newResources {
    rs, found := existRes[res.Name()+res.Kind()]
        if found {
            val, exist := rs.Annotations()[MaxDurationAnnKey]
            if exist {
                duration, err = time.ParseDuration(val)
                if err != nil {
                return err
                }

                if time.Now().After(rs.CreatedAt().Add(duration)) {
                    err = addNonceMod.Apply(res)
                    if err != nil {
                        return err
                    }
                }
            }
        }
    }
}
```

pkg/kapp/clusterapply/add_or_update_change.go
```diff
@@ -69,6 +69,11 @@ func (c AddOrUpdateChange) ApplyStrategy() (ApplyStrategy, error) {
 	case ctldiff.ChangeOpUpdate:
 		newRes := c.change.NewResource()
 
+		_, found := newRes.Annotations()[ctldiff.MaxDurationAnnKey]
+		if found {
+			return UpdateAlwaysReplaceStrategy{c}, nil
```
100mik: Why should it always pick this strategy implicitly if it has the annotation?

Won't the diff stage pick up the update since the resource is modified in the diff stage?

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+)
+
+const (
+	MaxDurationAnnKey   = "kapp.k14s.io/max-duration"
```
100mik: This would become a private variable if we stop using this in the apply stage.

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,70 @@
...
+	lastRenewTimeAnnKey = "kapp.k14s.io/last-renewed-time"
+)
+
+func CheckAndCalculateResWithPeriodicAnn(existingResources, newResources []ctlres.Resource) error {
+	var (
+		duration time.Duration
+		err      error
+	)
+
+	addNonceMod := ctlres.StringMapAppendMod{
```
100mik: Should we call this something else as Nonce is a separate feature?

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,70 @@
...
+		ResourceMatcher: ctlres.AllMatcher{},
+		Path:            ctlres.NewPathFromStrings([]string{"metadata", "annotations"}),
+		KVs: map[string]string{
+			lastRenewTimeAnnKey: fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339)),
+		},
+	}
+	existRes := existResMap(existingResources)
+
+	for _, res := range newResources {
+		rs, found := existRes[res.Name()+res.Kind()]
```
100mik: nit: rs => existingRes

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,70 @@
...
+	)
+
+	addNonceMod := ctlres.StringMapAppendMod{
+		ResourceMatcher: ctlres.AllMatcher{},
+		Path:            ctlres.NewPathFromStrings([]string{"metadata", "annotations"}),
+		KVs: map[string]string{
+			lastRenewTimeAnnKey: fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339)),
+		},
+	}
+	existRes := existResMap(existingResources)
```
100mik: nit: existRes => existingResourceMap

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,70 @@
...
+
+	ctlres "github.com/k14s/kapp/pkg/kapp/resources"
+)
+
+const (
+	MaxDurationAnnKey   = "kapp.k14s.io/max-duration"
+	lastRenewTimeAnnKey = "kapp.k14s.io/last-renewed-time"
+)
+
+func CheckAndCalculateResWithPeriodicAnn(existingResources, newResources []ctlres.Resource) error {
```
100mik: Let's scope this to a struct rather than make this an exported function?

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return nil
+}
+
+func existResMap(existingResources []ctlres.Resource) map[string]ctlres.Resource {
+	exRes := make(map[string]ctlres.Resource)
+	for _, rs := range existingResources {
+		if _, found := rs.Annotations()[MaxDurationAnnKey]; found {
+			if _, found := rs.Annotations()[versionedResAnnKey]; found {
+				name := strings.Split(rs.Name(), "-ver-")[0]
```
100mik: Can we reuse existing logic defined by versioned resources instead of duplicating it?

pkg/kapp/clusterapply/add_or_update_change.go
kumaritanushree: Diff stage will pick the update and update the resource with the update strategy which is already set for that particular resource. For an example, for resource with plain update strategy kapp will do simple update for that resource and in simple update resource's creationTimestamp does not change. Hence our required condition (creationTimestap + max-duration < current time) will always pass after that. 

To change the creationTimestamp on every update we need to delete and create the resource and for that we have to use update strategy as always-replace.

pkg/kapp/diff/prepare_res_with_periodic_ann.go
kumaritanushree: Same reason as above. This ann is needed in apply stage to update update-strategy for resources having MaxDurationAnnKey set.

pkg/kapp/clusterapply/add_or_update_change.go
100mik: Lets depend on the added timestamp as well and not always pick this strategy

pkg/kapp/diff/prepare_res_with_periodic_ann.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+	}
+	existRes := existResMap(existingResources)
+
+	for _, res := range newResources {
+		rs, found := existRes[res.Name()+res.Kind()]
+		if found {
+			val, exist := rs.Annotations()[MaxDurationAnnKey]
+			if exist {
+				duration, err = time.ParseDuration(val)
```
100mik: We should use `res` as opposed to `val` here

Something @cppforlife pointed out ^

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +85,50 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	existingNonVerRsGrouped map[string]ctlres.Resource) error {
+
+	var err error
+	for _, res := range d.newRs {
+		if _, found := res.Annotations()[MaxDurationAnnKey]; found {
+			resKey := VersionedResource{res, nil}.UniqVersionedKey().String()
+			if exRes, found := existingNonVerRsGrouped[resKey]; found {
+				err = updateLastRenewedTime(res, exRes)
+			} else if exRes, found := existingRsGrouped[resKey]; found {
+				err = updateLastRenewedTime(res, exRes[len(exRes)-1])
```
rohitagg2020: nit: adding a comment will make it more clear why we are doing exRes[len(exRes)-1]

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +85,50 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+			} else {
+				return nil
```
rohitagg2020: Is it needed?

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +85,50 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+			}
+		}
+	}
+	return err
```
rohitagg2020: Seems like we are going to override the error while iterating. Imagine a scenario: total 3 iteration of for loop. Error came in second only. In that case, we will send nil bcoz third iteration, err is nil. I am not clear What needs to be done for 2nd iteration resource? should we fail or let the program proceed?

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +85,50 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	}
+
+	val := exRes.Annotations()[MaxDurationAnnKey]
+	duration, err := time.ParseDuration(val)
+	if err != nil {
+		return err
+	}
+
+	if time.Now().After(exRes.CreatedAt().Add(duration)) {
+		err := updateLastRenewAnn.Apply(res)
```
rohitagg2020: return `updateLastRenewAnn.Apply(res)`

sethiyash: +1

test/e2e/priodic_update_test.go
```diff
@@ -0,0 +1,163 @@
...
+import (
+	"regexp"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPeriodicUpdate(t *testing.T) {
```
sethiyash: Should we also add a test case here when the resource is non-versioned and the `kapp.k14s.io/max-duration` annotation is present but the `update-strategy` is not `always-replace` to validate we are throwing the error in that scenario? 
(Just a thought not a blocker)

pkg/kapp/diff/change_set_with_versioned_rs.go
kumaritanushree: As soon as we encountered error for any resource we will throw an error and exit the program. Will correct the logic as well.

test/e2e/priodic_update_test.go
kumaritanushree: added

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -16,6 +17,9 @@ const (
 	versionedResAnnKey        = "kapp.k14s.io/versioned"               // Value is ignored
 	versionedResOrigAnnKey    = "kapp.k14s.io/versioned-keep-original" // Value is ignored
 	versionedResNumVersAnnKey = "kapp.k14s.io/num-versions"
+	maxDurationAnnKey         = "kapp.k14s.io/max-duration"
+	lastRenewTimeAnnKey       = "kapp.k14s.io/last-renewed-time"
+	updateStrategyAnnKey      = "kapp.k14s.io/update-strategy"
```
sethiyash: Please remove this as it is not getting used now and golint is also failing due to this only.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -35,14 +39,20 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
 	existingRs := existingVersionedResources(d.existingRs)
 	existingRsGrouped := d.groupResources(existingRs.Versioned)
 
+	existingNonVerRsGrouped := d.groupNonVerResources(existingRs.NonVersioned)
```
praveenrewar: We are not grouping resources here, we are just creating a map of resource id and resource, maybe we should just do it inline in the `checkForMaxDuration` function directly?

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -35,14 +39,20 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	err := d.checkForMaxDurationAnn(existingRsGrouped, existingNonVerRsGrouped)
```
praveenrewar: `checkForMaxDurationAnn`: Hmm, I think we are doing a lot more than just checking for the max-duration annotation. I can't think of a name right now that would describe the actual functionality over here, but let's try to see if we can find something better.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +86,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	existingNonVerRsGrouped map[string]ctlres.Resource) error {
+
+	for _, res := range d.newRs {
+		if val, found := res.Annotations()[maxDurationAnnKey]; found {
+
+			// Validate value of annotation `kapp.k14s.io/max-duration`.
+			duration, err := time.ParseDuration(val)
+			if err != nil {
+				return fmt.Errorf("%s for resource: %s", err.Error(), res.Description())
+			}
```
praveenrewar: nit: remove the newline before the comment
I don't think we need this comment, instead let's try to make the error a bit more clear, so that it's easier to understand for both users and readers.

Something like 
```go
return fmt.Errorf("Unable to parse max-duration for resource %s": %s, res.Description(), err.Error())
```

pkg/kapp/diff/prepare_res_with_periodic_ann.go
kumaritanushree: Closing this and all other outdated comments as changes has been updated with new implementation logic.

pkg/kapp/diff/change_set_with_versioned_rs.go
kumaritanushree: removed it.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +86,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+					return err
+				}
+			} else if exRes, found := existingRsGrouped[resKey]; found {
+				// exRes contains list of all versioned resources currently present on the cluster for particular resource and are sorted
+				// here we will be using last created resource.
+				err = updateLastRenewedTime(res, exRes[len(exRes)-1], duration)
+				if err != nil {
+					return err
+				}
+			}
```
praveenrewar: I think we can just get the exRes in the conditions and call the `updateLastRenewedTime` after that? 
Something like this:
```go
                         exRes, found := existingNonVerRsGrouped[resKey]
                         if !found {
                                 exResGroup, found := existingRsGrouped[resKey]
                                 exRes = exResGroup[len(exResGroup)-1]
                          }
                          err = updateLastRenewedTime(res, exRes, duration)
		          if err != nil {
				 return err
		          }
```

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +86,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	return nil
+}
+
+func updateLastRenewedTime(res, exRes ctlres.Resource, duration time.Duration) error {
+	var (
+		lastRenewed time.Time
+		err         error
+	)
+
+	val := exRes.Annotations()[lastRenewTimeAnnKey]
```
praveenrewar: Maybe we can extract this in `checkForDuration` itself? Then we wont have to pass on the exRes.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +86,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	if val != "" {
+		lastRenewed, err = time.Parse(time.RFC3339, val)
+		if err != nil {
+			return fmt.Errorf("%s for resource: %s", err.Error(), res.Description())
```
praveenrewar: Let's use a more informative error message similar to one above.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +86,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+
+func addOrUpdateLastRenewedAnn(res ctlres.Resource) error {
+	lastRenewAnn := ctlres.StringMapAppendMod{
+		ResourceMatcher: ctlres.AllMatcher{},
+		Path:            ctlres.NewPathFromStrings([]string{"metadata", "annotations"}),
+		KVs: map[string]string{
+			lastRenewTimeAnnKey: fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339)),
+		},
+	}
+	return lastRenewAnn.Apply(res)
```
praveenrewar: I think we can do this inline in `updateLastRenewedTime` itself.
```suggestion
	return ctlres.StringMapAppendMod{
		ResourceMatcher: ctlres.AllMatcher{},
		Path:            ctlres.NewPathFromStrings([]string{"metadata", "annotations"}),
		KVs: map[string]string{
			lastRenewTimeAnnKey: fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339)),
		},
	}.Apply(res)
```

pkg/kapp/config/default.go
```diff
@@ -48,6 +48,12 @@ rebaseRules:
   resourceMatchers:
   - allMatcher: {}
 
+- path: [metadata, annotations, "kapp.k14s.io/last-renewed-time"]
+  type: copy
+  sources: [new, existing]
```
rohitagg2020: If this is not suppose to be set by user, then should we make it `sources: [existing]`

pkg/kapp/diff/change_set_with_versioned_rs.go
kumaritanushree: How about -> checkAndUpdateResWithMaxDurationAnn ?

pkg/kapp/diff/change_set_with_versioned_rs.go
kumaritanushree: updated

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +84,76 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func updateLastRenewedTime(res ctlres.Resource, duration time.Duration, lastRenewed, creationTime time.Time) error {
+
+	if creationTime.Before(lastRenewed) {
```
rohitagg2020: `creationTime` will be always before `lastRenewed` in case of subsequent update scenario. Am I missing something? Can we update the `creationTime` field of a resource or server will override it?

pkg/kapp/config/default.go
kumaritanushree: This is not set by user but kapp internally will update this value when condition matches. If will do `sources: [existing]` then it will always copy the old value (which is in exiting res) which we do not want.

pkg/kapp/diff/change_set_with_versioned_rs.go
kumaritanushree: If resource got recreated for some reason(reason could be ->user has made some update before max-duration expiry) after we updated last-renewed-time then it will be. This can happen if user has set update-strategy as always-replace for non-versioned resource or resource is versioned.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +84,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	return nil
+}
+
+func addLastRenewedTimeAnn(res, exRes ctlres.Resource, duration time.Duration) error {
+	var (
+		lastRenewed, creationTime time.Time
+		err                       error
+	)
+
+	if val, found := exRes.Annotations()[lastRenewTimeAnnKey]; found {
```
rohitagg2020: Can we simplify it like this(Reason: Our logic is based on lastRenewedTime rather than creationTime:

if val, found := exRes.Annotations()[lastRenewTimeAnnKey]; found {
   ...
} else {
lastRenewed = exRes.CreatedAt()
}

if time.Now().Before(lastRenewed.Add(duration)) {
		return nil
	}
....

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -35,14 +38,19 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
 	existingRs := existingVersionedResources(d.existingRs)
 	existingRsGrouped := d.groupResources(existingRs.Versioned)
 
+	err := d.checkAndUpdateResWithMaxDurationAnn(existingRsGrouped, existingRs)
```
100mik: Let's do this in `deploy.go` after we move the logic out to a separate file

100mik: The pre-reqs for running this are `existingVersionedResources(...)` which is already public and `d.groupResources(...)` which can be scoped to a separate struct

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +84,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
 	return allChanges, nil
 }
 
+func (d ChangeSetWithVersionedRs) checkAndUpdateResWithMaxDurationAnn(existingRsGrouped map[string][]ctlres.Resource,
+	existingRs versionedResources) error {
+
+	exNonVersionResMap := map[string]ctlres.Resource{}
+
+	for _, res := range existingRs.NonVersioned {
+		resKey := VersionedResource{res, nil}.UniqVersionedKey().String()
```
100mik: Lets use `ctlres.NewUniqueResourceKey(...)` directly instead

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +84,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
 	return allChanges, nil
 }
 
+func (d ChangeSetWithVersionedRs) checkAndUpdateResWithMaxDurationAnn(existingRsGrouped map[string][]ctlres.Resource,
```
100mik: Let's move this to another file scoped to a struct, something like `ResourcesWithRenewedResources`

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +84,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	}
+
+	for _, res := range d.newRs {
+		if val, found := res.Annotations()[maxDurationAnnKey]; found {
+			duration, err := time.ParseDuration(val)
+			if err != nil {
+				return fmt.Errorf("Unable to parse max-duration for resource %s: %s", res.Description(), err.Error())
+			}
+
+			resKey := VersionedResource{res, nil}.UniqVersionedKey().String()
```
100mik: Same as above, we do not need `VersionedResource{}` here

pkg/kapp/diff/change_set_with_versioned_rs.go
100mik: We want to use the found check to decide which time to use which makes sense.
I would also say we should split up the conditional into two lines (no strong opinions on this one)

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +84,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func addLastRenewedTimeAnn(res, exRes ctlres.Resource, duration time.Duration) error {
```
100mik: This function would be moved out too, and get scoped to the same struct as the new one

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -76,6 +84,72 @@ func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
...
+	return ctlres.StringMapAppendMod{
+		ResourceMatcher: ctlres.AllMatcher{},
+		Path:            ctlres.NewPathFromStrings([]string{"metadata", "annotations"}),
+		KVs: map[string]string{
+			lastRenewTimeAnnKey: fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339)),
+		},
+	}.Apply(res)
+}
+
 func (d ChangeSetWithVersionedRs) groupResources(rs []ctlres.Resource) map[string][]ctlres.Resource {
```
100mik: We need to create a new struct in this file, something like `groupedVersionedResources{groupedResources map[string]ctlres.Resource}` it can have the function `newGroupedVersionedResources(...)`  which does the exact same thing that this function does and returns an object of type `groupedVersionedResource`

100mik: We will refactor existing usage and can reuse this in our new file

100mik (commented):
A bunch of comments!
Main intention being to decouple renewed resource from versioned resources.
Still need to think about what we call things finally.

Kudos for the exhaustive tests, it handles every case I could think of! We should verify it does not add a flake because 2s might be less for slower hosts

pkg/kapp/diff/change_set_with_versioned_rs.go
kumaritanushree: As we had discussion over the call I have updated the changes.

pkg/kapp/diff/change_set_with_versioned_rs.go
kumaritanushree: This will require one more if check inside if case. 
Something like:
```
if val, found := exRes.Annotations()[lastRenewTimeAnnKey]; found {
...
      if lastRenewed.Before(exRes.CreatedAt()) {
		lastRenewed = exRes.CreatedAt()
	}
} else {
lastRenewed = exRes.CreatedAt()
}

if time.Now().Before(lastRenewed.Add(duration)) {
return nil
}
```
Existing implementation makes more sense to me.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -16,6 +16,8 @@ const (
 	versionedResAnnKey        = "kapp.k14s.io/versioned"               // Value is ignored
 	versionedResOrigAnnKey    = "kapp.k14s.io/versioned-keep-original" // Value is ignored
 	versionedResNumVersAnnKey = "kapp.k14s.io/num-versions"
+	maxDurationAnnKey         = "kapp.k14s.io/max-duration"
```
100mik: Lets move these two annotation values to the new file?

pkg/kapp/diff/changes_set_with_periodic_rs.go
```diff
@@ -0,0 +1,92 @@
+// Copyright 2022 VMware, Inc.
```
100mik: @kumaritanushree @praveenrewar what do you think about naming this something like `change_set_with_renewed_resources.go` ? 

I am thinking of cases where users might wanna renew in some other cases. For example, renew resource if the resource has a certain status. We could add those checks in here to expand on this feature if we see a case for it.

pkg/kapp/diff/change_set_with_versioned_rs.go
100mik: PS: I agree that `max-duration` is not the best name. But I cannot think of anything else either, I will leave this up to the rest of the folks.
Suggestions, `renew-after`, `resource-renew-duration`, `renew-duration` (I cannot say if these are better or worse to be frank)

100mik (commented):
How the feature is structured looks great to me!

@kumaritanushree mentioned that today if the user chooses to recreate on update, the last-renewed annotation is not retained. Maybe we should verify if the rebase rule is working. 

I do not see a lot of harm in the timer being reset in this case (right away at least). But worth verifying. I am pointing towards the rebase rule as it looks like the feature works otherwise in other case as we check existing resources directly. 

Lets also check the tests for flakiness since there's sleeping involved.

TL;DR Next steps:
1. Verify rebase rule
2. Ensure tests are not flaky (a few re runs on the GitHub action worker maybe?)

pkg/kapp/cmd/app/deploy.go
```diff
@@ -405,6 +405,11 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
 		changeFactory := ctldiff.NewChangeFactory(conf.RebaseMods(), conf.DiffAgainstLastAppliedFieldExclusionMods())
 		changeSetFactory := ctldiff.NewChangeSetFactory(o.DiffFlags.ChangeSetOpts, changeFactory)
 
+		err := ctldiff.NewChangeSetWithPeriodicRs(existingResources, newResources).Calculate()
```
praveenrewar: `NewChangeSetWithPeriodicRs.Calculate()` seems to indicate that we are calculating changes 🤔
How about we something like `RenewableResources` and `Prepare()`? (Just a thought, we could use something else)

pkg/kapp/diff/change_set_with_versioned_rs.go
praveenrewar: > Lets move these two annotation values to the new file?

+1

> Suggestions, renew-after, resource-renew-duration, renew-duration

I like `renew-after` and `renew-duration`

pkg/kapp/diff/changes_set_with_periodic_rs.go
praveenrewar: > what do you think about naming this something like change_set_with_renewed_resources.go ?

SGTM

pkg/kapp/diff/changes_set_with_periodic_rs.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (d ChangeSetWithPeriodicRes) Calculate() error {
+	existingRs := existingVersionedResources(d.existingRs)
+	existingVersionRsGrouped := newGroupedVersionedResources(existingRs.Versioned).groupedRes
+
+	return d.checkAndUpdateResWithMaxDurationAnn(existingVersionRsGrouped, existingRs)
+}
+
+func (d ChangeSetWithPeriodicRes) checkAndUpdateResWithMaxDurationAnn(existingVersionRsGrouped map[string][]ctlres.Resource,
```
praveenrewar: I think we can move this to `Calculate()`?

kumaritanushree: Renamed  api to Prepare as suggested above and moved the changes to it.

pkg/kapp/diff/changes_set_with_periodic_rs.go
kumaritanushree: As all related APIs, struct, annotation got changed, renamed file accordingly as well.

pkg/kapp/diff/renewable_resources.go
```diff
@@ -0,0 +1,99 @@
...
+		resKey := ctlres.NewUniqueResourceKey(res).String()
+		exNonVersionResMap[resKey] = res
+	}
+
+	for _, res := range d.newRs {
+		val, found := res.Annotations()[renewDurationAnnKey]
+		if found {
+			duration, err := time.ParseDuration(val)
+			if err != nil {
+				return fmt.Errorf("Unable to parse max-duration for resource %s: %s", res.Description(), err.Error())
```
praveenrewar: nit: max-duration => renew-duration

test/e2e/renewable-resources_test.go
```diff
@@ -0,0 +1,197 @@
+// Copyright 2022 VMware, Inc.
```
praveenrewar: nit: filenames should only have underscores

praveenrewar (commented):
Apart from the small nits, it looks good to me.

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -295,3 +275,44 @@ func existingVersionedResources(rs []ctlres.Resource) versionedResources {
 	}
 	return result
 }
+
+type groupedVersionedResources struct {
```
praveenrewar: I think we can revert this now to reduce some noise from this PR.

pkg/kapp/diff/renewable_resources.go
```diff
@@ -0,0 +1,84 @@
...
+		if err != nil {
+			return fmt.Errorf("Unable to parse last-renewed-time for resource %s: %s", res.Description(), err.Error())
+		}
+	}
+
+	if lastRenewed.Before(exRes.CreatedAt()) {
+		lastRenewed = exRes.CreatedAt()
+	}
+
+	renewTime = fmt.Sprintf("%v", time.Now().UTC().Format(time.RFC3339))
```
praveenrewar: This can be declared using a shothand operator, right?

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -295,3 +275,40 @@ func existingVersionedResources(rs []ctlres.Resource) versionedResources {
 	}
 	return result
 }
+
+func newGroupedVersionedResources(rs []ctlres.Resource) map[string][]ctlres.Resource {
```
100mik, line 279: I think I would prefer it if this function and the next one are either scoped to a struct, maybe we can move stuff that are used by other parts of the code to a single place at a later point. Not a blocker but a nice to have

100mik (approved):
LGTM! 
I would prefer not having the unstopped functions, but dedicating a new structure to each one of them feels a bit extra

pkg/kapp/diff/change_set_with_versioned_rs.go
```diff
@@ -33,7 +33,7 @@ func NewChangeSetWithVersionedRs(existingRs, newRs []ctlres.Resource,
 
 func (d ChangeSetWithVersionedRs) Calculate() ([]Change, error) {
 	existingRs := existingVersionedResources(d.existingRs)
-	existingRsGrouped := d.groupResources(existingRs.Versioned)
+	existingRsGrouped := newGroupedVersionedResources(existingRs.Versioned)
```
sethiyash, line 36: Should we also rename this variable to `existingVersionedRsGrouped`? Just a though not a blocker though.

sethiyash (commented):
LGTM!

pkg/kapp/diff/change_set_with_versioned_rs.go
sethiyash: +1

PR #574, opened 2022-08-09
bug fix: app metadata file should write to disk even when deploy fails

DennisDenuto:

bug fix when kapp deploy fails. Continue to write out the app metadata file with the latest state change. (Instead of not writing out a file when a deploy fails)

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -220,30 +220,27 @@ func (o *DeployOptions) Run() error {
 	}
 
 	err = touch.Do(func() error {
+		defer func() {
+			if o.DeployFlags.AppMetadataFile != "" {
+				meta, err := app.Meta()
+				if err != nil {
+					return
+				}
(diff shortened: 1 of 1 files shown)


pkg/kapp/cmd/app/deploy.go
```diff
@@ -220,30 +220,27 @@ func (o *DeployOptions) Run() error {
 	}
 
 	err = touch.Do(func() error {
+		defer func() {
+			if o.DeployFlags.AppMetadataFile != "" {
+				meta, err := app.Meta()
+				if err != nil {
```
praveenrewar, line 226: We don't seem to be doing anything with the errors. Should we just print it or maybe add a TODO to handle it in the future?
I am also thinking if it would make sense to write to the file when we are updating the used GVs and GKs. Currently we do it [here](https://github.com/vmware-tanzu/carvel-kapp/blob/71d5c375cb511732bf2e24bbc9ec24ba2c859069/pkg/kapp/cmd/app/deploy.go#L196) and below on line number 242.

praveenrewar (approved):
LGTM!
If needed, we can print the errors in later PRs.

PR #597, opened 2022-09-04
Add Scripts for podman-based development

criztovyl:

During restructuring #517 I extracted these changes that could be added as a dedicate PR.

Containerized, Podman-based, kapp development.

```release-note
NONE
```

The other `hack` scripts are not documented, so I have not added docs for the new scripts (yet?). 

But people familiar with Bash and Podman should be able to understand the scripts.

##### Review Checklist:

(Does not seem applicable.)

- [x]  Follows the [developer guidelines](https://carvel.dev/shared/docs/latest/development_guidelines/)
  a link to that PR
  change

```docs
None
```

### DIFF
--- a/hack/.gitignore
+++ b/hack/.gitignore
@@ -1,0 +1,1 @@
+.bash_history
--- a/hack/.gitignore
+++ b/hack/.gitignore
@@ -1,1 +1,2 @@
 .bash_history
+.dlv
--- a/hack/build-podman.sh
+++ b/hack/build-podman.sh
@@ -2,4 +2,4 @@
 
 cd $(dirname ${BASH_SOURCE[0]})
 
-./podman.sh hack/build.sh
+./podman.sh build
--- a/hack/build-podman.sh
+++ b/hack/build-podman.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd $(dirname ${BASH_SOURCE[0]})
+
+podman run --rm -v gopath:/go -v gobuild:/root/.cache/go-build -v $(realpath $PWD/..):/mnt -w /mnt docker.io/golang hack/build.sh
--- a/hack/podman.sh
+++ b/hack/podman.sh
@@ -2,4 +2,8 @@
 
 cd $(dirname ${BASH_SOURCE[0]})
 
-podman run --rm -v gopath:/go -v gobuild:/root/.cache/go-build -v $(realpath $PWD/..):/mnt -w /mnt docker.io/golang $1
+action=$1
+
+[ -x "$action.sh" ] && action=hack/$action.sh
+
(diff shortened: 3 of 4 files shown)


hack/.gitignore
```diff
@@ -0,0 +1,2 @@
+.bash_history
+.dlv
```
praveenrewar, line 2: Should we add these in the `.gitignore` present in the root?

hack/build-podman.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd $(dirname ${BASH_SOURCE[0]})
+
+./podman.sh build
```
praveenrewar, line 5: Do we need to run the podman.sh script from the hack directory for it to work? If so, can we make some changes to be able to run it from the root itself (like other scripts are doing, test-all.sh for example). It seems like we are adding one extra layer of complexity in build-podman.sh for cd'ing to the hack directory.

hack/podman.sh
```diff
@@ -0,0 +1,35 @@
...
+gobuild_vol=${KAPP_GOBUILD_VOL:-gobuild}
+
+# SYS_PTRACE for dlv
+# host network for minikube
+podman run --rm \
+    -v $gopath_vol:/go \
+    -v $gobuild_vol:/root/.cache/go-build \
+    -v $HOME/.kube:/root/.kube \
+    -v $HOME/.minikube:$HOME/.minikube \
+    -v $(realpath $PWD/..):/mnt -w /mnt \
```
praveenrewar, line 31: Maybe add a comment that realpath would require `brew install coreutils` on OSX?

praveenrewar (commented):
When I run the script on my local, I get this error `Error: statfs /Users/prewar/Workspace/carvel-kapp: no such file or directory`. I am not sure what I might be doing wrong. I haven't gone through the script in detail but will go through it in some time.

hack/.gitignore
criztovyl: You mean instead of here? I can change it if you prefer.

hack/build-podman.sh
criztovyl: I'll check that.

hack/podman.sh
criztovyl: My podman scripts most likely are generally only compatible with Linux, not sure how the PWD mount and host network will work on OSX where Podman runs in a VM.

criztovyl:
Thinking it over, the kapp repo might not be the best place to add these scripts, in theory the are generic for podman-/container-based go development.

Maybe I'll rather put these to a personal repo (so, not add them here at all), what do you think?

praveenrewar:
> in theory the are generic for podman-/container-based go development.

Yep!

> Maybe I'll rather put these to a personal repo (so, not add them here at all), what do you think?

Sounds good to me :)

PR #598, opened 2022-09-05
Obj null check to avoid failure on non-typed character like NULL as value for array type key/object

### LINKED ISSUE #89: kapp template rules do not work against non-typed fields such as `initContainers: null`
When trying to deploy the [bitnami postgresql chart](https://github.com/bitnami/charts/blob/301bb18b0f8455d6f64c26aedd50ee9d7c74c221/bitnami/postgresql/templates/statefulset.yaml#L74) as part of cf-for-k8s via kapp, we get the following error:

`
kapp: Error: ObjectRefSetMod for path 'spec,template,spec,initContainers,(all),env,(all),valueFrom,configMapKeyRef' on resource 'statefulset/cf-db-postgresql (apps/v1) namespace: cf-db': Unexpected non-array found: <nil>
`

The relevant configuration yaml looks like this (with irrelevant lines omitted):

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: cf-db-postgresql
spec:
  template:
    spec:
      initContainers: null
```

When we try to deploy the same YAML with `kubectl apply -f`, we don't get any error.

We're using kapp version 0.22.0.

Thanks!
Jen & @jamespollard8

kumaritanushree:
Added null check to avoid non-typed character error (like envFrom: null/<empty>) while applying template rules in case of versioned resources.

Fixes #89 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/test/e2e/non_typed_array_fields_test.go
+++ b/test/e2e/non_typed_array_fields_test.go
@@ -1,5 +1,6 @@
-// Copyright 2020 VMware, Inc.
+// Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
+
 package e2e
 
 import (
@@ -7,7 +8,7 @@ import (
 	"testing"
 )
 
-func TestNonTypedCharError(t *testing.T) {
+func TestNonTypedArrayFields(t *testing.T) {
 	env := BuildEnv(t)
 	logger := Logger{}
 	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
@@ -15,26 +16,38 @@ func TestNonTypedCharError(t *testing.T) {
 
 	yaml := `
 ---
-apiVersion: apps/v1
-kind: Deployment
+apiVersion: v1
+kind: Service
 metadata:
-  name: simple-app
+  name: simple-svc
+  labels:
+    app: simple-svc
 spec:
+  ports:
+  - port: 80
+    name: test
+  clusterIP: None
+  selector:
+    app: simple-svc
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: test
+spec:
+  replicas: 1
   selector:
     matchLabels:
-      simple-app: ""
+      app: simple-svc
   template:
     metadata:
       labels:
-        simple-app: ""
+        app: simple-svc
     spec:
+      initContainers: null
       containers:
       - name: simple-app
         image: docker.io/dkalinin/k8s-simple-app@sha256:4c8b96d4fffdfae29258d94a22ae4ad1fe36139d47288b8960d9958d1e63a9d0
-        env:
-        - name: HELLO_MSG
-          value: stranger
-        envFrom:
 ---
 apiVersion: v1
 kind: ConfigMap
@@ -46,7 +59,7 @@ data:
   hello_msg: carvel
 `
 
-	name := "test-non-typed-character-error"
+	name := "test-non-typed-array-fields"
 	cleanUp := func() {
 		kapp.Run([]string{"delete", "-a", name})
 	}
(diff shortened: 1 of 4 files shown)


100mik (dismissed):
LGTM

test/e2e/non_typed_char_error_test.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 VMware, Inc.
```
praveenrewar: nit: 2020 -> 2022 and new line after the headers.

test/e2e/non_typed_char_error_test.go
```diff
@@ -0,0 +1,62 @@
...
+// SPDX-License-Identifier: Apache-2.0
+package e2e
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestNonTypedCharError(t *testing.T) {
```
praveenrewar: TestNonTypedArrayFields ? (Might as well want to change the file name and name on line number 49 based on this).

test/e2e/non_typed_char_error_test.go
```diff
@@ -0,0 +1,62 @@
...
+        envFrom:
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: simple-cm
+  annotations:
+    kapp.k14s.io/versioned: ""
+data:
+  hello_msg: carvel
```
praveenrewar:
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: test
spec:
  template:
    spec:
      initContainers: null
```
Should we add this from the issue?

test/e2e/non_typed_char_error_test.go
kumaritanushree: updated

test/e2e/non_typed_char_error_test.go
kumaritanushree: updated

test/e2e/non_typed_char_error_test.go
kumaritanushree: done

test/e2e/non_typed_char_error_test.go
100mik: very nitty, but on a second thought should the name point out that it is testing template rules in this case? (`TestTemplateRulesWithNonTypedArrayFields` maybe, even though it is a mouthful)

Do not feel very strongly about this but I think it does not hurt to be more explicit

test/e2e/non_typed_array_fields_test.go
```diff
@@ -0,0 +1,76 @@
...
+spec:
+  ports:
+  - port: 80
+    name: test
+  clusterIP: None
+  selector:
+    app: simple-svc
+---
+apiVersion: apps/v1
+kind: StatefulSet
```
100mik: nit: I think we changed it to a StatefulSet to address the example in the original issue,

However, I think a Deployment pod template could have both `envFrom: null` and `initContainers:`.
Thought I would point that out since a StatefulSet "does more" than a Deployment and that is not something we require for this case.

100mik (commented):
Just a thought

test/e2e/non_typed_char_error_test.go
kumaritanushree: Agree as this says exactly where the issue was. Let wait for other's opinion on it. @praveenrewar

praveenrewar: Works for me 👍🏻

test/e2e/non_typed_array_fields_test.go
praveenrewar: Yeah, I was thinking of just keeping the StatefulSet, right now I think the Service and the ConfigMap don't mean anything in the test. We can keep the previous example of Deployment and ConfigMap if we want, and add the initContainers along with envFrom 👍🏻

rohitagg2020 (approved):

100mik (approved):
Ship it! 🚢

PR #601, opened 2022-09-09
Save app-metadata-file-output even if no change is present

praveenrewar:
Save app-metadata-file even if no change is present

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -178,6 +178,8 @@ func (o *DeployOptions) Run() error {
 	}
 
 	if o.DiffFlags.Run || hasNoChanges {
+		o.writeAppMetadataToFile(app)
+
 		if o.DiffFlags.Run && o.DiffFlags.ExitStatus {
 			return DeployDiffExitStatus{hasNoChanges}
 		}
@@ -220,18 +222,7 @@ func (o *DeployOptions) Run() error {
 	}
 
 	err = touch.Do(func() error {
-		defer func() {
-			if o.DeployFlags.AppMetadataFile != "" {
-				meta, err := app.Meta()
-				if err != nil {
-					return
-				}
-				err = os.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)
-				if err != nil {
-					return
-				}
-			}
-		}()
+		defer o.writeAppMetadataToFile(app)
 
 		err := clusterChangeSet.Apply(clusterChangesGraph)
 		if err != nil {
@@ -470,6 +461,17 @@ func (o *DeployOptions) existingPodResources(existingResources []ctlres.Resource
 		}
 	}
 	return existingPods
+}
+
+func (o *DeployOptions) writeAppMetadataToFile(app ctlapp.App) error {
+	if o.DeployFlags.AppMetadataFile != "" {
+		meta, err := app.Meta()
+		if err != nil {
+			return err
+		}
+		return os.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)
+	}
+	return nil
 }
 
 const (
--- a/test/e2e/app_change_test.go
+++ b/test/e2e/app_change_test.go
@@ -133,6 +133,8 @@ metadata:
 	assert.NoError(t, err)
 	secondDeploy, err := os.CreateTemp(os.TempDir(), "output2")
 	assert.NoError(t, err)
+	thirdDeploy, err := os.CreateTemp(os.TempDir(), "output3")
+	assert.NoError(t, err)
 
 	defer func() {
 		os.Remove(firstDeploy.Name())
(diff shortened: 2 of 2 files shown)


pkg/kapp/cmd/app/deploy.go
```diff
@@ -220,18 +222,7 @@ func (o *DeployOptions) Run() error {
...
-				if err != nil {
-					return
-				}
-				err = os.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)
-				if err != nil {
-					return
-				}
-			}
-		}()
+		defer o.writeAppMetadataToFile(app)
```
100mik, line 225: Is there any harm in doing this just once when we are sure we have the app (we can defer the function once, around line 116 I believe would work)?

I do not see why we *need* to write the file before `touch.Do` exits 🤔

100mik (commented):
Just a thought!

pkg/kapp/cmd/app/deploy.go
praveenrewar: > Is there any harm in doing this just once when we are sure we have the app (we can defer the function once, around line 116 I believe would work)?

Good point!! I will check if that's doable.

> I do not see why we need to write the file before touch.Do exits 🤔

We need to save the metadata file even if there's no change to the app (we are returning early in that case and do not reach touch.Do)

100mik: > We need to save the metadata file even if there's no change to the app (we are returning early in that case and do not reach touch.Do)

Yep that makes sense. I was wondering if the deferring needs to be done inside the function passed to `touch.Do` for some reason instead of doing it once earlier, that is all. I could not think of anything right off the bat.

praveenrewar: So, we have the [AskForConformation](https://github.com/vmware-tanzu/carvel-kapp/blob/68322f5ba9dc153aac6b76d5ee97255883a7a46f/pkg/kapp/cmd/app/deploy.go#L190) on line 190 and I think that we shouldn't write to metadata before this, i.e. before the user gives the confirmation. The only special case to this would be when there are no changes or diff run flag is being used (which is where I have added the defer call). I am still thinking what would be the right place to have a single defer call, but do let me know if you have any thoughts on this.

100mik: Oh I see, so we only wanna write the file if
1. There are no changes
2. Changes get applied

100mik: The only change I see in that case is that when there are no changes, the call does not need to be deferred.

Another benefit of this would be that when we grab new GKs before confirmation, those do not get written unless the changes have been applied.

Resolving this and putting a comment above.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -178,6 +178,8 @@ func (o *DeployOptions) Run() error {
 	}
 
 	if o.DiffFlags.Run || hasNoChanges {
+		defer o.writeAppMetadataToFile(app)
```
100mik: Do we need to defer here?

Looks like we return after this no matter what. No room for erroring.

praveenrewar: Makes sense, updated.

test/e2e/app_change_test.go
```diff
@@ -133,6 +133,8 @@ metadata:
 	assert.NoError(t, err)
 	secondDeploy, err := os.CreateTemp(os.TempDir(), "output2")
 	assert.NoError(t, err)
+	thirdDeploy, err := os.CreateTemp(os.TempDir(), "output3")
```
100mik, line 136: I think this entire test should go into a separate file, discussed this with @praveenrewar and that could go in as a separate PR to keep this one less noisy

@praveenrewar good first issue maybe? October is around the corner 😆

100mik (approved):
LGTM!

test/e2e/app_change_test.go
praveenrewar: Yeah, separate PR makes sense 👍🏻

sethiyash:
Looks Good.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -472,6 +463,20 @@ func (o *DeployOptions) existingPodResources(existingResources []ctlres.Resource
 	return existingPods
 }
 
+func (o *DeployOptions) writeAppMetadataToFile(app ctlapp.App) error {
+	if o.DeployFlags.AppMetadataFile != "" {
+		meta, err := app.Meta()
+		if err != nil {
+			return err
+		}
+		err = os.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)
```
rohitagg2020: Change to `return os.WriteFile(o.DeployFlags.AppMetadataFile, []byte(meta.AsString()), os.ModePerm)`

praveenrewar: Done.

rohitagg2020 (approved):

kumaritanushree:
LGTM

100mik (approved):
LGTM 🚀

100mik:
Created #604 for some minor refactoring

PR #610, opened 2022-09-19
Skip gatekeeper in test-external.sh

praveenrewar:
- Skip gatekeeper in test-external.sh
- gatekeeper uses PodSecurityPolicy which is removed in kubernetes 1.25. We can add gatekeeper v3.10.0 when it's available (as it migrates away from PodSecurityPolicy)

```release-note
NONE
```


100mik (approved):
Sounds reasonable, we can manually test against a cluster running an older version if we want to

Issue #62, opened 2019-12-18
support path key prefix selection in Config


cwlbraa:
I'd like to write a rebase rule that always pulls all of the `cert-manager.io/` annotations for secrets out of the cluster, but right now I think I have to write that like this:

```
rebaseRules:
- path: [metadata, annotations, "cert-manager.io/alt-names"]
  type: copy
  sources: [new, existing]
  resourceMatchers: &opaqueSecrets
  - apiVersionKindMatcher: {apiVersion: v1, kind: Secret, type: Opaque}

- path: [metadata, annotations, "cert-manager.io/certificate-name"]
  type: copy
  sources: [new, existing]
  resourceMatchers: *opaqueSecrets

- path: [metadata, annotations, "cert-manager.io/common-name"]
  type: copy
  sources: [new, existing]
  resourceMatchers: *opaqueSecrets

- path: [metadata, annotations, "cert-manager.io/ip-sans"]
  type: copy
  sources: [new, existing]
  resourceMatchers: *opaqueSecrets

- path: [metadata, annotations, "cert-manager.io/issuer-kind"]
  type: copy
  sources: [new, existing]
  resourceMatchers: *opaqueSecrets

- path: [metadata, annotations, "cert-manager.io/issuer-name"]
  type: copy
  sources: [new, existing]
  resourceMatchers: *opaqueSecrets

- path: [metadata, annotations, "cert-manager.io/uri-sans"]
  type: copy
  sources: [new, existing]
  resourceMatchers: *opaqueSecrets
```

when really what I want is something like:

```
rebaseRules:
- path: [metadata, annotations, {prefixed: "cert-manager.io/"}]
  type: copy
  sources: [new, existing]
  - apiVersionKindMatcher: {apiVersion: v1, kind: Secret, type: Opaque}
```

**Tasks**
- [x] rebase rules to support regex as new field in config path
- [ ] diff stage to support new field. changes in  (mod_field_copy, remove, overlay_contract mod - resources package) 
- [ ] test ownership rules, label scoping rule
- [ ] e2e test case
- [ ] documentation

sethiyash:
Here is the [Proposal](https://hackmd.io/VgeLq3yDRR67aLElFvjOyA) on the different approaches which is still in the discussion phase.

sethiyash:
Internally we decide to start supporting `regex` in path similar to the [ `allIndexes`](https://carvel.dev/kapp/docs/v0.53.0/config/#paths) we had earlier. 
```
[...]

PR #636, opened 2022-11-08
Supporting regex in config path

sethiyash:
Start supporting regex in config path that will match all the path which adhere to regex 

#62 

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```


100mik:
Interesting stuff!
Looks like the linter is not happy and there is some commented/dead code, will take a look at this post cleanup 🚀

test/e2e/config_test.go
```diff
@@ -689,6 +690,152 @@ rules:
...
+    foo2: bar2
+  labels:
+  name: game-test
+  namespace: kapp-test
+`
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		out = clearKeys(fieldsExcludedInMatch, out)
+
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		require.Contains(t, out, expectedOutput, "output does not match")
```
rohitagg2020: Not sure what the test case is trying to achieve.

sethiyash: This is just to verify that both configmaps are getting created successfully.

sethiyash: If you all think it's not required I can remove it and just have the below two cases

pkg/kapp/resources/mod_path.go
```diff
@@ -20,15 +20,16 @@ type ResourceModWithMultiple interface {
 type Path []*PathPart
 
 type PathPart struct {
-	MapKey     *string
-	ArrayIndex *PathPartArrayIndex
+	MapKey        *string
+	IndexAndRegex *PathPartIndexAndRegex
```
praveenrewar: I feel that Regex could be a different value here. 
- It would reduce a lot of noise in the PR.
- Index and Regex might have similar use cases, but they are actually different in the sense that one works for arrays while the other doesn't, so it does make sense to have a different structure for Regex.

test/e2e/config_test.go
praveenrewar: Yeah, seems like an unnecessary step in this test.

test/e2e/config_test.go
```diff
@@ -689,6 +690,152 @@ rules:
...
+	fieldsExcludedInMatch := []string{"kapp.k14s.io/app", "creationTimestamp:", "resourceVersion:", "uid:", "selfLink:", "kapp.k14s.io/association"}
+	name := "test-config-path-regex"
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("creating an app with multiple resources", func() {
```
praveenrewar: nit: deploy configmaps with annotations

test/e2e/config_test.go
```diff
@@ -689,6 +690,152 @@ rules:
...
+  namespace: kapp-test
+`
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		out = clearKeys(fieldsExcludedInMatch, out)
+
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		require.Contains(t, out, expectedOutput, "output does not match")
+	})
+
+	logger.Section("Deploy with new yaml with removed annotations and copying from existing resource", func() {
```
praveenrewar: nit: deploy configmaps without annotations
And then have a comment above `require.Contains` which talks about the copying and regex

rohitagg2020: Yes, the written test doesn't make much sense as we are not trying to verify the functionality implemented in this PR.

sethiyash: Adopted.

test/e2e/config_test.go
sethiyash: Adopted.

test/e2e/config_test.go
sethiyash: Adopted.

pkg/kapp/resources/mod_path.go
sethiyash: If everyone is aligned to have it as a different value I can make that change but this will be a significant change again and have to look into parsing the regex part separately.

test/e2e/config_test.go
```diff
@@ -689,6 +690,121 @@ rules:
...
+	logger.Section("deploy configmaps without annotations", func() {
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--diff-changes-yaml"}, RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml2 + fmt.Sprintf(configYaml, "copy"))})
+		expectedOutput := ``
+
+		out = strings.TrimSpace(replaceTarget(replaceSpaces(replaceTs(out))))
+		out = clearKeys(fieldsExcludedInMatch, out)
+
+		expectedOutput = strings.TrimSpace(replaceSpaces(expectedOutput))
+		// rebase rule should copy the annotations matched with regex
+		require.Contains(t, out, expectedOutput, "output does not match")
```
rohitagg2020: I think we should do something which shows that `foo*` got copied over.

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -73,7 +74,7 @@ func (t FieldCopyMod) apply(obj interface{}, path Path, fullPath Path, srcs map[
 			}
 
 		case part.ArrayIndex != nil:
-			if isLast {
+			if isLast && part.RegexObj == nil {
```
praveenrewar: Question: Can both ArrayIndex and RegexObj be non-nil at the same time?

pkg/kapp/resources/mod_path.go
```diff
@@ -21,6 +21,7 @@ type Path []*PathPart
 
 type PathPart struct {
 	MapKey     *string
+	RegexObj   *PathPartRegex
```
praveenrewar: Since it's already an object, should it just be `Regex` and not `RegexObj`? (Then we will have to use `Regex.Regex` in some places which might sound weird but I couldn't come up with any other alternative)

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -120,6 +121,26 @@ func (t FieldCopyMod) apply(obj interface{}, path Path, fullPath Path, srcs map[
 				panic(fmt.Sprintf("Unknown array index: %#v", part.ArrayIndex))
 			}
 
+		case part.RegexObj != nil:
+			if part.RegexObj.Regex != nil {
```
praveenrewar: Since we are not doing anything when `RegexObj.Regex` is nil, maybe we can do something like `case part.RegexObj != nil && part.RegexObj.Regex != nil`?

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -200,3 +221,53 @@ func (t FieldCopyMod) obtainValue(obj interface{}, path Path) (interface{}, bool
...
+
+			var found bool
+			obj, found = typedObj[*part.MapKey]
+			if !found {
+				return matchedKeys, nil
+			}
+
+		case part.RegexObj.Regex != nil:
+			switch {
+			case part.RegexObj.Regex != nil:
```
praveenrewar: Same as above

pkg/kapp/resources/mod_field_remove.go
```diff
@@ -92,6 +92,19 @@ func (t FieldRemoveMod) apply(obj interface{}, path Path) error {
 			default:
 				panic(fmt.Sprintf("Unknown array index: %#v", part.ArrayIndex))
 			}
+		case part.RegexObj.Regex != nil:
```
praveenrewar: Shouldn't we first check for `part.RegexObj` to be non nil? (Also, same as above, we can check both in `case`)

praveenrewar (commented):
Still going through, but submitting for the time being...

pkg/kapp/resources/mod_field_copy.go
sethiyash: No so as of now we will be having one as nil that's how we are unmarshalling and now also we have removed regex from ArrayIndex I can remove this condition as well.

test/e2e/config_test.go
sethiyash: Should I print the ConfigMaps with kubectl? (But I think that will make the test output look messy)

pkg/kapp/resources/mod_field_copy.go
sethiyash: Adopted

pkg/kapp/resources/mod_field_remove.go
sethiyash: Adopted

pkg/kapp/resources/mod_field_copy.go
sethiyash: Adopted

pkg/kapp/resources/mod_path.go
sethiyash: Adopted

test/e2e/config_test.go
sethiyash: Or should I use some tool to just print the annotations of both the configmaps?

test/e2e/config_test.go
```diff
@@ -689,6 +690,121 @@ rules:
...
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy configmaps with annotations", func() {
+		_, _ = kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--diff-changes-yaml"},
+			RunOpts{IntoNs: true, StdinReader: strings.NewReader(yaml1)})
```
kumaritanushree: Why are we using `diff-changes-yaml` here? 
Suggestion: Also I think we should have check for successful run of this cmd or check for the existence of resources we are deploying here.

pkg/kapp/resources/mod_field_remove.go
```diff
@@ -92,6 +92,19 @@ func (t FieldRemoveMod) apply(obj interface{}, path Path) error {
...
+				return err
+			}
+			for _, key := range matchedKeys {
+				newPath := append(Path{&PathPart{MapKey: &key}}, path[i+1:]...)
+				err := t.apply(obj, newPath)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
```
praveenrewar: We can probably get rid of the `obtainMatchingRegexKeys` by combining some stuff (as it just seems to be moving the complexity from one place to another):
```
			regex, err := regexp.Compile(*part.Regex.Regex)
			if err != nil {
				return err
			}

			typedObj, ok := obj.(map[string]interface{})
			if !ok && typedObj != nil {
				return fmt.Errorf("Unexpected non-map found: %T", obj)
			}

			for key := range typedObj {
				if regex.MatchString(key) {
					newPath := append(Path{&PathPart{MapKey: &key}}, path[i+1:]...)
					err := t.apply(obj, newPath)
					if err != nil {
						return err
					}
				}
			}
			return nil
```

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -200,3 +219,50 @@ func (t FieldCopyMod) obtainValue(obj interface{}, path Path) (interface{}, bool
 
 	return obj, true, nil
 }
+
+func (t FieldCopyMod) matchRegexWithSources(path Path, srcs map[FieldCopyModSource]Resource) ([]string, error) {
+	// always looking into existing resource
```
praveenrewar: *question*: Why do we want to look at existing resources only?

pkg/kapp/resources/mod_field_remove.go
```diff
@@ -57,7 +58,7 @@ func (t FieldRemoveMod) apply(obj interface{}, path Path) error {
 			}
 
 		case part.ArrayIndex != nil:
-			if isLast {
+			if isLast && part.Regex.Regex == nil {
```
praveenrewar: We can revert this as well.

test/e2e/config_test.go
rohitagg2020: No new tool is needed. you can use kubectl only. It has provision to print only things you are interested in. e.g.
```kubectl get cm/kapp-controller-config -n kapp-controller -o jsonpath='{.metadata.annotations}'```

Comment (kumaritanushree) [inline test/e2e/config_test.go]: This test is not working as expected, I think expectation is to get empty yaml as output of `diff-changes-yaml` but  both configmaps are getting updated here. Seems rebase rule is not getting applied properly.

Comment (praveenrewar) [reply test/e2e/config_test.go]: I think the fact that there is no diff should be fine to show that foo* got copied. Although, let's use `--diff-run` and `--diff-exit-status` like [this](https://github.com/vmware-tanzu/carvel-kapp/blob/1ecc9149f578d6e700b27cea4b28cac70ca32a16/test/e2e/config_test.go#L616) instead of `--diff-changes-yaml` to make it more clean.

File: test/e2e/config_test.go [hunk 15]
```diff
@@ -689,6 +690,121 @@ rules:
...
+data:
+  player_initial_lives: "3"
+`
+
+	configYaml := `
+---
+apiVersion: kapp.k14s.io/v1alpha1
+kind: Config
+rebaseRules:
+  - path: [metadata, annotations, {regex: "^foo"}]
```
praveenrewar, line 902: Should we also add some more cases (non-happy paths)? Like empty regex?

pkg/kapp/resources/mod_field_remove.go
sethiyash: Yeah, I can do that. I just abstracted the logic in different function to convey its purpose. 
Also, If we want to rid of this function from here then we should have to do for `mod_field_copy.go` as well.

pkg/kapp/resources/mod_field_copy.go
sethiyash: (Yeah so I'm also sceptical about this that's why put the comment. There was a question mark earlier but after the first iteration, I haven't got any comment on this so I removed the question mark)

Was not sure about the use case of copying from a new resource
But, I think It will be ideal to look for all the sources one by one and as soon as we got the field we can exit

test/e2e/config_test.go
sethiyash: Updated the test cases now.

test/e2e/config_test.go
sethiyash: stopped using `diff-changes-yaml`

test/e2e/config_test.go
sethiyash: Empty regex will match any string so that won't make any difference here. Added a test case to check for invalid regex path like this - `path: [metadata, annotations, {regex:  }]`

pkg/kapp/resources/mod_field_copy.go
sethiyash: Made the changes to iterate on the given sources

pkg/kapp/resources/mod_field_remove.go
sethiyash: Done

test/e2e/config_test.go
sethiyash: Done

pkg/kapp/resources/mod_field_remove.go
praveenrewar: > If we want to rid of this function from here then we should have to do for mod_field_copy.go as well.

We probably can.

pkg/kapp/resources/mod_path.go
```diff
@@ -82,6 +87,15 @@ func (p Path) ContainsNonMapKeys() bool {
 	return false
 }
 
+func (p Path) ContainsNonMapKeysAndRegex() bool {
+	for _, part := range p {
+		if part.MapKey == nil && part.Regex.Regex == nil {
```
praveenrewar: Should check for `part.Regex` first

kumaritanushree:
LGTM

rohitagg2020:
LGTM

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -27,29 +28,36 @@ func (t FieldCopyMod) ApplyFromMultiple(res Resource, srcs map[FieldCopyModSourc
...
-	if err != nil {
-		return fmt.Errorf("FieldCopyMod for path '%s' on resource '%s': %s", t.Path.AsString(), res.Description(), err)
-	}
-
-	if updated {
-		res.setUnstructured(updatedRes.unstructured())
+	for _, src := range t.Sources {
+		// Make a copy of resource, to avoid modifications
+		// that may be done even in case when there is nothing to copy
+		updatedRes := res.DeepCopy()
```
rohitagg2020: Can we move it above line 39 from performance persepective.

sethiyash: We have thought about this but as we are iterating on the sources like `[new,existing]` provided in Config so then whatever update happened for `new` resource will stay and can reflect when we iterate for `existing`

praveenrewar: Let's move it after the `if !found` condition.

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -30,29 +31,36 @@ func (t FieldCopyMod) IsResourceMatching(res Resource) bool {
...
+		}
+		updated, err := t.apply(updatedRes.unstructured().Object, source.unstructured().Object, t.Path, Path{}, srcs)
+		if err != nil {
+			return fmt.Errorf("FieldCopyMod for path '%s' on resource '%s': %s", t.Path.AsString(), res.Description(), err)
+		}
+		if updated {
+			res.setUnstructured(updatedRes.unstructured())
+		}
 	}
-
```
praveenrewar, line 45: nit: Let's keep the newline

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -123,6 +154,24 @@ func (t FieldCopyMod) apply(obj interface{}, path Path, fullPath Path, srcs map[
...
+				newFullPath := fullPath[:len(fullPath)-1]
+				updated, err := t.apply(obj, srcObj, newPath, newFullPath, srcs)
+				if err != nil {
+					return false, err
+				}
+				if !updated {
+					allUpdated = false
+				}
+			}
+			return allUpdated, nil
```
praveenrewar: Shouldn't we return true if even one path is updated?

pkg/kapp/resources/mod_field_remove.go
```diff
@@ -78,7 +79,6 @@ func (t FieldRemoveMod) apply(obj interface{}, path Path) error {
 						return err
 					}
 				}
-
```
praveenrewar: nit: Let's keep the new line

pkg/kapp/resources/mod_field_remove.go
```diff
@@ -96,6 +96,19 @@ func (t FieldRemoveMod) apply(obj interface{}, path Path) error {
 			default:
 				panic(fmt.Sprintf("Unknown array index: %#v", part.ArrayIndex))
 			}
+		case part.Regex != nil && part.Regex.Regex != nil:
+			matchedKeys, err := t.obtainMatchingRegexKeys(obj, *part.Regex.Regex)
```
praveenrewar: I guess we will need the source object for remove as well?

sethiyash: As `sources` in rebaseRules is only used with `copy` not with `remove` so that's not available and hence not required here.

pkg/kapp/resources/mod_field_copy.go
sethiyash: copy just in one path can be considered as successful? If we get `n` matchedKeys, all keys should be copied before telling the operation is successful. Is it not the fair assumption?

pkg/kapp/resources/mod_field_copy.go
sethiyash: Adopted.

pkg/kapp/resources/mod_field_copy.go
praveenrewar: Even if one of the paths is copied successfully, we should be setting this, right?
```go
res.setUnstructured(updatedRes.unstructured())
```

sethiyash: make sense. Made the changes.

pkg/kapp/resources/mod_field_copy.go
```diff
@@ -203,3 +258,21 @@ func (t FieldCopyMod) obtainValue(obj interface{}, path Path) (interface{}, bool
 
 	return obj, true, nil
 }
+
+func (t FieldCopyMod) matchRegexWithSrcObj(regexString string, srcObj interface{}) ([]string, error) {
```
praveenrewar: We don't seem to be using anything from `t` (`FieldCopyMod`) here, so should we keep it as a function and reuse it in mod_field_remove.go?

sethiyash: Sounds Good, Done.

praveenrewar (approved):
LGTM!

PR #647, opened 2022-11-30
Bump Dependencies

praveenrewar:
- Bump go to v1.19.4
- Bump golang.org/x/net to v0.4.0


.github/workflows/golangci-lint.yml
```diff
@@ -12,10 +12,10 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - name: Install Go 1.18
-        uses: actions/setup-go@v3
+      - name: Set up Go
+        uses: actions/setup-go@v2
         with:
-          go-version: "1.18"
```
100mik, line 18: Do we want to bump the version of lint being used as well?

praveenrewar: Sure, maybe a separate PR?

100mik: Makes sense since it only affects the action

100mik (approved):
LGTM

PR #659, opened 2022-12-13
Avoid listing labeled resources for newly created app

### LINKED ISSUE #599: Performance enhancements
We rely on `list` api calls to get information from the cluster which could put some burden on the cluster if the number of objects returned is high. When the number of apps being deployed using kapp increases (in cases of kapp-controller packages), this becomes a problem as the time taken to deploy the apps increases after a certain point without any burden on the cpu or memory of the cluster nodes.
- Throttling warnings when there are multiple kapp apps being used at the same time.
- `socket: too many open files` when ulimit is set to a low number (256)

**Describe the solution you'd like*
We need to minimise the list calls as much as possible (Replacing them with get or watch is also an option).

**Tasks**
* [x] Instead of trying to get [all](https://github.com/vmware-tanzu/carvel-kapp/blob/9e863ee3668282e236e34142b221d75629e637a4/pkg/kapp/resources/resource_types.go#L120) the server resources, get only the ones that are related to the available GKs (anyway we get rid of the others [later](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/resources/identified_resources_list.go#L42) so they are not being used) - Cancelled for the time being
  - [x] Spike: This require more code changes and makes the code less readable. So we will just revisit this later if required.
* [x] When we deploy an app, we first list the labeled resources (GVs) and then try to get the non labeled resources one by one one that are not found in the first step. When an app is deployed for the first time, the first step would always return nil, so maybe we could skip that?
  - [x] Spike: How would this fit into kapp code base? -> [PR](https://github.com/vmware-tanzu/carvel-kapp/pull/659)
               
* [ ] Use watch instead of get and list while waiting for resources to reconcile. Using watch will be helpful for resources that take more time to reconcile (for example deployments), but for resources that reconcile almost immediately (for example configmap), it might bring some overhead.
   - [x] Spike: Test if reducing the wait time interval help with this? 
              We increased the wait-check-interval to 3s as it is reducing api calls and not affecting deployment time much. 
              PR for the same is [here](https://github.com/vmware-tanzu/carvel-kapp/pull/637) and the data collected while doing spike can check [here](https://docs.google.com/spreadsheets/d/1ymlvw9ChFnOdGFg__kIYdx1z6TowDTk4ObLUNqE569c/edit#gid=485873872)
   - [ ] Spike: Test if using watch impacts performance. -> As we have increased wait-check-interval now and it is giving better result,  we can have a look on watch later. Prioritising second steps for now. 
* [ ] When we have a CRD and CR present in the same manifest, we try to fetch the server resources again to find the CRD (since it wasn't present in the cached server resources). We should somehow avoid doing this as we wouldn't find the CRD this time as well. (No need to work on this if we already work on the first one)
* [ ] Now that we have added the resource namespaces to the fallbackAllowedNamespaces, should we always use fallbackAllowedNamespaces instead of checking resources cluster wide?
  - [ ] Spike: Figure out if scoping to fallbackAllowedNamespaces could have any side effects (testing).
* [ ] Currently we store the unique GKs in the meta ConfigMap and we do a list on the GKs, since list calls are more expensive we can check if doing get calls for all the resources is less expensive than list calls for unique GKs.  
* [x] Improving performance enhancement specifically during diff stage. With go profiling, it was noticed that there are too many calls to [deepCopy](https://github.com/carvel-dev/kapp/blob/b75689aedd76c2e022a8497473acd1e5facf2210/pkg/kapp/resources/resource.go#L236) and [AsYAMLBytes](https://github.com/carvel-dev/kapp/blob/develop/pkg/kapp/resources/resource.go#L248). [PR](https://github.com/carvel-dev/kapp/pull/731)

It might be worth understanding the [API priority and fairness](https://kubernetes.io/docs/concepts/cluster-administration/flow-control/).

---
Vote on this request

kumaritanushree:
When we deploy an app, we first list the labeled resources (GVs) and then try to get the non labeled resources one by one one that are not found in the first step. When an app is deployed for the first time, the first step would always return nil.

This PR has changes to avoid listing labeled resources for newly created app

Fixes #599 

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -92,6 +92,7 @@ func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 }
 
 func (o *DeployOptions) Run() error {
+	var newApp bool
 	failingAPIServicesPolicy := o.ResourceTypesFlags.FailingAPIServicePolicy()
 
 	app, supportObjs, err := Factory(o.depsFactory, o.AppFlags, o.ResourceTypesFlags, o.logger)
@@ -107,7 +108,7 @@ func (o *DeployOptions) Run() error {
 	if o.PrevAppFlags.PrevAppName != "" {
 		err = app.RenamePrevApp(o.PrevAppFlags.PrevAppName, appLabels, o.DiffFlags.Run)
 	} else {
-		err = app.CreateOrUpdate(appLabels, o.DiffFlags.Run)
+		newApp, err = app.CreateOrUpdate(appLabels, o.DiffFlags.Run)
 	}
 
 	if err != nil {
@@ -158,7 +159,7 @@ func (o *DeployOptions) Run() error {
 	}
 
 	existingResources, existingPodRs, err := o.existingResources(
-		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...))
+		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...), newApp)
 	if err != nil {
 		return err
 	}
@@ -352,7 +353,7 @@ func (o *DeployOptions) newResourcesFromFiles() ([]ctlres.Resource, error) {
 
 func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 	labeledResources *ctlres.LabeledResources, resourceFilter ctlres.ResourceFilter,
-	apps ctlapp.Apps, usedGKs []schema.GroupKind, resourceNamespaces []string) ([]ctlres.Resource, []ctlres.Resource, error) {
+	apps ctlapp.Apps, usedGKs []schema.GroupKind, resourceNamespaces []string, newApp bool) ([]ctlres.Resource, []ctlres.Resource, error) {
 
 	labelErrorResolutionFunc := func(key string, val string) string {
 		items, _ := apps.List(nil)
@@ -378,6 +379,7 @@ func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 		IdentifiedResourcesListOpts: ctlres.IdentifiedResourcesListOpts{
 			GKsScope:           usedGKs,
 			ResourceNamespaces: resourceNamespaces,
+			NewApp:             newApp,
 		},
 	}
 
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -92,7 +92,7 @@ func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 }
 
 func (o *DeployOptions) Run() error {
-	var newApp bool
+	var isNewApp bool
 	failingAPIServicesPolicy := o.ResourceTypesFlags.FailingAPIServicePolicy()
 
 	app, supportObjs, err := Factory(o.depsFactory, o.AppFlags, o.ResourceTypesFlags, o.logger)
@@ -108,7 +108,7 @@ func (o *DeployOptions) Run() error {
 	if o.PrevAppFlags.PrevAppName != "" {
 		err = app.RenamePrevApp(o.PrevAppFlags.PrevAppName, appLabels, o.DiffFlags.Run)
 	} else {
-		newApp, err = app.CreateOrUpdate(appLabels, o.DiffFlags.Run)
+		isNewApp, err = app.CreateOrUpdate(appLabels, o.DiffFlags.Run)
 	}
 
 	if err != nil {
@@ -159,7 +159,7 @@ func (o *DeployOptions) Run() error {
 	}
 
 	existingResources, existingPodRs, err := o.existingResources(
-		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...), newApp)
+		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...), isNewApp)
 	if err != nil {
 		return err
 	}
@@ -353,7 +353,7 @@ func (o *DeployOptions) newResourcesFromFiles() ([]ctlres.Resource, error) {
 
 func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 	labeledResources *ctlres.LabeledResources, resourceFilter ctlres.ResourceFilter,
-	apps ctlapp.Apps, usedGKs []schema.GroupKind, resourceNamespaces []string, newApp bool) ([]ctlres.Resource, []ctlres.Resource, error) {
+	apps ctlapp.Apps, usedGKs []schema.GroupKind, resourceNamespaces []string, isNewApp bool) ([]ctlres.Resource, []ctlres.Resource, error) {
 
 	labelErrorResolutionFunc := func(key string, val string) string {
 		items, _ := apps.List(nil)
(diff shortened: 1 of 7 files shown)


pkg/kapp/cmd/app/deploy.go
```diff
@@ -92,6 +92,7 @@ func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 }
 
 func (o *DeployOptions) Run() error {
+	var newApp bool
```
100mik: We can declare this inline

pkg/kapp/cmd/app/deploy.go
```diff
@@ -107,7 +108,7 @@ func (o *DeployOptions) Run() error {
 	if o.PrevAppFlags.PrevAppName != "" {
 		err = app.RenamePrevApp(o.PrevAppFlags.PrevAppName, appLabels, o.DiffFlags.Run)
 	} else {
-		err = app.CreateOrUpdate(appLabels, o.DiffFlags.Run)
+		newApp, err = app.CreateOrUpdate(appLabels, o.DiffFlags.Run)
```
100mik: `newApp` => `isNewApp`

pkg/kapp/cmd/app/deploy.go
```diff
@@ -378,6 +379,7 @@ func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 		IdentifiedResourcesListOpts: ctlres.IdentifiedResourcesListOpts{
 			GKsScope:           usedGKs,
 			ResourceNamespaces: resourceNamespaces,
+			NewApp:             newApp,
```
100mik: Let's use `isNewApp` and `IsNewApp`

100mik (commented):
A few nitpicks!

pkg/kapp/resources/identified_resources_list.go
```diff
@@ -25,6 +26,11 @@ func (r IdentifiedResources) List(labelSelector labels.Selector, resRefs []Resou
 		return nil, err
 	}
 
+	// avoid listing labeled resources for newly created app
+	if opts.NewApp {
+		return nil, nil
+	}
```
cppforlife: what about call above that happens and gets throw away?

kumaritanushree: In that call results are being memoized [here](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/resources/resource_types.go#L157) that is why added return after that.

praveenrewar: I guess the question then is, do we need that memoized list if it's a new app?

kumaritanushree: hmm, nice question. Will have a look again to check in which cases this memoized info are being used.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -158,7 +159,7 @@ func (o *DeployOptions) Run() error {
 	}
 
 	existingResources, existingPodRs, err := o.existingResources(
-		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...))
+		newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...), newApp)
```
rohitagg2020: Can we explore based on whether it is newApp or not, if we make the call to `o.existingResources(...)`. Something like:
```
if !newApp {
existingResources, existingPodRs, err := o.existingResources(newResources, labeledResources, resourceFilter, supportObjs.Apps, usedGKs, append(meta.LastChange.Namespaces, nsNames...))
}
```
This way we will have minimal code change. Needs regression testing as well.

praveenrewar: We still need to check the existing non labeled resources even if it is a new app.
https://github.com/vmware-tanzu/carvel-kapp/blob/7e7b3a0485238f1e6c1f481835f102c12db95ac2/pkg/kapp/resources/labeled_resources.go#L119

praveenrewar: But I agree that we could probably make the check a little before then we have done now.
Maybe [here](https://github.com/vmware-tanzu/carvel-kapp/blob/7e7b3a0485238f1e6c1f481835f102c12db95ac2/pkg/kapp/resources/labeled_resources.go#L110)? (Once [this](https://github.com/vmware-tanzu/carvel-kapp/pull/659/files#r1051870816) is finalised)

pkg/kapp/resources/identified_resources_list.go
kumaritanushree: At this stage kapp is using memoized info, which is being memoized at [this](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/cmd/app/deploy.go#L145) stage while preparing newResources, so we can skip this and return bit early. Updated changes as well.

pkg/kapp/cmd/app/deploy.go
kumaritanushree: Can not declare  inline as variable is being declared inside if block.

sethiyash (approved):
LGTM!

pkg/kapp/cmd/app/deploy.go
rohitagg2020: I agree. Also, did we test the scenario: A resource already exists on cluster, and we try to kappify it. Also, can we have e2e test case around it if not already there?

100mik: Does not seem like a test like that is there. Since we are touching those parts of the code, I think it makes sense to have a test which verifies using the diff summary that kapp adopts existing resources.

100mik:
The changes themselves look good to me!
I am good with this after we add a test making sure that we are including non-labelled resources.

pkg/kapp/cmd/app/deploy.go
kumaritanushree: added e2e test

test/e2e/app_deploy_test.go
```diff
@@ -0,0 +1,105 @@
...
+package e2e
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAppDeploy(t *testing.T) {
```
praveenrewar: Should we add this test to [create_update_delete_test.go](https://github.com/vmware-tanzu/carvel-kapp/blob/develop/test/e2e/create_update_delete_test.go) with a more descriptive name?

test/e2e/app_deploy_test.go
```diff
@@ -0,0 +1,105 @@
...
+	name2 := "test-app-deploy2"
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", name})
+		kapp.Run([]string{"delete", "-a", name2})
+	}
+
+	cleanUp()
+	defer cleanUp()
+
+	logger.Section("deploy cm1 using kubectl", func() {
```
praveenrewar: nit: I don't think we need a logger section for this as we are not testing kapp in this step.

test/e2e/app_deploy_test.go
kumaritanushree: Yeah, it make more sense to move it here. Updated changes.

test/e2e/create_update_delete_test.go
```diff
@@ -443,3 +443,95 @@ data:
...
+		expectedOutput := `
+@@ update configmap/cm2 (v1) namespace: kapp-test @@
+  ...
+  1,  1   data:
+  2     -   key: value
+      2 +   key: value2
+  3,  3   kind: ConfigMap
+  4,  4   metadata:
+`
+		require.Contains(t, out, expectedOutput, "output does not match")
```
100mik, line 535: What are we trying to check for over here? (Just clarifying)

test/e2e/create_update_delete_test.go
```diff
@@ -443,3 +443,95 @@ data:
 		cleanUp()
 	})
 }
+
+func TestAppDeploy_With_Existing_OR_New_Res(t *testing.T) {
```
100mik, line 447: Should we also delete the app in the end? Right now the cleanup function _should_ error out if it fails.

100mik: I see that the above test relies on cleanUp as well, so this is fine.

100mik (approved):
The changes look good to me! Just a few clarifications

test/e2e/create_update_delete_test.go
kumaritanushree: Trying to check that changes are working with existing app as well.

100mik: Oh makes sense, looks like I read this wrong

praveenrewar (approved):

PR #661, opened 2022-12-14
Garbage collect app change configmap in delete command

### LINKED ISSUE #644: Failure in `kapp delete` results in creation of app-change ConfigMaps with no limit on the number of app-changes
**What steps did you take:**
- Deploy an app
- Delete the app (with no permission to delete ConfigMap, to simulate a failure)

Notice that the resources are deleted but the app metadata ConfigMap still exists and there is an additional app-change ConfigMap for the last change.

- Try to delete the app again 

Notice that there is another app-change CM present now.

**What happened:**
`kapp delete` creates app change ConfigMaps which are then deleted, but in case the app deletion fails for some reason, these ConfigMaps remain on the cluster. If there is a controller or a script which tries to delete the app at certain intervals then the number of CMs goes onto increase without any limit.

**What did you expect:**
- App change ConfigMaps should not be created for delete command in the first place (this is debatable)
- In case we are creating the CMs for delete command, there should be garbage collection happening in delete just like deploy command.

**Environment:**
- kapp version (use `kapp --version`): 0.54.0
- OS (e.g. from `/etc/os-release`): darwin
- Kubernetes version (use `kubectl version`) 1.25.0

---
Vote on this request

sethiyash:
Today, We don't garbage collect the app-change configmap in the delete command started doing that just like we do in deploy command.

Fixes #644

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -149,6 +149,15 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
+	if !shouldFullyDeleteApp {
+		defer func() {
+			_, numDeleted, _ := app.GCChanges(ctlapp.AppChangesMaxToKeepDefault, nil)
+			if numDeleted > 0 {
+				o.ui.PrintLinef("Deleted %d older app changes", numDeleted)
+			}
+		}()
+	}
+
 	touch := ctlapp.Touch{App: app, Description: "delete", IgnoreSuccessErr: true}
 
 	err = touch.Do(func() error {
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -154,6 +154,12 @@ func (o *DeleteOptions) Run() error {
 	err = touch.Do(func() error {
 		err := clusterChangeSet.Apply(clusterChangesGraph)
 		if err != nil {
+			if shouldFullyDeleteApp {
+				_, numDeleted, _ := app.GCChanges(5, nil)
+				if numDeleted > 0 {
+					o.ui.PrintLinef("Deleted %d older app changes", numDeleted)
+				}
+			}
 			return err
 		}
 		if shouldFullyDeleteApp {


pkg/kapp/cmd/app/delete.go
```diff
@@ -149,6 +151,13 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
+	defer func() {
+		_, numDeleted, _ := app.GCChanges(o.DeleteFlags.AppChangesMaxToKeep, nil)
+		if numDeleted > 0 {
+			o.ui.PrintLinef("Deleted %d older app changes", numDeleted)
+		}
+	}()
+
```
cppforlife: app.Delete already deletes app changes (https://github.com/vmware-tanzu/carvel-kapp/blob/develop/pkg/kapp/app/recorded_app.go#L362-L365)

sethiyash: I believe this is of help when the delete command is executed successfully but is it fair to consider a case when the delete command is getting failed every time it gets called by some script in a certain interval? 
In the case, when delete is not succeeding the above changes can be helpful where it will remove all the app-change once it reaches the threshold.

praveenrewar: A few thoughts:
- Maybe we should add this change (without the defer) [here](https://github.com/vmware-tanzu/carvel-kapp/blob/1c1ed44974d33259d9f2233d9821a1340edb2ab4/pkg/kapp/cmd/app/delete.go#L157), as this is the only scenario where the deletion of app-changes won't happen.
- Also, it probably doesn't make sense to have this flag (`app-changes-max-to-keep`) for delete command since we are deleting *all* the app changes (in success scenario at least). Instead, we could just maybe keep the last app-change only and delete rest of them.

100mik: This makes sense to me ^
Does not need to be deferred and the flag seems counter-intuitive 🤔 

However, do we want to have a more conservative default? Maybe keep 5 or even 10 app-changes?
Arguably, we should reduce the default for `kapp deploy` as well but that would require us to look into whether folks are relying on them today.

We often have folks struggling to look at older reconciliations, even though this is much less likely to be useful in the case of delete but it does not hurt to be safer.

rohitagg2020: I have got an interesting thought. If somebody tries to delete based on filters, some resources will be deleted while others will stay. In that case, do we want to remove the history or not?

praveenrewar: I was thinking we should use `shouldFullyDeleteApp` to check if that's the case.

100mik:
@sethiyash we should probably finalise on this and take it forward.

@praveenrewar if I am correct the only change left is to remove the flag and default to keeping 5 (or any other number of) changes?

praveenrewar:
> @praveenrewar if I am correct the only change left is to remove the flag and default to keeping 5 (or any other number of) changes?

Yep, that's correct.

pkg/kapp/cmd/app/delete.go
sethiyash: Made the changes to make sure even in case of partially deleting based on filters we garbage collect the app-changes

pkg/kapp/cmd/app/delete.go
```diff
@@ -153,12 +153,16 @@ func (o *DeleteOptions) Run() error {
 
 	err = touch.Do(func() error {
 		err := clusterChangeSet.Apply(clusterChangesGraph)
+		if err == nil && shouldFullyDeleteApp {
+			return app.Delete()
+		}
+		_, numDeleted, _ := app.GCChanges(5, nil)
```
praveenrewar: At this stage, `err` could be `nil` and `shouldFullyDeleteApp` could be false, and in that case, I think we should not garbage collect app-changes.
How about we just add this in the existing `err != nil` check and keep the shouldFullyDeleteApp condition?
```go
if shouldFullyDeleteApp {
    _, numDeleted, _ := app.GCChanges(5, nil)
    if numDeleted > 0 {
	o.ui.PrintLinef("Deleted %d older app changes", numDeleted)
    }
}
```

sethiyash: I was thinking of the scenario when the user will partially delete the resources with the help of filters then in theory it can also lead to an enormous amount of app changes if some script is doing it (not sure whether it is much practical or not) 
Do you think should we ignore such scenarios and just garbage collect when the deletion is unsuccessful?

sethiyash: After discussion with Praveen, we have decided to keep app changes the same as the deploy default which is 200 when the user partially deletes app based on some filters and max app-changes to 5 when deletion is unsuccessful due to error.

pkg/kapp/cmd/app/delete.go
```diff
@@ -149,11 +149,26 @@ func (o *DeleteOptions) Run() error {
 		return err
 	}
 
+	if !shouldFullyDeleteApp {
+		defer func() {
+			_, numDeleted, _ := app.GCChanges(ctlapp.AppChangesMaxToKeepDefault, nil)
```
sethiyash, line 154: Scenario when the user partially deletes apps using filters we are garbage collecting once it reaches the threshold of 200 similar to the deploy stage. (let me know if you think it is not an ideal number)

sethiyash:
Deleted it accidentally while force pushing it from local opening a new PR

PR #676, opened 2023-01-21
Change org and repository name

praveenrewar:
Change org and repository name

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -14,5 +14,5 @@ jobs:
     steps:
       - uses: actions/add-to-project@v0.3.0
         with:
-          project-url: https://github.com/orgs/vmware-tanzu/projects/16
+          project-url: https://github.com/orgs/carvel-dev/projects/1
           github-token: ${{ secrets.CARVEL_ADD_TO_PROJECT_TOKEN }}
--- a/README.md
+++ b/README.md
@@ -5,8 +5,8 @@
 - Website: https://carvel.dev/kapp
 - Slack: [#carvel in Kubernetes slack](https://slack.kubernetes.io)
 - [Docs](docs/README.md) with topics about diff, apply, gitops, config, _blog posts and talks_ etc.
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ Please join us during our online community meetings. Details can be found on our
 
 You can chat with us on Kubernetes Slack in the #carvel channel and follow us on Twitter at @carvel_dev.
 
-Check out which organizations are using and contributing to Carvel: [Adopter's list](https://github.com/vmware-tanzu/carvel/blob/master/ADOPTERS.md)
+Check out which organizations are using and contributing to Carvel: [Adopter's list](https://github.com/carvel-dev/carvel/blob/master/ADOPTERS.md)
```
sethiyash, line 43: should we also change the branch from master to develop here? (not a blocker as it still works today by redirecting it to the default branch with a popup)
```suggestion
Check out which organizations are using and contributing to Carvel: [Adopter's list](https://github.com/carvel-dev/carvel/blob/develop/ADOPTERS.md)
```

sethiyash (approved):
LGTM

README.md
praveenrewar: Sure, but maybe a separate PR?

sethiyash: Yeah.

kumaritanushree:
LGTM

.github/workflows/add-to-project.yml
```diff
@@ -14,5 +14,5 @@ jobs:
     steps:
       - uses: actions/add-to-project@v0.3.0
         with:
-          project-url: https://github.com/orgs/vmware-tanzu/projects/16
+          project-url: https://github.com/orgs/carvel-dev/projects/1
```
rohitagg2020, line 17: Not much idea, but it seems this will add it in the `All` tab. 

Do we want to make it `https://github.com/orgs/carvel-dev/projects/6` which correspond to `kapp` tab? Although I do see earlier as well we were pointing to `All` tab.

rohitagg2020 (approved):

Issue #646, opened 2022-11-29
App name cannot have more than 63 characters


DimpleRajaVamsi:
**What steps did you take:**
Created an app with the app name having more than 63 characters.

**What happened:**
App creation failed because the app-change ConfigMaps uses the app name as a label.
```
kapp: Error: Creating app change:
  ConfigMap "byoi-2213-photon-ns-antrea-11ns1-c1-guest-cluster-auth-serzfmpw" is invalid: metadata.labels:
    Invalid value: "byoi-2213-photon-ns-antrea-11ns1-c1-guest-cluster-auth-service-ctrl": must be no more than 63 characters
```

**What did you expect:**
App to be created successfully.

**Anything else you would like to add:**
This causes issues for App CRs and PackageInstall CRs from kapp-controller because a kapp app is created using the name of app/package install.

**Environment:**
- kapp version (use `kapp --version`): kapp controller version is `v0.41.2`
- OS (e.g. from `/etc/os-release`): `Photon`
- Kubernetes version (use `kubectl version`): `1.22.13`

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

praveenrewar:
Currently, the app name is used to generate the label value for the `app-change-app` label in the app change ConfigMaps, so if the app name is more than 63 characters, the label value becomes invalid as there is a limit on the number of characters for a label.

We will use a different property (the label generated by kapp for each app) of an app to generate the label key in app change ConfigMaps. The label generated by kapp is unique for an app and it has limited number of characters so the app name length won't effect it.

##### Current
App Name -> `just-another-name`
`kapp.k14s.io/app-change-app` -> `just-another-name`

##### Proposed
App Name -> `just-another-name`
kapp generated label -> `19023903840`
`kapp.k14s.io/app-change-app` -> `19023903840`

However doing so would introduce a breaking change in terms of how app-changes are generated, listed and deleted.
We can also make this behaviour opt-in so that folks can migrate to it when they feel comfortable in doing so, but that would add an extra overhead in terms of release of the change (i.e. releasing it as opt in first and in future versions making it opt out and then mandatory).

Adding a few scenarios to elaborate both the approaches:

## 1. Introduce it as a breaking change
[...]

PR #685, opened 2023-01-25
Use app label as app change label for long app names

praveenrewar:
Use app label to as app change label for long app names
If app names have more than 63 character, use the app label as app change label instead of the app name as there is a hard limit on the number of characters that can be present in a label value

This is the first step towards fixing #646 

```release-note
Use app label as app-change label if app name has more than 63 characters
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -359,7 +359,12 @@ func (a *RecordedApp) Delete() error {
 		return err
 	}
 
-	err = NewRecordedAppChanges(a.nsName, a.name, a.coreClient).DeleteAll()
+	meta, err := a.meta()
+	if err != nil {
+		return err
+	}
+
+	err = NewRecordedAppChanges(a.nsName, a.name, meta.LabelValue, a.coreClient).DeleteAll()
 	if err != nil {
 		return fmt.Errorf("Deleting app changes: %w", err)
 	}
@@ -498,7 +503,11 @@ func (a *RecordedApp) meta() (Meta, error) {
 }
 
 func (a *RecordedApp) Changes() ([]Change, error) {
-	return NewRecordedAppChanges(a.nsName, a.name, a.coreClient).List()
+	meta, err := a.meta()
+	if err != nil {
+		return nil, err
+	}
+	return NewRecordedAppChanges(a.nsName, a.name, meta.LabelValue, a.coreClient).List()
 }
 
 func (a *RecordedApp) LastChange() (Change, error) {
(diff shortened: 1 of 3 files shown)


pkg/kapp/app/recorded_app_changes.go
```diff
@@ -122,3 +125,10 @@ func (a RecordedAppChanges) Begin(meta ChangeMeta) (*ChangeImpl, error) {
 
 	return change, nil
 }
+
+func (a RecordedAppChanges) changeLabelValue() string {
```
cppforlife, line 129: anything we are doing to deal with renames of an app? (or is this future work?)

pkg/kapp/app/recorded_app.go
```diff
@@ -522,7 +531,12 @@ func (a *RecordedApp) LastChange() (Change, error) {
 }
 
 func (a *RecordedApp) BeginChange(meta ChangeMeta) (Change, error) {
-	change, err := NewRecordedAppChanges(a.nsName, a.name, a.coreClient).Begin(meta)
+	appMeta, err := a.meta()
+	if err != nil {
+		return nil, err
+	}
```
cppforlife, line 537: for these additional meta calls, im assuming we arent making additional API calls?

rohitagg2020:
LGTM

pkg/kapp/app/recorded_app_changes.go
praveenrewar: As of now, we abandon all the app-changes in case of a rename (with user confirmation), so this check should be good for now to resolve the issue. The only difference that will be there now is that we will not be able to list the app-changes with old names. Example:
```
$ kapp deploy -a very-long-name -f config.yml

$ kapp rename -a very-long-name --new-name short-name

$ kapp app-changes list -a very-long-name
kapp: Error: App 'very-long-name' (namespace: name-limit) does not exist:
```
The error would happen because of the `meta` calls.

For future, we can completely migrate to using the label.

pkg/kapp/app/recorded_app.go
praveenrewar: For most cases we don't make an additional API call. 
In some cases there would be an additional API call (for example `kapp app-change` commands).

praveenrewar: I initially added `setMeta()` in a couple of places to make sure that we don't make an unnecessary call, but I reverted it because it would probably make sense to add that in additional PRs. (This won't effect `kapp app-change` commands though)

100mik:
Linked the related issue ^

100mik (approved):
LGTM!

PR #709, opened 2023-02-14
Combining steps to run kapp test in release.yml

rohitagg2020:
Combining steps to run kapp test in release.yml in a single step.

Signed-off-by: rohitagg2020 <rohit.aggarwal2020@gmail.com>

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,33 +33,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Setup Minikube
-        run: |
-          curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
-          install minikube-linux-amd64 /usr/local/bin/minikube
-          minikube start --driver=docker --memory 4096
-          eval $(minikube docker-env --shell=bash)
-
-          # Ensure that there is no existing kapp installed
-          rm -f /tmp/bin/kapp
-
-      - name: Create kapp binary
-        run: |
-          set -e -x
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -24,8 +24,8 @@ jobs:
           go-version: 1.19.5
 
       - name: Run GoReleaser
-        # GoReleaser v4.2.0
-        uses: goreleaser/goreleaser-action@f82d6c1c344bcacabba2c841718984797f664a6b
+        # GoReleaser v2.5.0
+        uses: goreleaser/goreleaser-action@5e15885530fb01d81d1f24e8a6f54ebbd0fed7eb
```
praveenrewar: Should we have a separate commit for this (separation of concern and also to cherry pick)

.github/workflows/release.yml
```diff
@@ -43,21 +43,20 @@ jobs:
           # Ensure that there is no existing kapp installed
           rm -f /tmp/bin/kapp
 
-      - name: Create kapp binary
+      - name: Run Test cases
```
praveenrewar: Should we also move this step to the end? The tests can fail due to various reasons and sometimes we might wanna go ahead with the release even if the tests have failed, so if we move this step to the end then we will know if the checksums are matching. But even if the tests fail and the remaining steps are still executed, then I guess it's fine to have it here itself.

.github/workflows/release.yml
rohitagg2020: Do we want to squash these commits later?

rohitagg2020:
Raised another [PR](https://github.com/carvel-dev/kapp/pull/713).

PR #710, opened 2023-02-16
Use app label as app-change label for new apps

### LINKED ISSUE #646: App name cannot have more than 63 characters

praveenrewar:
- Introduce a new label `kapp.k14s.io/app-change-app-label` for app changes which uses the app label as value
- Use the old label to list and delete app-changes created for exiting apps, but also add this new label to them.
- Use the new label to list and delete app-changes created for new apps, but add the old label if length of app name is less than the maximum allowed length of a label key to keep the app-change backward compatible

Fixes #646 

```release-note
Breaking change!!
App label generated by kapp will be used as app-change label instead of the app name for all the new apps.
- Using app name as a label value in app changes puts a limit on the number of characters a app name can have
- If an app deployed with new version of kapp (and having more than 63 characters) is used with an old version of kapp, then the old version won't be able to detect the app-changes created by the new version.
```

In future, after x number of releases, we can stop using the legacy label and start using the new label for existing apps as well. Since all the app-changes created for them would most probably have both the labels, it won't result in any loss of app-changes.

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -26,13 +26,15 @@ const (
 	KappIsConfigmapMigratedAnnotationKey   = "kapp.k14s.io/is-configmap-migrated"
 	KappIsConfigmapMigratedAnnotationValue = ""
 	AppSuffix                              = ".apps.k14s.io"
+	KappAppChangesUseAppLabelAnnotationKey = "kapp.k14s.io/app-changes-use-app-label"
 )
 
 type RecordedApp struct {
-	name              string
-	nsName            string
-	isMigrated        bool
-	creationTimestamp time.Time
+	name                  string
+	nsName                string
+	isMigrated            bool
+	creationTimestamp     time.Time
+	appChangesUseAppLabel bool
 
 	coreClient             kubernetes.Interface
 	identifiedResources    ctlres.IdentifiedResources
@@ -46,7 +48,7 @@ func NewRecordedApp(name, nsName string, creationTimestamp time.Time, coreClient
 	identifiedResources ctlres.IdentifiedResources, appInDiffNsHintMsgFunc func(string) string, logger logger.Logger) *RecordedApp {
 
 	// Always trim suffix, even if user added it manually (to avoid double migration)
-	return &RecordedApp{strings.TrimSuffix(name, AppSuffix), nsName, false, creationTimestamp, coreClient, identifiedResources, appInDiffNsHintMsgFunc,
+	return &RecordedApp{strings.TrimSuffix(name, AppSuffix), nsName, false, creationTimestamp, false, coreClient, identifiedResources, appInDiffNsHintMsgFunc,
 		nil, logger.NewPrefixed("RecordedApp")}
 }
 
@@ -196,6 +198,10 @@ func (a *RecordedApp) create(labels map[string]string, isDiffRun bool) error {
 			Labels: map[string]string{
 				KappIsAppLabelKey: kappIsAppLabelValue,
 			},
+			// Use app label as app change label for all new apps
+			Annotations: map[string]string{
+				KappAppChangesUseAppLabelAnnotationKey: "",
+			},
 		},
 		Data: Meta{
 			LabelKey:   kappAppLabelKey,
@@ -336,7 +342,7 @@ func (a *RecordedApp) Delete() error {
 		return err
 	}
 
-	err = NewRecordedAppChanges(a.nsName, a.name, meta.LabelValue, a.coreClient).DeleteAll()
+	err = NewRecordedAppChanges(a.nsName, a.name, meta.LabelValue, a.appChangesUseAppLabel, a.coreClient).DeleteAll()
 	if err != nil {
 		return fmt.Errorf("Deleting app changes: %w", err)
 	}
@@ -429,6 +435,9 @@ func (a *RecordedApp) isMigrationEnabled() bool {
 func (a *RecordedApp) Meta() (Meta, error) { return a.meta() }
 
 func (a *RecordedApp) setMeta(app corev1.ConfigMap) (Meta, error) {
+	// TODO: Should this be moved out of this function?
+	_, a.appChangesUseAppLabel = app.Annotations[KappAppChangesUseAppLabelAnnotationKey]
+
 	meta, err := NewAppMetaFromData(app.Data)
 	if err != nil {
 		errMsg := "App '%s' (namespace: %s) backed by ConfigMap '%s' did not contain parseable app metadata: %w"
@@ -479,7 +488,8 @@ func (a *RecordedApp) Changes() ([]Change, error) {
 	if err != nil {
 		return nil, err
 	}
-	return NewRecordedAppChanges(a.nsName, a.name, meta.LabelValue, a.coreClient).List()
+
+	return NewRecordedAppChanges(a.nsName, a.name, meta.LabelValue, a.appChangesUseAppLabel, a.coreClient).List()
 }
 
 func (a *RecordedApp) LastChange() (Change, error) {
@@ -508,7 +518,7 @@ func (a *RecordedApp) BeginChange(meta ChangeMeta) (Change, error) {
 		return nil, err
 	}
 
-	change, err := NewRecordedAppChanges(a.nsName, a.name, appMeta.LabelValue, a.coreClient).Begin(meta)
+	change, err := NewRecordedAppChanges(a.nsName, a.name, appMeta.LabelValue, a.appChangesUseAppLabel, a.coreClient).Begin(meta)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/kapp/app/recorded_app_changes.go
+++ b/pkg/kapp/app/recorded_app_changes.go
@@ -17,22 +17,24 @@ import (
 )
 
 const (
-	isChangeLabelKey   = "kapp.k14s.io/is-app-change"
-	isChangeLabelValue = ""
-	changeLabelKey     = "kapp.k14s.io/app-change-app" // holds app name or label
+	isChangeLabelKey     = "kapp.k14s.io/is-app-change"
+	isChangeLabelValue   = ""
+	legacyChangeLabelKey = "kapp.k14s.io/app-change-app"       // holds app name
+	changeLabelKey       = "kapp.k14s.io/app-change-app-label" // holds app label
 )
 
 type RecordedAppChanges struct {
-	nsName  string
-	appName string
+	nsName           string
+	appName          string
+	changeLabelValue string
 
-	appLabelValue string
+	appChangeUsesAppLabel bool
 
 	coreClient kubernetes.Interface
 }
 
-func NewRecordedAppChanges(nsName, appName, appLabelValue string, coreClient kubernetes.Interface) RecordedAppChanges {
-	return RecordedAppChanges{nsName, appName, appLabelValue, coreClient}
+func NewRecordedAppChanges(nsName, appName, changeLabelValue string, appChangeUsesAppLabel bool, coreClient kubernetes.Interface) RecordedAppChanges {
+	return RecordedAppChanges{nsName, appName, changeLabelValue, appChangeUsesAppLabel, coreClient}
 }
 
 func (a RecordedAppChanges) List() ([]Change, error) {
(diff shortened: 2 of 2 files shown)


pkg/kapp/app/recorded_app.go
```diff
@@ -26,13 +26,16 @@ const (
 	KappIsConfigmapMigratedAnnotationKey   = "kapp.k14s.io/is-configmap-migrated"
 	KappIsConfigmapMigratedAnnotationValue = ""
 	AppSuffix                              = ".apps.k14s.io"
+	KappIsAppLabelAsAppChangeLabelKey      = "kapp.k14s.io/is-app-label-app-change-label"
```
100mik: "kapp.k14s.io/uses-app-label" maybe? A bit of a tongue twister right now

pkg/kapp/app/recorded_app.go
```diff
@@ -585,6 +604,19 @@ func (a *RecordedApp) update(doFunc func(*Meta)) error {
 	return nil
 }
 
+func (a *RecordedApp) changeLabelValue() (string, error) {
```
100mik: nit: `appChangeLabelValue` feels more comprehensive to me.
Same for all the places where we use `changeLabelValue` 🤔

pkg/kapp/app/recorded_app.go
100mik: Or I guess `changes-use-app-id` is more appropriate

pkg/kapp/app/recorded_app.go
```diff
@@ -26,13 +26,16 @@ const (
...
 type RecordedApp struct {
-	name              string
-	nsName            string
-	isMigrated        bool
-	creationTimestamp time.Time
+	name                     string
+	nsName                   string
+	isMigrated               bool
+	creationTimestamp        time.Time
+	isAppLabelAppChangeLabel bool
```
100mik: `isAppLabelAppChangeLabel`  => `UsesAppID` ?

pkg/kapp/app/recorded_app.go
```diff
@@ -146,6 +150,7 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string, isDiffRun bool) (
 	}
 
 	if foundNonMigratedApp {
+		_, a.isAppLabelAppChangeLabel = app.Annotations[KappIsAppLabelAsAppChangeLabelKey]
```
100mik: Since we are doing this for both migrated and non-migrated apps, why don't we populate this value before we make checks for migration?
I do not see the "name migration" and "label migration" being coupled in any way, am I missing something?

pkg/kapp/app/recorded_app.go
```diff
@@ -174,6 +179,10 @@ func (a *RecordedApp) create(labels map[string]string, isDiffRun bool) error {
 			Labels: map[string]string{
 				KappIsAppLabelKey: kappIsAppLabelValue,
 			},
+			// Use app label as app change label for all new apps
+			Annotations: map[string]string{
+				KappIsAppLabelAsAppChangeLabelKey: "",
```
100mik: `KappIsAppLabelAsAppChangeLabelKey` => `KappChangesUseAppID` ?
I get that we are trying to say that the AppChange label uses the label that app resources have. I just think it can be simpler 🤔 
Any other ideas?

pkg/kapp/app/recorded_app.go
```diff
@@ -415,6 +427,9 @@ func (a *RecordedApp) Rename(newName string, newNamespace string) error {
 func (a *RecordedApp) renameConfigMap(app *corev1.ConfigMap, name, ns string) error {
 	oldName := app.Name
 
+	// Use app label as app change label for new, renamed or migrated apps
```
100mik: How about apps that have been migrated, but use app label name instead of ID right now. It is not super clear to me how we are handling those cases 🤔 
1. Will these only be renamed _if_ the user renames the app?
2. Does renaming the app to the same name rename CMs?

100mik (commented):
A few thoughts and I am thinking about whether migrated apps that use the label name today is being considered.

pkg/kapp/app/recorded_app.go
praveenrewar: > "kapp.k14s.io/uses-app-label" maybe? A bit of a tongue twister right now

Yeah, I agree, but "uses-app-label" might not be clearly indicating that this is for app changes.

> Or I guess changes-use-app-id is more appropriate

Hmm, I am not sure if we ever use `app-id` as a term for the app label value 🤔

pkg/kapp/app/recorded_app.go
praveenrewar: Sure, I can make the change. I think I kept it `changeLabelValue` only because the method is on RecordedApp, so it will become `app.appChangeLabelValue` in some places.

pkg/kapp/app/recorded_app.go
praveenrewar: Maybe "app-changes-use-app-label"?

pkg/kapp/app/recorded_app.go
100mik: I would prefer the latter because "app" and "appChange" are two separate values from my perspective 🤔

pkg/kapp/app/recorded_app.go
praveenrewar: Yeah, I was thinking of keeping it outside as well, the only reason I didn't do it is because if `!foundNonMigratedApp` then app would also be empty, but that should be okay, so I am totally fine with making this change.

pkg/kapp/app/recorded_app.go
praveenrewar: > but use app label name instead of ID

You mean app name instead of ID? If the app is _already_ migrated before deploying with newer version, then there won't be any change, but if migration happens with the newer version of kapp, then it will start using the app label instead of the name.

> Will these only be renamed if the user renames the app?

Yep. So, we will start using app labels instead of app name for either new apps or when a new meta configmap is being created (i.e. the existing app changes are already being abandoned), which happens in case of rename, migration and using `--prev-app` flag with deploy. I have mentioned all the scenarios in the table.

> Does renaming the app to the same name rename CMs?

We first create the new configmap and then delete the existing one, so if the user provides the same name as `--new-name` to `kapp rename` command, then it will through an error saying the configmap already exists.

pkg/kapp/app/recorded_app.go
praveenrewar: Makes sense.

pkg/kapp/app/recorded_app.go
rohitagg2020: `app-change-uses-app-label`?

100mik: That feels more comprehensive ^
Lets do this

pkg/kapp/app/recorded_app.go
100mik: `AppChangesUseAppLabel` in line with what we discussed earlier

100mik:
I see that you called out that we won't be any change for existing + migrated apps. That checks out.
The only sore point I have is this means that we will never entirely stop using the old way.
The changes seem to be in line with the behaviour that you are calling out.

What I was wondering is, since app-changes are something we garbage collect. Would it make sense to do this at an app-change level?
1. All new app-changes have app label
2. We query for both labels, _unless_ the `app-changes-use-app-label` annotation is set.
3. We set the annotation when we get an empty list when looking for app-changes with the old label

So the **huge** downside is we make one more call till the app is migrated. And from a kapp-controller perspective. For stable packages that do not have diffs, this might be forever till they are reinstalled.

I am trying to figure out how we can _potentially_ move to using *just* the new label. The path we are taking right now feels like we will be handling both cases for a long time since we cannot definitely said that no `migrated` or `existing` apps exist on the users cluster. And we end up leaving orphaned config maps when we do decide to move to the new one altogether.

Since, this is a case that is consciously not being handled in the PR. We can however track how we can "taper off"  using old labels as a separate issue since this fix might be needed sooner rather than later.

TL;DR: The path to us not supporting the old label at all in the future is not clear. But we might track that separately and improve on the cases that we do not handle.
@praveenrewar thoughts?

praveenrewar:
@100mik Thanks for bringing this up. I was planning to take that up separately, but now I think that we should at least decide how we migrate the app-changes for existing apps before making this change. 

How about we add one more label to app-changes (not label value, but new label itself). For new apps we will use the new label itself, for old apps we will use the old label but also add the new label to the app changes. After some releases we will switch to using the new label only.

New app-changes for existing apps:
```
labels:
    kapp.k14s.io/app-change-app: foo
    kapp.k14s.io/app-change-app-label: 101920101920
    kapp.k14s.io/is-app-change: ""
```

New app-changes for new apps:
```
labels:
    kapp.k14s.io/app-change-app-label: 101920101920
    kapp.k14s.io/is-app-change: ""
```

Does this make sense? I will see how these changes looks from code perspective.

pkg/kapp/app/recorded_app.go
100mik: Maybe do it after a check on `app` ? That would make sense I think

pkg/kapp/app/recorded_app.go
100mik: We can change this in line with the others !

pkg/kapp/app/recorded_app.go
100mik: The function `RenamePrevApp` seems to have similar logic and should also reflect this change?

100mik:
Yeah, I think adding both labels now does give us a part forward. We should do that if it is feasible 🤔

pkg/kapp/app/recorded_app.go
praveenrewar: > Maybe do it after a check on app ?

We return from the check on app.

100mik (approved):
The changes look good to me, adding both labels leave more room for us to plan out the migration!
I was wondering if there is a straightforward way of putting together tests for this, but I believe we do not have any form of tests involving older versions of `kapp`. So I am happy with these changes as long as we have verified the tabulated test cases. Let me know if you need a hand with the verification!

100mik (approved):
The additional condition makes sense!

pkg/kapp/app/recorded_app_changes.go
```diff
@@ -17,22 +17,24 @@ import (
...
 
-	appLabelValue string
+	appChangeUsesAppLabel bool
 
 	coreClient kubernetes.Interface
 }
 
-func NewRecordedAppChanges(nsName, appName, appLabelValue string, coreClient kubernetes.Interface) RecordedAppChanges {
-	return RecordedAppChanges{nsName, appName, appLabelValue, coreClient}
+func NewRecordedAppChanges(nsName, appName, changeLabelValue string, appChangeUsesAppLabel bool, coreClient kubernetes.Interface) RecordedAppChanges {
```
rohitagg2020, line 36:
```suggestion
func NewRecordedAppChanges(nsName, appName, appChangeLabelValue string, appChangeUsesAppLabel bool, coreClient kubernetes.Interface) RecordedAppChanges {
```

rohitagg2020 (approved):

pkg/kapp/app/recorded_app_changes.go
kumaritanushree: +1

praveenrewar: (I kept it `changeLabelValue` to be consistent with `changeLabelKey`)
Do you want me to make this change before merging?

PR #712, opened 2023-02-21
Use CreateOrUpdate for --prev-app flag

praveenrewar:
- Use CreateOrUpdate for --prev-app flag
- Remove RenamePrevApp method
- Do not list labeled resources for a new app when --prev-app flag is used

RenamePrevApp has the same functionality as CreateOrUpdate + logic for prev app, and having them as separate method is leading to increase in complexity when new changes are made because of the various different code paths that we need to cover.

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/app/recorded_app.go
+++ b/pkg/kapp/app/recorded_app.go
@@ -127,14 +127,13 @@ func (a *RecordedApp) UpdateUsedGVsAndGKs(gvs []schema.GroupVersion, gks []schem
 	})
 }
 
-func (a *RecordedApp) CreateOrUpdate(labels map[string]string, isDiffRun bool) (bool, error) {
+func (a *RecordedApp) CreateOrUpdate(prevAppName string, labels map[string]string, isDiffRun bool) (bool, error) {
 	defer a.logger.DebugFunc("CreateOrUpdate").Finish()
 
 	app, foundMigratedApp, err := a.find(a.fqName())
 	if err != nil {
 		return false, err
 	}
-
 	if foundMigratedApp {
 		a.isMigrated = true
 		return false, a.updateApp(app, labels)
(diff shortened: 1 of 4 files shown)


pkg/kapp/app/recorded_app.go
```diff
@@ -152,6 +152,32 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string, isDiffRun bool) (
...
+	if foundMigratedPrevApp {
+		a.isMigrated = true
+		return false, a.renameConfigMap(app, a.fqName(), a.nsName)
+	}
+
+	app, foundNonMigratedPrevApp, err := a.find(prevAppName)
+	if err != nil {
+		return false, err
+	}
+
```
rohitagg2020: nit: I think we can remove newline here as all this is for one scenario. I think same can be done for others as well.

praveenrewar: Done

pkg/kapp/app/recorded_app.go
```diff
@@ -144,14 +143,37 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string, isDiffRun bool) (
...
-
 	if foundNonMigratedApp {
 		if a.isMigrationEnabled() {
 			return false, a.migrate(app, labels, a.fqName())
 		}
 		return false, a.updateApp(app, labels)
 	}
 
+	if prevAppName == "" {
+		return true, a.create(labels, isDiffRun)
```
100mik, line 154: Should we move all the logic in lines 157-176 inside this block? 
Taking a closer look but it feels that we are always returning here if prevAppName is not empty. The logic which follows after would be unreachable and is only valid if `prevAppName` is not empty?

pkg/kapp/app/recorded_app.go
```diff
@@ -144,14 +143,37 @@ func (a *RecordedApp) CreateOrUpdate(labels map[string]string, isDiffRun bool) (
...
+	}
+
+	app, foundMigratedPrevApp, err := a.find(prevAppName + AppSuffix)
```
100mik, line 157: `prevAppName` would be empty here as we are returning above otherwise 🤔

pkg/kapp/app/recorded_app.go
praveenrewar: > The logic which follows after would be unreachable and is only valid if prevAppName is not empty?

Yep, that's the intention.

100mik:
Additionally, does it make sense to move the logic related to `prevApp` before the other stuff? 
Also, thinking of a weird state where app with new name exists in which case with current logic we end up ignoring `prevAppName`. If this is the the existing or expected behaviour we can keep the order as is. 
I think that is the case as we would pretend as if nothing is happening if app has already been moved from `prev-app`.

pkg/kapp/app/recorded_app.go
100mik: If we always return when `prevApp` name is not empty. When are the lines below ever helpful?

pkg/kapp/app/recorded_app.go
praveenrewar: At this stage, `app` is not present and `prevAppName` is _not_ empty.

praveenrewar:
> Also, thinking of a weird state where app with new name exists in which case with current logic we end up ignoring prevAppName. If this is the the existing or expected behaviour we can keep the order as is.

Yep, this is the expected behaviour. If you take a look at the removed `RenamePrevApp`, we were doing the same thing, we are not changing that functionality in this PR.

pkg/kapp/app/recorded_app.go
100mik: Oh yes, looks like I misread the condition above. Makes sense!

pkg/kapp/app/recorded_app.go
praveenrewar: > If we always return when prevApp name is not empty

We are returning when `prevAppName` _is_ empty 🤔

100mik (approved):
LGTM!

rohitagg2020 (approved):

PR #713, opened 2023-02-23
Moving steps to run kapp test in release.yml and revert goreleaser version

rohitagg2020:
Moving steps to run kapp test in release.yml and revert goreleaser version

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,8 +24,8 @@ jobs:
           go-version: 1.19.5
 
       - name: Run GoReleaser
-        # GoReleaser v4.2.0
-        uses: goreleaser/goreleaser-action@f82d6c1c344bcacabba2c841718984797f664a6b
+        # GoReleaser v2.5.0
+        uses: goreleaser/goreleaser-action@5e15885530fb01d81d1f24e8a6f54ebbd0fed7eb
         if: startsWith(github.ref, 'refs/tags/')
         with:
           version: 0.184.0
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,33 +33,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Setup Minikube
-        run: |
-          curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
-          install minikube-linux-amd64 /usr/local/bin/minikube
-          minikube start --driver=docker --memory 4096
-          eval $(minikube docker-env --shell=bash)
-
-          # Ensure that there is no existing kapp installed
-          rm -f /tmp/bin/kapp
-
-      - name: Create kapp binary
-        run: |
-          set -e -x
-          VERSION=`echo ${{ github.ref }}  | grep -Eo '[0-9].*'`
-          ./hack/build.sh
-
-      - name: Add kapp binary to the Path
-        run: |
-          mkdir bin
-          mv kapp bin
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -122,4 +95,28 @@ jobs:
...
+          ./hack/build.sh
+
+          # Add binary to the path
+          mkdir bin
+          mv kapp bin
+          PATH=$PATH:$PWD/bin
+          echo $PATH
+
+          # Run test cases
+          ./hack/test-external.sh
```
100mik, line 122: Add a new line at the end of file?

.github/workflows/release.yml
```diff
@@ -122,4 +95,28 @@ jobs:
...
           )
-          
\ No newline at end of file
+          
+      - name: Run Test cases
+        run: |
+          # Setup minikube
+          curl -LO https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64
+          install minikube-linux-amd64 /usr/local/bin/minikube
+          minikube start --driver=docker --memory 4096
```
100mik, line 104: I am fine with using Minikube if all the other tools are, but we should probably switch to kind at some point

.github/workflows/release.yml
```diff
@@ -24,42 +24,15 @@ jobs:
           go-version: 1.19.5
 
       - name: Run GoReleaser
-        # GoReleaser v4.2.0
-        uses: goreleaser/goreleaser-action@f82d6c1c344bcacabba2c841718984797f664a6b
+        # GoReleaser v2.5.0
```
100mik, line 27: Is the downgrade intentional? I remember Praveen mentioning that there is another PR up for this, if that is the case this looks good!

.github/workflows/release.yml
rohitagg2020: May be we can have separate PR for that.

.github/workflows/release.yml
rohitagg2020: Yes, We have seen issues with alpha releases with `4.2.0`.

.github/workflows/release.yml
rohitagg2020: Not necessarily. Deliberately removed it as it was difficult to remember it every time so that diff didn't come.

kumaritanushree:
LGTM

.github/workflows/release.yml
praveenrewar: > I am fine with using Minikube if all the other tools are

We actually don't run the tests during a release in other tools.

praveenrewar (approved):
LGTM!
If this is tested on a fork, then we can merge it.

.github/workflows/release.yml
100mik: @rohitagg2020 I would insist on this, mainly because it sets GitHub off and something we stick to.
the formatter would be doing this for all the `.go` files already.

praveenrewar: There *is* a line at the end of the file. Am I missing something here?

rohitagg2020: I have checked just now. @100mik There were two new lines at the end. I have removed one extra line at the end. We already have files ending like these in the other [workflows](https://github.com/carvel-dev/kapp/tree/develop/.github/workflows)

100mik: Ah alright, I think the file didn't have one before, GitHub threw me off with a warning.

100mik (approved):
LGTM!

PR #731, opened 2023-04-03
Improving performance enhancement specifically during diff stage

rohitagg2020:
Improving performance enhancement specifically during diff stage.

By `go profiling`, we have noticed that there are too many calls to [deepCopy](https://github.com/carvel-dev/kapp/blob/b75689aedd76c2e022a8497473acd1e5facf2210/pkg/kapp/resources/resource.go#L236) and [AsYAMLBytes](https://github.com/carvel-dev/kapp/blob/develop/pkg/kapp/resources/resource.go#L248)

**Improvements done:**
1. While rebas'ing a resource, we wont do `Deepcopy` until the resource being rebased matches with the resource for which rebase resource is defined.
2. `default.go` which was already in yaml was unnecessary being converted to struct and then back to yaml. Removed this unnecessary conversion.
3. Reduced calls to `AsYAMLBytes`.

**Only one breaking scenario w.r.t. changes proposed:**
1. kapp deploy (config with deployment) with v0.15.0 or earlier. Most of the times, revision is not set while storing md5. But sometimes it can. Lets move forward with scenario where it can.
2. There has been no changes.
3. kapp deploy (config with deployment) with these changes. In this case, the user will see the diff’s.

The chances of this scenario are extremely low…

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/config/default.go
+++ b/pkg/kapp/config/default.go
@@ -700,10 +700,18 @@ changeRuleBindings:
       - hasNamespaceMatcher: {}
 `
 
-var defaultConfigRes = ctlres.MustNewResourceFromBytes([]byte(defaultConfigYAML))
-
 func NewDefaultConfigString() string { return defaultConfigYAML }
 
 func NewConfFromResourcesWithDefaults(resources []ctlres.Resource) ([]ctlres.Resource, Conf, error) {
-	return NewConfFromResources(append([]ctlres.Resource{defaultConfigRes}, resources...))
+	resources, conf, err := NewConfFromResources(resources)
+	if err != nil {
+		return nil, Conf{}, err
+	}
+
+	defaultConfig, err := newConfigFromYAMLBytes([]byte(defaultConfigYAML), "config/default (kapp.k14s.io/v1alpha1)")
+	if err != nil {
+		return nil, Conf{}, err
+	}
+
+	return resources, Conf{append([]Config{defaultConfig}, conf.configs...)}, err
 }
--- a/pkg/kapp/diff/change.go
+++ b/pkg/kapp/diff/change.go
@@ -45,6 +45,7 @@ type ChangeImpl struct {
 
 	configurableTextDiff *ConfigurableTextDiff
 	opsDiff              *OpsDiff
+	changeOpVal          ChangeOp
 }
 
 var _ Change = &ChangeImpl{}
@@ -86,6 +87,13 @@ func (d *ChangeImpl) AppliedResource() ctlres.Resource         { return d.applie
 func (d *ChangeImpl) ClusterOriginalResource() ctlres.Resource { return d.clusterOriginalRes }
 
 func (d *ChangeImpl) Op() ChangeOp {
+	if d.changeOpVal == "" {
+		d.changeOpVal = d.op()
+	}
+	return d.changeOpVal
+}
+
+func (d *ChangeImpl) op() ChangeOp {
 	if d.newRes != nil {
 		if _, hasNoopAnnotation := d.newRes.Annotations()[ctlres.NoopAnnKey]; hasNoopAnnotation {
 			return ChangeOpNoop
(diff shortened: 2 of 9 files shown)


pkg/kapp/diff/resource_with_history.go
```diff
@@ -153,10 +153,10 @@ func (r ResourceWithHistory) recalculateLastAppliedChange() ([]Change, string, s
 	// Continue to calculate historyless change with excluded fields
 	// (previous kapp versions did so, and we do not want to fallback
 	// to diffing against list resources).
-	recalculatedChange1, err := r.calculateChangePrev1(lastAppliedRes)
-	if err != nil {
-		return nil, "", "" // TODO deal with error?
-	}
+	// recalculatedChange1, err := r.calculateChangePrev1(lastAppliedRes)
```
rohitagg2020: I will remove it..

kumaritanushree:
LGTM

pkg/kapp/diff/rebased_resource.go
```diff
@@ -42,17 +42,19 @@ func (r RebasedResource) Resource() (ctlres.Resource, error) {
...
-			ctlres.FieldCopyModSourceNew:      r.newRes.DeepCopy(),
-			ctlres.FieldCopyModSourceExisting: r.existingRes.DeepCopy(),
-			// Might be useful for more advanced rebase rules like ytt-based
-			ctlres.FieldCopyModSource("_current"): result.DeepCopy(),
-		}
-
-		err := t.ApplyFromMultiple(result, resSources)
-		if err != nil {
-			return nil, fmt.Errorf("Applying rebase rule to %s: %w", resultDesc, err)
+		if t.IsResourceMatching(result) {
```
praveenrewar, line 45: Should we do this inline instead?
```go
        if res == nil || !t.ResourceMatcher.Matches(res) {
		continue
	}
```

pkg/kapp/diff/change.go
```diff
@@ -86,6 +87,13 @@ func (d *ChangeImpl) AppliedResource() ctlres.Resource         { return d.applie
 func (d *ChangeImpl) ClusterOriginalResource() ctlres.Resource { return d.clusterOriginalRes }
 
 func (d *ChangeImpl) Op() ChangeOp {
+	if d.changeOpVal == "" {
+		d.changeOpVal = d.op()
```
praveenrewar, line 91: I am trying to think if caching ChangeOp can have any side effects...

pkg/kapp/config/default.go
```diff
@@ -700,10 +703,24 @@ changeRuleBindings:
       - hasNamespaceMatcher: {}
 `
 
-var defaultConfigRes = ctlres.MustNewResourceFromBytes([]byte(defaultConfigYAML))
-
 func NewDefaultConfigString() string { return defaultConfigYAML }
 
 func NewConfFromResourcesWithDefaults(resources []ctlres.Resource) ([]ctlres.Resource, Conf, error) {
-	return NewConfFromResources(append([]ctlres.Resource{defaultConfigRes}, resources...))
```
praveenrewar, line 708: Does this help in significant performance improvement? (I hope we are not reducing code readability for an insignificant amount of improvement)

pkg/kapp/diff/resource_with_history.go
praveenrewar: Could you also remove the comments for the removed method `// Continue to calculate historyless change...`

pkg/kapp/config/default.go
rohitagg2020: Yes. In profiling as well, we saw that this was contributing considerable time to the CPU processing.

pkg/kapp/diff/change.go
rohitagg2020: I have checked the code. It doesnt feel like there will be any side effects. Maybe we can discuss more on it.

pkg/kapp/diff/rebased_resource.go
rohitagg2020: `t` is an interface. Thus we cant access the ResourceMatcher object directly.

pkg/kapp/diff/resource_with_history.go
rohitagg2020: Done

praveenrewar (approved):

Issue #729, opened 2023-03-28
kapp unblocks dependent changes if the changes which are depended upon already exists


praveenrewar:
**What steps did you take:**
`jobs.yaml`
```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: successful-job-1
  annotations:
    kapp.k14s.io/change-rule: "upsert after upserting abc"
spec:
  template:
    metadata:
      name: successful-job-1
    spec:
      containers:
        - name: successful-job-1
          image: busybox
          command: ["/bin/sh", "-c", "sleep 10"]
      restartPolicy: Never
---
apiVersion: batch/v1
kind: Job
metadata:
  name: successful-job-2
  annotations:
    kapp.k14s.io/change-group: "abc"
spec:
  template:
    metadata:
      name: successful-job-2
    spec:
      containers:
        - name: successful-job-2
          image: busybox
          command: ["/bin/sh", "-c", "sleep 10"]
      restartPolicy: Never
```
Deployed these jobs and stopped in between to create only job 1 without waiting for it.
```bash
kapp deploy -a order -f order.yaml
Target cluster 'https://192.168.64.12:8443' (nodes: minikube)

Changes

Namespace  Name              Kind  Age  Op      Op st.  Wait to    Rs  Ri
default    successful-job-1  Job   -    create  -       reconcile  -   -
^          successful-job-2  Job   -    create  -       reconcile  -   -

Op:      2 create, 0 delete, 0 update, 0 noop, 0 exists
Wait to: 2 reconcile, 0 delete, 0 noop

Continue? [yN]: y

6:28:20PM: ---- applying 1 changes [0/2 done] ----
6:28:20PM: create job/successful-job-2 (batch/v1) namespace: default
6:28:20PM: ---- waiting on 1 changes [0/2 done] ----
6:28:20PM: ongoing: reconcile job/successful-job-2 (batch/v1) namespace: default
6:28:20PM:  ^ Waiting to complete (0 active, 0 failed, 0 succeeded)
6:28:20PM:  L ongoing: waiting on pod/successful-job-2-jlwbz (v1) namespace: default
6:28:20PM:     ^ Pending: ContainerCreating
^C
```
Deploy the jobs again
```bash
kapp deploy -a order -f order.yaml
Target cluster 'https://192.168.64.12:8443' (nodes: minikube)

Changes

Namespace  Name              Kind  Age  Op      Op st.  Wait to    Rs       Ri
[...]

PR #760, opened 2023-05-30
feat: new config for excluding fields on history-less resources

### LINKED ISSUE #746: diff in Status field causes conflicts during apply stage in newer version of kapp
**What steps did you take:**
I tried deploying a resource and I noticed that there was some diff present for the status field, when I said yes to the prompt and kapp started applying the changes, it threw a conflict error suggesting that the diff has changed.

**What happened:**
The status of the resource changed in the background while kapp was trying to apply the changes which caused an update in the diff.

**What did you expect:**
I expected kapp to ignore the status completely during diffs.

In earlier versions of kapp (< v0.55.0), we used to copy the entire status field, but starting v0.55.0, we stopped doing that and instead used diffAgainstLastAppliedFieldExclusionRules for status field, which works fine during smart diff, but shows the diff when non-smart diff is being used.

Relavant slack thread: https://kubernetes.slack.com/archives/CH8KCCKA5/p1682582134618589

**Environment:**

- kapp version (use `kapp --version`): v0.55.0

---
Vote on this request

theurichde:

This PR introduces a new diff field exclusion configuration for resources without a valid history

Fixes #746 

```release-note
New diff field exclusion configuration for resources without a valid history
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/config/default.go
+++ b/pkg/kapp/config/default.go
@@ -231,6 +231,7 @@ diffAgainstLastAppliedFieldExclusionRules:
 - path: [metadata, annotations, "deployment.kubernetes.io/revision"]
   resourceMatchers: *appsV1DeploymentWithRevAnnKey
 
+diffAgainstExistingFieldExclusionRules:
 - path: [status]
   resourceMatchers:
   - allMatcher: {}
--- a/pkg/kapp/diff/resource_with_history.go
+++ b/pkg/kapp/diff/resource_with_history.go
@@ -36,9 +36,8 @@ func NewResourceWithHistory(resource ctlres.Resource,
 
 	return ResourceWithHistory{resource.DeepCopy(), changeFactory, diffAgainstLastAppliedFieldExclusionMods}
 }
-
-func (r ResourceWithHistory) HistorylessResource() (ctlres.Resource, error) {
-	return resourceWithoutHistory{r.resource}.Resource()
+func NewResourceWithoutHistory(resource ctlres.Resource, fieldExclusionMods []ctlres.FieldRemoveMod) ResourceWithoutHistory {
+	return ResourceWithoutHistory{res: resource, fieldExclusionMods: fieldExclusionMods}
 }
 
 // LastAppliedResource will return "last applied" resource that was saved
@@ -156,12 +155,12 @@ func (r ResourceWithHistory) recalculateLastAppliedChange() ([]Change, string, s
 
 func (r ResourceWithHistory) newExactHistorylessChange(existingRes, newRes ctlres.Resource) (Change, error) {
 	// If annotations are not removed line numbers will be mismatched
-	existingRes, err := resourceWithoutHistory{existingRes}.Resource()
+	existingRes, err := ResourceWithoutHistory{existingRes, nil}.Resource()
 	if err != nil {
 		return nil, err
 	}
 
-	newRes, err = resourceWithoutHistory{newRes}.Resource()
+	newRes, err = ResourceWithoutHistory{newRes, nil}.Resource()
 	if err != nil {
 		return nil, err
 	}
@@ -169,11 +168,12 @@ func (r ResourceWithHistory) newExactHistorylessChange(existingRes, newRes ctlre
 	return r.changeFactory.NewExactChange(existingRes, newRes)
 }
 
-type resourceWithoutHistory struct {
-	res ctlres.Resource
+type ResourceWithoutHistory struct {
+	res                ctlres.Resource
+	fieldExclusionMods []ctlres.FieldRemoveMod
 }
 
-func (r resourceWithoutHistory) Resource() (ctlres.Resource, error) {
+func (r ResourceWithoutHistory) Resource() (ctlres.Resource, error) {
 	res := r.res.DeepCopy()
 
 	for _, t := range r.removeAppliedResAnnKeysMods() {
@@ -183,10 +183,17 @@ func (r resourceWithoutHistory) Resource() (ctlres.Resource, error) {
 		}
 	}
 
+	for _, t := range r.fieldExclusionMods {
+		err := t.Apply(res)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return res, nil
 }
 
-func (resourceWithoutHistory) removeAppliedResAnnKeysMods() []ctlres.ResourceMod {
+func (ResourceWithoutHistory) removeAppliedResAnnKeysMods() []ctlres.ResourceMod {
 	return []ctlres.ResourceMod{
 		ctlres.FieldRemoveMod{
 			ResourceMatcher: ctlres.AllMatcher{},
--- a/pkg/kapp/cmd/app/delete.go
+++ b/pkg/kapp/cmd/app/delete.go
@@ -233,7 +233,7 @@ func (o *DeleteOptions) calculateAndPresentChanges(existingResources []ctlres.Re
 	)
 
 	{ // Figure out changes for X existing resources -> 0 new resources
-		changeFactory := ctldiff.NewChangeFactory(nil, nil)
+		changeFactory := ctldiff.NewChangeFactory(nil, nil, nil)
 		changeSetFactory := ctldiff.NewChangeSetFactory(o.DiffFlags.ChangeSetOpts, changeFactory)
 
 		changes, err := changeSetFactory.New(existingResources, nil).Calculate()
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -405,7 +405,7 @@ func (o *DeployOptions) calculateAndPresentChanges(existingResources,
 	var clusterChangeSet ctlcap.ClusterChangeSet
 
 	{ // Figure out changes for X existing resources -> X new resources
-		changeFactory := ctldiff.NewChangeFactory(conf.RebaseMods(), conf.DiffAgainstLastAppliedFieldExclusionMods())
+		changeFactory := ctldiff.NewChangeFactory(conf.RebaseMods(), conf.DiffAgainstLastAppliedFieldExclusionMods(), conf.DiffAgainstExistingFieldExclusionMods())
 		changeSetFactory := ctldiff.NewChangeSetFactory(o.DiffFlags.ChangeSetOpts, changeFactory)
 
 		err := ctldiff.NewRenewableResources(existingResources, newResources).Prepare()
--- a/pkg/kapp/cmd/app/inspect.go
+++ b/pkg/kapp/cmd/app/inspect.go
@@ -97,7 +97,7 @@ func (o *InspectOptions) Run() error {
 	switch {
 	case o.Raw:
 		for _, res := range resources {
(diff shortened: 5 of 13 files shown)


theurichde:
Hello @praveenrewar! 

I would like to have a brief review on this draft. I would like to ensure, that I am on the right track here. Unit tests are green, e2e tests are running right at the moment.

Of course, I will add unit tests for this specific feature, too.

Additionally, I would like to refactor it a bit more, but first let's see if the overall approach is the right one ;-)

praveenrewar:
Hey @theurichde! Thank you so much for working on this ❤️  
At first glance, it looks good to me, I will take a closer look at it in the morning.

praveenrewar:
Hi @theurichde, the overall approach looks good to me, as you have mentioned we could do some refactoring, but I think we are in the right direction 🚀 

We are completely ignoring the status field from diff, but we should probably ensure that `status` is available for ytt rebase rules and other places where it is used to determine if the resource is in the desired state (the changes shouldn't effect this functionality as we have an original copy of the resource for other uses but we should double check it to be on the safe side). The e2e tests should cover most of these scenarios except for ytt rebase rules using status field.

pkg/kapp/config/default.go
```diff
@@ -235,6 +235,11 @@ diffAgainstLastAppliedFieldExclusionRules:
   resourceMatchers:
   - allMatcher: {}
 
+diffWithoutHistoryFieldExclusionRules:
```
praveenrewar: Should we call it `diffAgainstExistingFieldExclusionRules`? Since this is a new configuration, it could also be used by users and `WithoutHistory` seems like an internal implementation.

pkg/kapp/diff/resource_with_history.go
```diff
@@ -156,24 +155,25 @@ func (r ResourceWithHistory) recalculateLastAppliedChange() ([]Change, string, s
...
 	if err != nil {
 		return nil, err
 	}
 
 	return r.changeFactory.NewExactChange(existingRes, newRes)
 }
 
-type resourceWithoutHistory struct {
-	res ctlres.Resource
+type ResourceWithoutHistory struct {
```
praveenrewar, line 171: Maybe we can keep it private and pass down the exclusion rules through `ResourceWithHistory` itself?

theurichde: We can't keep it private as it is used in `inspect.go:100` - but if it is really a thorn in the flesh, I can try to refactor it.

praveenrewar: We were using the HistoryLessResource method in inspect.go which would return a ctlres.Resource, so we wouldn’t need this as public. But, I am good with this change as well, the main reason I wanted to keep it as it is to have minimal changes in the PR.

pkg/kapp/config/default.go
```diff
@@ -235,6 +235,11 @@ diffAgainstLastAppliedFieldExclusionRules:
   resourceMatchers:
   - allMatcher: {}
 
+diffAgainstExistingFieldExclusionRules:
+- path: [status]
```
praveenrewar, line 235: Now that we have added status field here, we can remove it from under `diffAgainstLastAppliedFieldExclusionRules` as it won't have any benefits.

test/e2e/custom_wait_rules_test.go
```diff
@@ -28,7 +28,9 @@ waitRules:
         resource.star: |
           def is_done(resource):
               state = resource.status.currentState
-              if state == "Failed":
+              if state == "Progressing":
+                return {"done": False, "successful": True, "message": "Current state as Failed"}
```
praveenrewar: nit: "Current state as Progressing"

test/e2e/custom_wait_rules_test.go
```diff
@@ -98,8 +105,28 @@ status:
...
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{
+			StdinReader: strings.NewReader(crdYaml + crYaml)})
+
+		patch := `[{ "op": "add", "path": "/status", "value": {"currentState": "Progressing"}}]`
+		PatchClusterResource("CronTab", "my-new-cron-object-1", env.Namespace, patch, kubectl)
+
+		go func() {
+			for {
+				// make sure the waitRule is waiting for the resource before patching it to "Running"
+				time.Sleep(5 * time.Second)
```
praveenrewar:
```suggestion
				time.Sleep(1 * time.Second)
```

test/e2e/custom_wait_rules_test.go
```diff
@@ -222,34 +273,38 @@ metadata:
...
+Op:      1 create, 0 delete, 0 update, 1 noop, 0 exists
 Wait to: 2 reconcile, 0 delete, 0 noop
 
-<replaced>: ---- applying 1 changes [0/2 done] ----
-<replaced>: create crontab/my-new-cron-object-1 (stable.example.com/v1) namespace: kapp-test
-<replaced>: ---- waiting on 1 changes [0/2 done] ----
-<replaced>: ongoing: reconcile crontab/my-new-cron-object-1 (stable.example.com/v1) namespace: kapp-test
-<replaced>:  ^ Allowing blocked changes to proceed: Unblock blocked changes
-<replaced>: ---- applying 1 changes [1/2 done] ----
+<replaced>: ---- applying 2 changes [0/2 done] ----
```
praveenrewar: We want to test that when the status reaches Progressing for the CR, we apply the ConfigMap, but due to #729 that doesn't happen, the ConfigMap gets applied before that.

praveenrewar: I just realised that we can use default values for these fields in the CRD definition instead of patching them initially. Something like this:
```yaml
            status:
              type: object
              default: {}
              properties:
                currentState:
                  type: string
                  default: Progressing
```

praveenrewar: ^ We can do this in other tests as well.

theurichde: I just tried that out, but it's not working as expected. The CRD has the default value set, but it's not applied to the CR 
when deployed via kapp 😢 

**edit:** That works well! I missed the empty default object on type object. Awesome!

test/e2e/custom_wait_rules_test.go
theurichde: I hope the test won't flap with just 1s wait time 😅

pkg/kapp/config/default.go
theurichde: aaand it's gone

test/e2e/custom_wait_rules_test.go
theurichde: > We want to test that when the status reaches Progressing for the CR, we apply the ConfigMap, but due to #729 that doesn't happen, the ConfigMap gets applied before that.

So, there is no ToDo for this right now on our side, right?

praveenrewar: Yesh, I think we are good 🚀 
Update: (Sorry, still taking a look at the changes, I think we need to make a small change)

test/e2e/custom_wait_rules_test.go
```diff
@@ -222,34 +260,37 @@ metadata:
...
+		go patchConfigMap()
 
-		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{
-			StdinReader: strings.NewReader(yaml)})
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{StdinReader: strings.NewReader(cr)})
+
+		//remove age column from output, as we can't rely on it
+		columns := "namespace,name,kind,op,'op strategy','wait to','reconcile state','reconcile info'"
+		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "--column", columns}, RunOpts{
+			StdinReader: strings.NewReader(yaml + cr)})
```
praveenrewar: We don't need to deploy twice now as the default value for status would be populated and we don't need the first patch.
Also, we are not asserting on the `Age` column and it's just empty so I am slightly confused with the comment `remove age column from output, as we can't rely on it`
```suggestion
		out, _ := kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{StdinReader: strings.NewReader(yaml + cr)})
```

test/e2e/custom_wait_rules_test.go
```diff
@@ -152,15 +186,31 @@ spec:
...
                 currentState:
                   type: string
+                  default: Progressing
   scope: Namespaced
   names:
     plural: crontabs
     singular: crontab
     kind: CronTab`
 
+	cr := `
```
praveenrewar: nit: Since we are back to deploying `yaml` and `cr` together, maybe we could revert it and move this back to yaml? (I am just trying to keep the changes in the PR to a minimum, I apologise for the nit-picking, feel free to ignore it)

test/e2e/custom_wait_rules_test.go
praveenrewar: Added a comment above, we should be good once we make that change.

test/e2e/custom_wait_rules_test.go
```diff
@@ -98,8 +107,25 @@ status:
 	defer cleanUp()
 
 	logger.Section("deploy resource with current state as running", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name}, RunOpts{
```
praveenrewar: I think we can just deploy the crd with a default value of Running.

praveenrewar (commented):
I feel that the custom_wait_rules_test and order_test are creating a hindrance in this PR and they are not even related to the changes being made. Do you want to pair to quickly wrap this or would you be comfortable if I made a separate (quick) PR to refactor them so they are not effected by the changes being made in this PR (You have spent some time working on them so it's totally fine if you would want to continue and finish it here itself :) )?

praveenrewar:
> Do you want to pair to quickly wrap this or would you be comfortable if I made a separate (quick) PR to refactor them so they are not effected by the changes being made in this PR

I just created a [PR](https://github.com/carvel-dev/kapp/pull/782) to refactor those tests, let me know if you would want me to merge it and rebase this PR on top of it or feel free to copy those changes.

test/e2e/custom_wait_rules_test.go
theurichde: In the actual output we would have the age, which can differ from time to time, which would lead to a flaky test

test/e2e/custom_wait_rules_test.go
theurichde: Sure thing! Nit-picking is also welcome, no problem 😊

theurichde:
> want me to merge it and rebase this PR on top of it

Okidoki, go for it 😍

praveenrewar:
> Okidoki, go for it 😍

Done 🚀

100mik (approved):
LGTM!

praveenrewar (approved):
Thank you so much for working on this @theurichde 🙇🏻

PR #799, opened 2023-08-14
Add fs.FS support to deploy, inspect, diff

### LINKED ISSUE #800: Allow kapp deploy go code to be used with an fs.FS
I am doing a POC where I want to call `app.NewDeployCmd()` and then execute it programatically. I also want to pass it an `fs.FS` implementation because the content I want to apply to my cluster is coming from an in-memory file system, not from the OS.

The ability to pass an `fs.FS` when invoking deploy programmatically.

No

---
Vote on this request

ncdc:

Allow deploy, inspect, and diff to use an `fs.FS` for the filesystem so they can be invoked programmatically with filesystems other than just the normal OS one.

Fixes #800

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/resources/file_resources.go
+++ b/pkg/kapp/resources/file_resources.go
@@ -113,7 +113,7 @@ func isDir(fsys fs.FS, path string) (bool, error) {
 	if fsys == nil {
 		fileInfo, err := os.Stat(path)
 		if err != nil {
-			return false, fmt.Errorf("error stat'ing file %q: %v", path, err)
+			return false, err
 		}
 		return fileInfo.IsDir(), nil
 	}
@@ -122,7 +122,7 @@ func isDir(fsys fs.FS, path string) (bool, error) {
 	case fs.StatFS:
 		fileInfo, err := t.Stat(path)
 		if err != nil {
-			return false, fmt.Errorf("error stat'ing file %q: %v", path, err)
+			return false, err
 		}
 		return fileInfo.IsDir(), nil
 	case fs.FS:
@@ -134,7 +134,7 @@ func isDir(fsys fs.FS, path string) (bool, error) {
 
 		fileInfo, err := f.Stat()
 		if err != nil {
-			return false, fmt.Errorf("error stat'ing file %q: %v", path, err)
+			return false, err
 		}
 		return fileInfo.IsDir(), nil
 	default:
--- a/pkg/kapp/resources/file_resources.go
+++ b/pkg/kapp/resources/file_resources.go
@@ -5,6 +5,7 @@ package resources
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"sort"
@@ -19,7 +20,12 @@ type FileResource struct {
 	fileSrc FileSource
 }
 
-func NewFileResources(file string) ([]FileResource, error) {
+// NewFileResources inspects file and returns a slice of FileResource objects. If file is "-", a FileResource for STDIN
+// is returned. If it is prefixed with either http:// or https://, a FileResource that supports an HTTP transport is
+// returned. If file is a directory, one FileResource object is returned for each file in the directory with an allowed
+// extension (.json, .yml, .yaml). If file is not a directory, a FileResource object is returned for that one file. If
+// fsys is nil, NewFileResources uses the OS's file system. Otherwise, it uses the passed in file system.
+func NewFileResources(fsys fs.FS, file string) ([]FileResource, error) {
 	var fileRs []FileResource
 
 	switch {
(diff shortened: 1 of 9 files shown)


100mik:
I am going to take a closer look at this in my daytime tomorrow. 

I was wondering if you could create an issue illustrating how this change helps and elaborating on your use case? That context would definitely be helpful!

ncdc:
I created #800 for this, thanks!

pkg/kapp/resources/file_resources.go
```diff
@@ -30,16 +36,22 @@ func NewFileResources(file string) ([]FileResource, error) {
...
 		}
 
-		if fileInfo.IsDir() {
-			var paths []string
+		if dir {
+			// The typical command line invocation won't set fsys. If it comes in nil, create a new DirFS rooted at
+			// file, then set file to '.' (current working directory) so the fs.WalkDir call below works correctly.
+			if fsys == nil {
+				fsys = os.DirFS(file)
+				file = "."
```
praveenrewar, line 49: Nice hack 🚀

pkg/kapp/resources/file_resources.go
```diff
@@ -94,3 +106,38 @@ func (r FileResource) Resources() ([]Resource, error) {
...
 	return resources, nil
 }
+
+// isDir returns if path is a directory. If fsys is nil, isDir calls os.Stat(path); otherwise, it checks path inside
+// fsys.
+func isDir(fsys fs.FS, path string) (bool, error) {
+	if fsys == nil {
+		fileInfo, err := os.Stat(path)
+		if err != nil {
+			return false, fmt.Errorf("error stat'ing file %q: %v", path, err)
```
praveenrewar: **nit**: Since the error from os.Stat() is of the format `stat <file-name>: error`, we could directly return err instead.
(Same in other places in isDir)

test/e2e/deploy_filesystem_test.go
```diff
@@ -0,0 +1,161 @@
...
+		"file1.yaml",      // file in root of fs
+		"dir2/file2.yaml", // file in subdir
+		"dir3",            // another directory
+	}
+
+	flagsFactory := core.NewFlagsFactory(configFactory, depsFactory)
+
+	deployCmd := app.NewDeployCmd(deployOptions, flagsFactory)
+
+	now := time.Now().Unix()
```
praveenrewar, line 72: **question**: I was really curious about why a label was being added and then noticed that we are looking for the cms in all the namespaces during validation and hence to avoid any other cms present, using a label makes sense. But then I got curious about why we need to look into all the namespaces, maybe just the env.Namespace should be enough? (Or maybe I am missing some use case that we are trying to cover)

test/e2e/deploy_filesystem_test.go
```diff
@@ -0,0 +1,161 @@
...
+	const allNamespaces = ""
+	configMaps, err := kubeClient.CoreV1().ConfigMaps(allNamespaces).List(ctx, metav1.ListOptions{LabelSelector: labelSelector})
+	require.NoError(t, err, "error listing ConfigMaps")
+	require.Len(t, configMaps.Items, 6)
+	expectedNames := sets.New[string]("cm1", "cm2", "cm3", "cm4", "cm5", "cm6")
+	actualNames := sets.New[string]()
+	for _, cm := range configMaps.Items {
+		actualNames.Insert(cm.Name)
+	}
+	require.Empty(t, cmp.Diff(expectedNames, actualNames))
```
praveenrewar: **nit**: Maybe we can use `require.ElementsMatch`? (We would still need to create an array of actualNames and expectedNames, but we won't have to use any other library)

test/e2e/deploy_filesystem_test.go
```diff
@@ -0,0 +1,161 @@
...
+		).
+		file(
+			"dir3/cm6.yaml",
+			testConfigMap(env.Namespace, "cm6", now),
+		).
+		toFS()
+
+	deployOptions.FileSystem = inMemFS
+
+	err := deployCmd.Execute()
```
praveenrewar: Could you add another Execute() to have test for update scenario as well?

praveenrewar (commented):
Thank you so much for the PR @ncdc ❤️ 
Overall it's looking really good (just left some minor comments).

test/e2e/deploy_filesystem_test.go
ncdc: I can switch it to just look in env.Namespace

test/e2e/deploy_filesystem_test.go
ncdc: cmp.Diff is already in the repo, although it's only indirectly used before this change. Using cmp.Diff in tests provides extremely clean, easy to understand reasons why a test failed.

```go
m1 := []string{"a", "b"}
m2 := []string{"a", "c"}
assert.ElementsMatch(t, m1, m2)
assert.Empty(t, cmp.Diff(m1, m2))
```

The ElementsMatch error yields:
```
 Error:          elements differ
                                
                                extra elements in list A:
                                ([]interface {}) (len=1) {
                                 (string) (len=1) "b"
                                }
                                
                                
                                extra elements in list B:
                                ([]interface {}) (len=1) {
                                 (string) (len=1) "c"
                                }
                                
                                
                                listA:
                                ([]string) (len=2) {
                                 (string) (len=1) "a",
                                 (string) (len=1) "b"
                                }
                                
                                
                                listB:
                                ([]string) (len=2) {
                                 (string) (len=1) "a",
                                 (string) (len=1) "c"
                                }
```

The cmp.Diff error yields
```
Error:          Should be empty, but was   []string{
                                        "a",
                                -       "b",
                                +       "c",
                                  }

```

praveenrewar (approved):
LGTM. 
@100mik Let me know if you are planning to take a look at it.

100mik:
Thanks for the PR again!

PR #807, opened 2023-08-28
Do not create change when app-changes-max-to-keep=0

### LINKED ISSUE #806: Setting --app-changes-max-to-keep to 0 shouldn't create and delete an app change
**What steps did you take:**
- Ran `kapp deploy -a my-app -f my-file --app-changes-max-to-keep=0` with a user that doesn't have delete permissions on configmaps.
- Ran `kapp delete -a my-app` with the same user.

**What happened:**
- App change was created even though I set the cap to 0. If the user has delete permissions, the app change is deleted at the end of deployment.
- kapp delete errors out saying unable to delete the app change as delete permissions are not provided. We want to keep the delete permissions to a minimum, so we can provide delete permissions for the meta configmap based on it's name, but not for the app change, as it uses generatedName.

**What did you expect:**
I expected kapp to not create an app change when --app-changes-max-to-keep is set to 0.

**Environment:**

- kapp version (use `kapp --version`):

---
Vote on this request

praveenrewar:
Do not create change when app-changes-max-to-keep=0.
Existing behaviour for --app-changes-max-to-keep=0:
- Create an app change before deployment begins, and garbage collect it at the end

New behaviour: 
- Do not create an app change at all, but still keep the last change details in the meta configmap.

If app changes are already present for an app, and we set the flag value to 0, then it would still need to clean up old app changes.

Fixes #806 

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -220,10 +220,11 @@ func (o *DeployOptions) Run() error {
 	}()
 
 	touch := ctlapp.Touch{
-		App:              app,
-		Description:      "update: " + changeSummary,
-		Namespaces:       nsNames,
-		IgnoreSuccessErr: true,
+		App:                 app,
+		Description:         "update: " + changeSummary,
+		Namespaces:          nsNames,
+		IgnoreSuccessErr:    true,
+		AppChangesMaxToKeep: o.DeployFlags.AppChangesMaxToKeep,
 	}
 
 	err = touch.Do(func() error {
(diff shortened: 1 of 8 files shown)


pkg/kapp/cmd/app/deploy.go
```diff
@@ -224,6 +224,7 @@ func (o *DeployOptions) Run() error {
 		Description:      "update: " + changeSummary,
 		Namespaces:       nsNames,
 		IgnoreSuccessErr: true,
+		CreateChange:     o.DeployFlags.AppChangesMaxToKeep > 0,
```
100mik: nit: I would pass the value `AppChangesmaxToKeep` down the tree and then move the condition itself to other functions. I think it might read better given the context 🤔

praveenrewar: Done.

100mik (approved):
LGTM!! 🚀

PR #814, opened 2023-09-01
Add --app-namespace flag

### LINKED ISSUE #815: Ability to set namespace for storing kapp meta configmaps
I want to set the namespace for storing the kapp metadata configmaps. Currently this is being done via the default namespace from the kubeconfig or by using the -n flag, but both of them set this for all the resources that are created by the app (unless the resources have a metadata.namespace field specified). This requires me to provide kapp with create/update and delete permissions for configmaps in my namespace. I would like to keep these separate so that I can isolate the rbac resources in a different namespace 

Add a flag --app-namespace in the kapp commands which can be used to specify the namespace to store the kapp meta configmaps.

The --app-namespace flag would take precedence over the -n flag, but if --app-namespace is not specified then kapp would behave in the same fashion as it does now, this will ensure that users who are already using kapp won't be impacted.

---
Vote on this request

praveenrewar:
Add `--app-namespace` flag to kapp deploy, inspect and delete commands. This can be used to specify the namespace to store kapp meta configmaps.

Fixes #815 

```release-note
Add `--app-namespace` flag to kapp deploy, inspect and delete commands. This can be used to specify the namespace to store kapp configmaps for managing app state.
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/app_flags.go
+++ b/pkg/kapp/cmd/app/app_flags.go
@@ -12,10 +12,12 @@ import (
 type Flags struct {
 	NamespaceFlags cmdcore.NamespaceFlags
 	Name           string
+	AppNamespace   string
 }
 
 func (s *Flags) Set(cmd *cobra.Command, flagsFactory cmdcore.FlagsFactory) {
 	s.NamespaceFlags.Set(cmd, flagsFactory)
 
 	cmd.Flags().StringVarP(&s.Name, "app", "a", s.Name, "Set app name (or label selector) (format: name, label:key=val, !key)")
+	cmd.Flags().StringVar(&s.AppNamespace, "app-namespace", s.AppNamespace, "Set app namespace (to store app state)")
 }
--- a/pkg/kapp/cmd/app/factory.go
+++ b/pkg/kapp/cmd/app/factory.go
@@ -18,8 +18,12 @@ type FactorySupportObjs struct {
 	Apps                ctlapp.Apps
 }
 
-func FactoryClients(depsFactory cmdcore.DepsFactory, nsFlags cmdcore.NamespaceFlags,
+func FactoryClients(depsFactory cmdcore.DepsFactory, nsFlags cmdcore.NamespaceFlags, appNamespace string,
 	resTypesFlags ResourceTypesFlags, logger logger.Logger) (FactorySupportObjs, error) {
+
+	if appNamespace == "" {
+		appNamespace = nsFlags.Name
+	}
 
 	coreClient, err := depsFactory.CoreClient()
 	if err != nil {
@@ -56,7 +60,7 @@ func FactoryClients(depsFactory cmdcore.DepsFactory, nsFlags cmdcore.NamespaceFl
 		CoreClient:          coreClient,
 		ResourceTypes:       resTypes,
 		IdentifiedResources: identifiedResources,
-		Apps:                ctlapp.NewApps(nsFlags.Name, coreClient, identifiedResources, logger),
+		Apps:                ctlapp.NewApps(appNamespace, coreClient, identifiedResources, logger),
 	}
 
 	return result, nil
@@ -65,7 +69,7 @@ func FactoryClients(depsFactory cmdcore.DepsFactory, nsFlags cmdcore.NamespaceFl
 func Factory(depsFactory cmdcore.DepsFactory, appFlags Flags,
 	resTypesFlags ResourceTypesFlags, logger logger.Logger) (ctlapp.App, FactorySupportObjs, error) {
 
-	supportingObjs, err := FactoryClients(depsFactory, appFlags.NamespaceFlags, resTypesFlags, logger)
+	supportingObjs, err := FactoryClients(depsFactory, appFlags.NamespaceFlags, appFlags.AppNamespace, resTypesFlags, logger)
 	if err != nil {
 		return nil, FactorySupportObjs{}, err
 	}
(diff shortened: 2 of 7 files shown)


pkg/kapp/cmd/app/app_flags.go
```diff
@@ -12,10 +12,12 @@ import (
...
 	NamespaceFlags cmdcore.NamespaceFlags
 	Name           string
+	AppNamespace   string
 }
 
 func (s *Flags) Set(cmd *cobra.Command, flagsFactory cmdcore.FlagsFactory) {
 	s.NamespaceFlags.Set(cmd, flagsFactory)
 
 	cmd.Flags().StringVarP(&s.Name, "app", "a", s.Name, "Set app name (or label selector) (format: name, label:key=val, !key)")
+	cmd.Flags().StringVar(&s.AppNamespace, "app-namespace", s.AppNamespace, "Set app namespace (to store app state)")
```
100mik, line 22: Should we indicate that this defaults to `-n` otherwise?

pkg/kapp/cmd/app/factory.go
```diff
@@ -18,7 +18,7 @@ type FactorySupportObjs struct {
 	Apps                ctlapp.Apps
 }
 
-func FactoryClients(depsFactory cmdcore.DepsFactory, nsFlags cmdcore.NamespaceFlags,
+func FactoryClients(depsFactory cmdcore.DepsFactory, appNamespace string,
```
100mik: nit: I am a bit iffy about replacing `nsFlags` with `appNamespace` here.

Primarily since we use the same factory for non-app specific commands as well. Looks like in that case we are essentially passing the value of another flag in it's stead. 

Making `app-namespace` a part of `nsFlags` might be a bit weird because of similar reasons. Maybe, we could have another argument. Maybe we should give this some thought.

pkg/kapp/cmd/appgroup/delete.go
```diff
@@ -55,7 +55,7 @@ func (o *DeleteOptions) Run() error {
 		return fmt.Errorf("Expected group name to be non-empty")
 	}
 
-	supportObjs, err := cmdapp.FactoryClients(o.depsFactory, o.AppGroupFlags.NamespaceFlags, cmdapp.ResourceTypesFlags{}, o.logger)
+	supportObjs, err := cmdapp.FactoryClients(o.depsFactory, o.AppGroupFlags.NamespaceFlags.Name, cmdapp.ResourceTypesFlags{}, o.logger)
```
100mik: Are we saying that `--app-namespace` won't apply to app groups? I am not sure if we should do that. 
Ideally, we should allow setting of a location for metadata cm for app groups as well. Do we have other flags that only apply to apps but not app groups?

test/e2e/app_namespace_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	logger.Section("deploy app with -n flag", func() {
+		kapp.RunWithOpts([]string{"deploy", "-f", "-", "-a", name, "-n", env.Namespace}, RunOpts{NoNamespace: true, StdinReader: strings.NewReader(yaml)})
+
+		// both app meta configmap and the resources should be present in the -n namespace
+		NewPresentClusterResource("configmap", name, env.Namespace, kubectl)
+		NewPresentClusterResource("configmap", "test-cm", env.Namespace, kubectl)
+	})
+
+	logger.Section("deploy same app with both -n and --app-namespace flag", func() {
```
100mik, line 44: I imagine we might have a case where we want to put resources with no ns specified to one location and have the app CM present in one.

100mik: Are we saying that migration of app CM is not allowed? Could use some clarity here :/

pkg/kapp/cmd/app/app_flags.go
praveenrewar: At first I thought of doing that too, but if we say `-n`, then that is option too, and we will have to mention everything from `-n`. I skipped it thinking that `-n` description already says that part.

pkg/kapp/cmd/app/factory.go
praveenrewar: The only reason we have the nsFlags here is to get the namespace of the app (be it any command). This could be coming from the app-namespace flag or the -n flag (line 68-69)

test/e2e/app_namespace_test.go
praveenrewar: > I imagine we might have a case where we want to put resources with no ns specified to one location and have the app CM present in one.

That would be testing the same thing, right? I mean I can create a service account and set the preferred namespace to env.Namespace, but that would be same as using -n env.Namespace.

praveenrewar: > Are we saying that migration of app CM is not allowed?

What kind of migration are we talking about? Like creating an app in namespace A and then creating in namespace B? (That would be a different app)

pkg/kapp/cmd/appgroup/delete.go
praveenrewar: Added app-namespace for app-group.

pkg/kapp/cmd/app/factory.go
```diff
@@ -42,7 +46,7 @@ func FactoryClients(depsFactory cmdcore.DepsFactory, nsFlags cmdcore.NamespaceFl
 	})
 
 	resourcesImplOpts := ctlres.ResourcesImplOpts{
-		FallbackAllowedNamespaces:        []string{nsFlags.Name},
+		FallbackAllowedNamespaces:        []string{appNamespace},
```
100mik: We should probably fallback to `nsFlags.Namespace`

100mik (approved):
Looks good!

100mik:
Looks like the tests are not happy, not sure if it is a flake 🤔

Issue #755, opened 2023-05-16
`kapp app-group deploy` should order based on folder names


Baune8D:
**What steps did you take:**
Deploy any app using `kapp app-group deploy`

**What happened:**
App deployment order is randomized on every execution.

**What did you expect:**
App deployment order should be alphabetically based on folder names.

**Anything else you would like to add:**
When reading directory using `ioutil.ReadDir` the ordering is correct, but when assigning to the result map here: https://github.com/carvel-dev/kapp/blob/df3cfc8d74bf79d52e96f8d1c1296d6f553f048b/pkg/kapp/cmd/appgroup/deploy.go#L137 ordering gets randomized.

**Environment:**

- kapp version (use `kapp --version`): 0.56.0
- OS (e.g. from `/etc/os-release`): macOS Ventura
- Kubernetes version (use `kubectl version`) 1.25.7

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

rjtch:
hey @Baune8D  i would like to work on this task

praveenrewar:
Thank you for showing interest in the fixing this @rjtch! I am assigning the issue to you, let me know if you need any help with it.

rjtch:
Hey @praveenrewar I would like to use a this library https://github.com/wk8/go-ordered-map to fix it. Is it okay or do you have any recommendations?

praveenrewar:
Hey @rjtch! I noticed that you are using a slice in your PR, which is what I would have recommended as well 🚀

PR #829, opened 2023-10-09
kapp app-group deploy should order based on folder names #755

rjtch:

Fixes #

```release-note

```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/appgroup/deploy.go
+++ b/pkg/kapp/cmd/appgroup/deploy.go
@@ -94,10 +94,6 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
-	for _, v := range existingAppsInGroup {
-		fmt.Printf("EXISTING Projects found : %s\n", v.Name())
-	}
-
 	// Delete apps that no longer are present in directories
 	for _, app := range existingAppsInGroup {
 		for _, v := range updatedApps {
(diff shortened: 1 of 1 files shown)


pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -66,14 +66,14 @@ func (o *DeployOptions) Run() error {
 
 	// TODO what if app is renamed? currently it
 	// will have conflicting resources with new-named app
-	updatedApps, err := o.appsToUpdate()
+	updatedApps, sortedApps, err := o.appsToUpdate()
```
praveenrewar: I think we can just get rid of the map here, the only place where we seem to be using it is to delete apps that are no longer present, and I think we can just use a nested loop instead of the map to check for it's existence (I don't think that anyone is going to have thousands of directories to be created as apps so we don't need to worry much about time complexity here)

rjtch: Okay @praveenrewar i will get rid of  it.

pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -94,12 +94,18 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	for _, v := range existingAppsInGroup {
+		fmt.Printf("EXISTING Projects found : %s\n", v.Name())
+	}
```
praveenrewar: remove debug print statements.

praveenrewar (commented):
Hey @rjtch Your changes looks good to me 🚀  (except the debug statement). Could you please sign your commits as well to make the DCO bot happy.
Ideally it would be good if we can add a test as well for this, but I am happy to have that in a separate PR.

PR #887, opened 2024-02-06
Add preflight check implementation, add permission validation preflight check

### LINKED ISSUE #855: Optional ability to check if client has permissions to perform CRUD actions before trying them
When deploying an app, if/when kapp encounters an error, it stops immediately. This can leave the content on the cluster in some intermediate or unknown state.

Add an optional setting for kapp to check if the client has permissions to perform every operation needed before executing any of them. If any of the checks fails, kapp makes no modifications to the cluster and returns an error.

The SubjectAccessReview API can be used for these checks.

The related kapp-controller issue is https://github.com/carvel-dev/kapp-controller/issues/1381.

---
Vote on this request

everettraven:

- Adds a flag (`--preflight`) to `kapp deploy` that can be used to enable/disable preflight checks. Can be used in the format `--preflight=CheckName=true,...`
- Adds a preflight check for permission validation that is disabled by default. Can be enabled with `--preflight=PermissionValidation=true`
    - For `(Cluster)Roles` it will ensure that a user has all the necessary permissions to create/update the resource without privilege escalation _or_ has the "escalate" permissions
    - For `(Cluster)RoleBindings` it will ensure that a user has all the necessary permissions to create/update the resource without privilege escalation _or_ has the "bind" permissions
    - For all other resources and operations, it will ensure it has the necessary permissions to perform the changes required by the change set

Fixes #855 

```release-note
Adds `--preflight` flag to the `kapp deploy` subcommand. This flag can be used to enable/disable specific preflight checks and follows the format `--preflight=CheckName=true,...`

Adds a PermissionValidation preflight check, disabled by default, that can be enabled with `--preflight=PermissionValidation=true`
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -4,6 +4,7 @@
 package app
 
 import (
+	"context"
 	"fmt"
 	"io/fs"
 	"os"
@@ -27,6 +28,7 @@ import (
 	ctldiffui "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/diffui"
 	"github.com/vmware-tanzu/carvel-kapp/pkg/kapp/logger"
 	ctllogs "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/logs"
+	"github.com/vmware-tanzu/carvel-kapp/pkg/kapp/preflight"
 	ctlres "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/resources"
 	ctlresm "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/resourcesmisc"
 )
@@ -50,11 +52,13 @@ type DeployOptions struct {
 	ResourceTypesFlags  ResourceTypesFlags
 	LabelFlags          LabelFlags
 
+	PreflightChecks *preflight.Registry
+
 	FileSystem fs.FS
 }
 
-func NewDeployOptions(ui ui.UI, depsFactory cmdcore.DepsFactory, logger logger.Logger) *DeployOptions {
-	return &DeployOptions{ui: ui, depsFactory: depsFactory, logger: logger}
+func NewDeployOptions(ui ui.UI, depsFactory cmdcore.DepsFactory, logger logger.Logger, preflights *preflight.Registry) *DeployOptions {
+	return &DeployOptions{ui: ui, depsFactory: depsFactory, logger: logger, PreflightChecks: preflights}
 }
 
 func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Command {
@@ -91,6 +95,7 @@ func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 	o.ResourceTypesFlags.Set(cmd)
 	o.LabelFlags.Set(cmd)
 	o.PrevAppFlags.Set(cmd)
+	o.PreflightChecks.AddFlags(cmd.Flags())
 
 	return cmd
 }
@@ -192,6 +197,13 @@ func (o *DeployOptions) Run() error {
 			return DeployApplyExitStatus{hasNoChanges}
 		}
 		return nil
+	}
+
+	if o.PreflightChecks != nil {
+		err = o.PreflightChecks.Run(context.Background(), clusterChangesGraph)
+		if err != nil {
+			return fmt.Errorf("preflight checks failed: %w", err)
+		}
 	}
 
 	err = o.ui.AskForConfirmation()
--- a/pkg/kapp/cmd/appgroup/deploy.go
+++ b/pkg/kapp/cmd/appgroup/deploy.go
@@ -15,6 +15,7 @@ import (
 	cmdcore "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/cmd/core"
 	cmdtools "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/cmd/tools"
 	"github.com/vmware-tanzu/carvel-kapp/pkg/kapp/logger"
+	"github.com/vmware-tanzu/carvel-kapp/pkg/kapp/preflight"
 )
 
 type DeployOptions struct {
@@ -22,9 +23,10 @@ type DeployOptions struct {
 	depsFactory cmdcore.DepsFactory
 	logger      logger.Logger
 
-	AppGroupFlags Flags
-	DeployFlags   DeployFlags
-	AppFlags      DeployAppFlags
+	AppGroupFlags   Flags
+	DeployFlags     DeployFlags
+	AppFlags        DeployAppFlags
+	PreflightChecks *preflight.Registry
 }
 
 type DeployAppFlags struct {
@@ -36,8 +38,8 @@ type DeployAppFlags struct {
 	LabelFlags          cmdapp.LabelFlags
 }
 
-func NewDeployOptions(ui ui.UI, depsFactory cmdcore.DepsFactory, logger logger.Logger) *DeployOptions {
-	return &DeployOptions{ui: ui, depsFactory: depsFactory, logger: logger}
+func NewDeployOptions(ui ui.UI, depsFactory cmdcore.DepsFactory, logger logger.Logger, preflights *preflight.Registry) *DeployOptions {
+	return &DeployOptions{ui: ui, depsFactory: depsFactory, logger: logger, PreflightChecks: preflights}
 }
 
 func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Command {
@@ -56,6 +58,7 @@ func NewDeployCmd(o *DeployOptions, flagsFactory cmdcore.FlagsFactory) *cobra.Co
 	o.AppFlags.DeleteApplyFlags.SetWithDefaults("delete", cmdapp.ApplyFlagsDeleteDefaults, cmd)
 	o.AppFlags.DeployFlags.Set(cmd)
 	o.AppFlags.LabelFlags.Set(cmd)
+	o.PreflightChecks.AddFlags(cmd.Flags())
 	return cmd
 }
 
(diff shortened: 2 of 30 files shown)


pkg/kapp/cmd/core/deps_factory.go
```diff
@@ -83,6 +87,29 @@ func (f *DepsFactoryImpl) CoreClient() (kubernetes.Interface, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	groupResources, err := restmapper.GetAPIGroupResources(disc)
+	if err != nil {
+		return nil, err
+	}
+
+	mapper := restmapper.NewDiscoveryRESTMapper(groupResources)
```
stevekuznetsov: Generally you want to use the deferred mapper so you don't send 300 requests to the server on startup.

pkg/kapp/permissions/validator.go
```diff
@@ -0,0 +1,508 @@
...
+			// rules such that the subrules contain
+			// at most one verb, one group, and one resource
+			// source at: https://github.com/kubernetes/component-helpers/blob/9a5801419916272fc9cec7a7822ed525721b99d3/auth/rbac/validation/policy_comparator.go#L56-L84
+			var subrules []rbacv1.PolicyRule = validation.BreakdownRule(rule)
+			for _, subrule := range subrules {
+				ssar := &authv1.SelfSubjectAccessReview{
+					Spec: authv1.SelfSubjectAccessReviewSpec{
+						ResourceAttributes: &authv1.ResourceAttributes{
+							Namespace: res.Namespace(),
+							Verb:      subrule.Verbs[0],
```
stevekuznetsov: I don't think there's any way to handle the set of permissions in a Role without having loops for each of these dimensions - it's totally valid and reasonable to specify more than one verb, group, or resource (or all of them!) in a rule, as well as non-resource URLs, etc. That's why the OLMv0 bits we changed recently compute the hash of the rules and expect the user to have pre-created RBAC with the tool, so you can do a simple O(1) lookup.

joelanford: I'd propose that we use `SelfSubjectRulesReview` here instead. I'm working on a kubectl plugin that might help or inspire: https://github.com/joelanford/kubectl-make_rbac

everettraven: >I don't think there's any way to handle the set of permissions in a Role without having loops for each of these dimensions - it's totally valid and reasonable to specify more than one verb, group, or resource (or all of them!) in a rule, as well as non-resource URLs, etc.

Yep, due to the limitation of the `SelfSubjectAccessReview` requiring a single value reference for verb, group, and resource I used a helper that essentially iterates all these values and returns a slice of all the policies broken into a policy where only a single group, resource, and verb value are provided. 


> I'd propose that we use `SelfSubjectRulesReview` here instead. I'm working on a kubectl plugin that might help or inspire: https://github.com/joelanford/kubectl-make_rbac

This is also a direction I'm open to looking into. I've recently made some changes locally that will allow for an easier transition to whatever underlying implementation we feel is better suited for the use case here. Since this is still an early draft and a WIP, I'm going to focus on finishing up the plumbing before making any changes in the "how" the permissions are validated.

pkg/kapp/cmd/core/deps_factory.go
everettraven: Deferred mapper used in https://github.com/carvel-dev/kapp/pull/887/commits/feea616b47b08fef28ad4f0fbd565c600bb37575

pkg/kapp/cmd/appgroup/deploy.go
```diff
@@ -146,7 +146,8 @@ func (o *DeployOptions) deployApp(app appGroupApp) error {
 	o.ui.PrintLinef("--- deploying app '%s' (namespace: %s) from %s",
 		app.Name, o.appNamespace(), app.Path)
 
-	deployOpts := cmdapp.NewDeployOptions(o.ui, o.depsFactory, o.logger)
+	// TODO: is this a location that needs proper preflight checks?
```
praveenrewar: We do try to keep `kapp app-group deploy` and `kapp deploy` options in sync, so we should try to add the preflight checks here as well.

pkg/kapp/preflight/registry.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	return nil
+}
+
+func (c *Registry) AddFlags(flags *pflag.FlagSet) {
+	knownChecks := []string{}
+	for name, _ := range c.known {
+		knownChecks = append(knownChecks, name)
+	}
+	flags.Var(c, preflightFlag, fmt.Sprintf("preflight checks to run. Available preflight checks are %s", strings.Join(knownChecks, ",")))
```
praveenrewar: I am wondering if having a map of all the preflight checks might become a little bit confusing in the help section 🤔
(I am a little bit inclined towards `--validate-options` as that seems more intuitive)

pkg/kapp/permissions/preflight.go
```diff
@@ -0,0 +1,78 @@
...
+			if err != nil {
+				return err
+			}
+
+			err = validator.Validate(context.Background(), change.Change.Resource(), "update")
+			if err != nil {
+				return err
+			}
+
+			// TODO: should check patch permissions?
```
praveenrewar: As of now we don't use patch, but we do plan on using patch to add the last applied annotation, until then I think it's fine to skip patch.

pkg/kapp/cmd/app/deploy.go
```diff
@@ -172,6 +177,13 @@ func (o *DeployOptions) Run() error {
 		return err
 	}
 
+	if o.PreflightChecks != nil {
```
praveenrewar: We can probably move this after we ask for a confirmation to deploy (`o.ui.AskForConfirmation`).

praveenrewar (commented):
Looks pretty good, thank you so much for putting in the effort @everettraven ❤️

I do want to double check my understanding on the different validations. We are trying to ensure that:
- User has permissions to create the resources that they want to create.
- For rbac resources, they should only be able to add the permissions that they currently have? 

I didn't think of the second use case before, really glad that you have covered this as well 🙇🏻

pkg/kapp/cmd/app/deploy.go
100mik: Would we not need a permissions check before we diff? Or do we want to consider permissions needed for the apply?

praveenrewar: Before apply, we just need `get` permission, if we don't have that, we will fail very early.

100mik: I think it would be helpful to do it earlier, especially in case someone is trying put together some roles. Having this as a way of debugging roles seems to be pretty cool !

everettraven: Ah, moved it after confirmation before seeing this. I'll move it back :)

pkg/kapp/permissions/validator.go
everettraven: Based on the discussion from the community meeting, the maintainers would prefer to get this in with SSAR and later investigate updating to use SSRR

pkg/kapp/preflight/registry.go
everettraven: Maybe? This tries to keep it cleaner by only doing a slice of the check names in the usage method. If there is concern with include the list of defaults I can remove it. I personally err on the side of more information rather than less though. IMO it should be self documenting and the current implementation should adhere to that in as succinct a way as possible.

pkg/kapp/cmd/appgroup/deploy.go
everettraven: Should be updated now

pkg/kapp/preflight/registry.go
praveenrewar: My main concern was that the help section for preflight flag would grow to be very long as we add more preflight checks, but it looks fine as of now, so we can think about it later.

pkg/kapp/cmd/app/deploy.go
praveenrewar: My main reason for having it down was to ensure that the `diff` flags work even if we don't have enough permissions. For example, if I am using `--diff-run` along with `--preflight` I think that kapp should exit gracefully after showing the diff (even if we don't have sufficient permissions to make the changes) because we are _not_ making any change.

everettraven: Ah, that makes sense. Thanks @praveenrewar for the extra context! Does it make sense to move this after the diff is shown but before the confirmation step? I'm not really opinionated here because I'm not as well versed in the user flow here as you folks so I'm happy to move it wherever makes the most sense.

everettraven: Fixed linter failures and moved the preflight checks execution to right before the confirmation prompt in the latest push

praveenrewar: > Does it make sense to move this after the diff is shown but before the confirmation step?

Hmm, I was thinking that a user could still say "no" during the confirmation, but it wouldn’t really be a _preflight_ check if we do it after confirmation, so I think it makes sense.

test/e2e/preflight_permission_validation_escalation_test.go
```diff
@@ -0,0 +1,142 @@
...
+		RemoveClusterResource(t, "ns", testName, "", kubectl)
+	}
+	cleanUp()
+	defer cleanUp()
+
+	kapp.RunWithOpts([]string{"deploy", "-a", baseName, "-f", "-"}, RunOpts{StdinReader: strings.NewReader(base)})
+	cleanUpContext := ScopedContext(t, kubectl, testName, scopedContext, scopedUser)
+	defer cleanUpContext()
+
+	// Verify that permissions are missing for applying a (Cluster)Role
```
praveenrewar: Could you update these comments. They are saying "missing" for all the tests cases.

praveenrewar: Should we also add a failing test case for escalation?

everettraven: Ah great catch on the comments. I meant to go back and update them. Must've gotten distracted and forgot 😁.

I'll go back and update the comments and add an e2e test specifically for failing privilege escalation

everettraven: In the latest push I:
- Removed the comments from all the tests as they aren't really necessary and I had originally used them as a check list for what I needed to implement
- Added an e2e test for failing privilege escalation

everettraven:
@praveenrewar I've updated the e2e tests to fix the CI failure. Checked that it passed locally with a fresh cluster this time around.

praveenrewar (approved):
LGTM! Thank you so much for working on this!!
@100mik I am keeping it open in case you are still taking a look.

100mik (approved):
LGTM! Thanks for the contribution ❤️ 

 @everettraven we will need some documentation changes to carvel-dev/carvel around the `--preflight` flag. I do not think it should block this PR! However, we should make sure we have these in place before we push out a release.

100mik:
I created a separate issue to track the required documentation changes 🙌🏼

PR #894, opened 2024-03-04
Update --preflight command line option

### LINKED ISSUE #893: Update `--preflight` flag to no longer follow the `Check=true` format
In some discussions around https://github.com/carvel-dev/carvel/pull/729 it's come to my attention that with all preflight checks being disable by default, it doesn't make much sense to require the enabling of preflight checks to follow the format of `--preflight=Check=true,OtherCheck=true,...` and could instead follow the format of `--preflight=Check,OtherCheck,...`. This makes it follow an "enabler" pattern where all preflight checks must be explicitly enabled

tmshort:
Fix #893

The `--preflight` command line options now only specifies those preflight checks that are to be enables.

This simplifies the command line `--preflight` option; we're looking to have additional configuration located in a configuration file, rather than on the command line.

Fixes #893 

```release-note
Update --preflight flag to be a simple list of preflight validators to enable.
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/preflight/registry.go
+++ b/pkg/kapp/preflight/registry.go
@@ -6,7 +6,6 @@ package preflight
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"strings"
 
 	"github.com/spf13/pflag"
@@ -31,16 +30,18 @@ func NewRegistry(checks map[string]Check) *Registry {
 }
 
 // String returns a string representation of the
-// preflight checks. It follows the format:
-// CheckName={true||false},...
+// enabled preflight checks. It follows the format:
+// CheckName,...
 // This method is needed so Registry implements
 // the pflag.Value interface
 func (c *Registry) String() string {
-	defaults := []string{}
+	enabled := []string{}
 	for k, v := range c.known {
-		defaults = append(defaults, fmt.Sprintf("%s=%v", k, v.Enabled()))
+		if v.Enabled() {
+			enabled = append(enabled, k)
+		}
 	}
-	return strings.Join(defaults, ",")
+	return strings.Join(enabled, ",")
 }
 
 // Type returns a string representing the type
@@ -51,9 +52,10 @@ func (c *Registry) Type() string {
 }
 
 // Set takes in a string in the format of
-// CheckName={true||false},...
+// CheckName,...
 // and sets the specified preflight check
-// as enabled if true, disabled if false
+// as enabled if listed, otherwise, sets as
+// disabled if not listed.
 // Returns an error if there is a problem
 // parsing the preflight checks
 func (c *Registry) Set(s string) error {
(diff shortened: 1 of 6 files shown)


pkg/kapp/preflight/registry.go
```diff
@@ -51,33 +52,32 @@ func (c *Registry) Type() string {
...
+// as enabled if listed, otherwise, sets as
+// disabled if not listed.
 // Returns an error if there is a problem
 // parsing the preflight checks
 func (c *Registry) Set(s string) error {
 	if c.known == nil {
 		return nil
 	}
 
+	enabled := map[string]struct{}{}
```
everettraven, line 66: Non blocking, but using the [sets package](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets) would make it so this can be a new `sets.Set` instead of doing this `map[string]struct{}{}` setup. Then you can do things like `sets.Insert(enabled, "key")` and `enabled.Has("key")`

everettraven (commented):
Other than a non-blocking nit this looks good to me. Probably will need to also update the e2e tests

everettraven:
More explicitly I suspect these 4 e2e tests will need to be updated:
- https://github.com/carvel-dev/kapp/blob/develop/test/e2e/preflight_permission_validation_escalation_test.go
- https://github.com/carvel-dev/kapp/blob/develop/test/e2e/preflight_permission_validation_failed_escalation_test.go
- https://github.com/carvel-dev/kapp/blob/develop/test/e2e/preflight_permission_validation_missing_test.go
- https://github.com/carvel-dev/kapp/blob/develop/test/e2e/preflight_permission_validation_test.go

tmshort:
> Other than a non-blocking nit this looks good to me. Probably will need to also update the e2e tests

I was looking at your original PR, and somehow missed you had updated e2e... I thought my local e2e failures were due to my environment... will update.

pkg/kapp/preflight/registry.go
tmshort: In this case it was a temporary storage, avoid disabling everything, then enabling only things we cared about. We don't actually care about the the value, just which one's have been enabled.
I suppose, if a `Set` were used, we could set the Set to disabled, then run through the passed in commands, and go through the Set and enable/disable...

There are so many ways to do this!

everettraven: A `sets.Set` is used exactly for the purpose of only caring about a key without the value. It's a generic map to an empty struct:
>type Set[T [comparable](https://pkg.go.dev/builtin#comparable)] map[T][Empty](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Empty)

Again, non-blocking because like you said there are many ways to do this but this is a nice helper for this that is in the kube apimachinery package :)

tmshort:
Doc PR: https://github.com/carvel-dev/carvel/pull/733

everettraven (approved):

pkg/kapp/preflight/registry.go
100mik: **[thought]** Would have been our first use of generics 😆 , but nothing to add on here. Sounds good to me either way!

100mik (approved):
The changes look good to me!

I am a bit worried that we might end up in a weird spot if we _do_ want some of these to be enabled by default in the future. But it is something we can revisit when that is the case 🙌🏼

100mik:
@everettraven I am curious about how the flag would affect the proposed way of configuring preflights [here](https://github.com/carvel-dev/carvel/pull/729/files#diff-4a58b74b898751ee41d36dd947d166e4e1e47bbb62869512fb192f269857cb7aR52), but we can discuss that further in the proposal itself

tmshort:
> @everettraven I am curious about how the flag would affect the proposed way of configuring preflights [here](https://github.com/carvel-dev/carvel/pull/729/files#diff-4a58b74b898751ee41d36dd947d166e4e1e47bbb62869512fb192f269857cb7aR52), but we can discuss that further in the proposal itself

We decided to move preflight configuration into the configuration file (rather than be part of the command line flags). That PR should be updated now to reflect that.

tmshort:
ping for approval?

rashmigottipati (approved):
overall, the changes look good to me!

100mik:
@praveenrewar merge away if this looks good to you, I am good with these changes. Let's explore where we are headed further in the proposal 🙌🏼

praveenrewar (approved):
LGTM! Thank you so much!

PR #900, opened 2024-03-12
Add support for preflight check configuration

### LINKED ISSUE #899: Add configuration support to kapp for preflight checks
[A description of the current challenge that you are experiencing.]

Preflight checks have no method of configuration. This adds a generic configuration for preflight checks.

[A clear and concise description of what you want to happen. If applicable a visual representation of the UX (ex: new CLI argument name, the behavior expected).]

This should implement https://github.com/carvel-dev/carvel/pull/729

---
Vote on this request

tmshort:

This adds configuration file support (`preflightRules`) for preflight checks. Each preflight check must add a `SetConfig()` method to accept configuration from the Preflight registry.

Fixes #899

```release-note
Add support for preflightRules configuration
```

##### Review Checklist:

  a link to that PR
  change

```docs
```
Proposal https://github.com/carvel-dev/carvel/pull/729

### DIFF
--- a/pkg/kapp/preflight/registry.go
+++ b/pkg/kapp/preflight/registry.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/spf13/pflag"
+	"github.com/vmware-tanzu/carvel-kapp/pkg/kapp/config"
 	ctldgraph "github.com/vmware-tanzu/carvel-kapp/pkg/kapp/diffgraph"
 )
 
@@ -17,6 +18,8 @@ const preflightFlag = "preflight"
 // Registry is a collection of preflight checks
 type Registry struct {
 	known map[string]Check
+	// Stores the enabled values from the command line
+	enabledFlag map[string]bool
 }
 
 // NewRegistry will return a new *Registry with the
@@ -59,25 +62,23 @@ func (c *Registry) Type() string {
 // Returns an error if there is a problem
 // parsing the preflight checks
 func (c *Registry) Set(s string) error {
-	if c.known == nil {
+	if c.known == nil || c.enabledFlag == nil {
 		return nil
 	}
 
-	enabled := map[string]struct{}{}
-	// enable those specified
+	// Using enabledFlag allows multiple --preflight check flags to be specified
 	mappings := strings.Split(s, ",")
 	for _, key := range mappings {
 		if _, ok := c.known[key]; !ok {
 			return fmt.Errorf("unknown preflight check %q specified", key)
 		}
-		c.known[key].SetEnabled(true)
-		enabled[key] = struct{}{}
+		c.enabledFlag[key] = true
 	}
-	// disable unspecified validators
+
+	// enable/disabled based on validators specified
 	for key := range c.known {
-		if _, ok := enabled[key]; !ok {
-			c.known[key].SetEnabled(false)
-		}
+		enabled, ok := c.enabledFlag[key]
+		c.known[key].SetEnabled(ok && enabled)
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


pkg/kapp/preflight/registry.go
```diff
@@ -101,9 +102,56 @@ func (c *Registry) AddCheck(name string, check Check) {
...
+		}
+		if _, ok := haveConfig[rule.Name]; ok {
+			return fmt.Errorf("duplicate preflight check in configuration: %q", rule.Name)
+		}
+		haveConfig[rule.Name] = true
+	}
+	return nil
+}
+
+func (c *Registry) SetConfig(conf config.Conf) error {
```
everettraven: @tmshort I know we have already had this conversation in the PR against my fork before the CRD proposal was approved, but I wanted to call it out here and get the opinions of the carvel maintainers as well.

IMO we should change this to not accept the full `config.Conf` type. Instead I think something like:
```suggestion
func (c *Registry) SetRules(rules ...config.PreflightRule) error {
```
would be better here for two reasons:
1. This method really only cares about the preflight rules anyways. Because of this it is easier and more clear to a user what information is really used and taken into account when using this method to configure a preflight registry. Keeping the parameters to the minimum thing needed is a better UX IMO.
2. Eliminates the parsing of YAML in the unit tests, reducing the, IMO, unrelated points of failure in the unit tests. I feel that changes to logic that would break the YAML parsing should break tests specifically associated with parsing of the YAML contents and not the tests for a function that assumes the parsing has already been done and takes in the type resulting from the YAML parsing.

I'm by no means saying this can't go in like this, I'm totally fine with that if the consensus is that what is currently implemented is the best approach. Just wanted to highlight what I think might make things a bit easier to understand and reason about as both a consumer of the library and a maintainer.

tmshort: _If_ this were the desired route, I'd prefer `[]PrelightRule` as the parameter, as that's the output of `config.PreflightRules()`. The above approach requires an unnecessary break-apart step, vs:
```
registry.SetConfig(conf.PreflightRules())
...
func (c *Registry) SetConfig(rules []PreflightRules) error {
...
```
I prefer the YAML config in the unit tests, as I think it's easier to read than golang structs.

everettraven: > I'd prefer []PrelightRule as the parameter, as that's the output of config.PreflightRules(). The above approach requires an unnecessary break-apart step, vs

Fair point, totally fine with me. I was just throwing something together as an example.

>I prefer the YAML config in the unit tests, as I think it's easier to read than golang structs.

Also fair, but I personally feel its easier to read the golang structs than the yaml in this case. To each their own :). This also isn't explicitly about removing the YAML config, I'm fine with that approach if the consensus is that the currently implemented approach. I personally think the removal of the YAML config is one benefit, but I think the other benefit I mentioned stands on its own.

tmshort: Yes, one could be done with (or without) the other.

praveenrewar: Personally, I prefer the YAML config in the tests as well, as it's easier to scan through them, however, I agree with @everettraven that we need not pass the whole config and just the preflight rules should be enough.

praveenrewar (commented):
Looks good to me apart from Bryce's comment. 
Should we add/update the preflight permissions e2e tests with a config example? (I am okay with just the unit tests as well)

Would would also have to add the preflight option in the [configuration](https://carvel.dev/kapp/docs/develop/config/) docs.

pkg/kapp/preflight/registry.go
tmshort: I can tweak the API to only pass the preflight rules.

tmshort:
> Looks good to me apart from Bryce's comment. Should we add/update the preflight permissions e2e tests with a config example? (I am okay with just the unit tests as well)
> 
> Would would also have to add the preflight option in the [configuration](https://carvel.dev/kapp/docs/develop/config/) docs.

I'll create a new PR for the docs.

The unit tests cover the code as written. IMHO, an e2e would be more appropriate once a preflight check adds configuration settings; otherwise, the e2e would effectively be a duplicate test.

tmshort:
Documentation PR: https://github.com/carvel-dev/carvel/pull/740

everettraven (approved):

praveenrewar (approved):
LGTM! Thank you so much for the contribution ❤️

PR #903, opened 2024-03-13
Update copyright header template to be entity agnostic, update existing headers

### LINKED ISSUE #902: Update copyright headers
Currently the copyright headers for all the tools say:
```
// Copyright 2023 VMware, Inc.
// SPDX-License-Identifier: Apache-2.0
```

Now that Carvel is a CNCF project, it should be something like
```
// Copyright 2023 The Carvel Authors.
// SPDX-License-Identifier: Apache-2.0
```

---
Vote on this request

everettraven:
- Updates the required code copyright header template to be entity agnostic
- Updates all the existing headers to use the new template and updates copyright year to 2024

Fixes #902 

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/cmd/kapp/kapp.go
+++ b/cmd/kapp/kapp.go
@@ -1,4 +1,4 @@
-// Copyright 2020 VMware, Inc.
+// Copyright 2024 The Carvel Authors.
 // SPDX-License-Identifier: Apache-2.0
 
 package main
--- a/pkg/kapp/config/default_test.go
+++ b/pkg/kapp/config/default_test.go
@@ -1,4 +1,4 @@
-// Copyright 2022 VMware, Inc.
+// Copyright 2024 The Carvel Authors.
 // SPDX-License-Identifier: Apache-2.0
 
 package config_test
--- a/code-header-template.txt
+++ b/code-header-template.txt
@@ -1,2 +1,2 @@
-Copyright {{copyright-year}} VMware, Inc.
+Copyright {{copyright-year}} The Carvel Authors.
 SPDX-License-Identifier: Apache-2.0
--- a/pkg/kapp/app/app_filter.go
+++ b/pkg/kapp/app/app_filter.go
@@ -1,4 +1,4 @@
-// Copyright 2020 VMware, Inc.
+// Copyright 2024 The Carvel Authors.
 // SPDX-License-Identifier: Apache-2.0
 
 package app
--- a/pkg/kapp/app/app_meta.go
+++ b/pkg/kapp/app/app_meta.go
@@ -1,4 +1,4 @@
-// Copyright 2020 VMware, Inc.
+// Copyright 2024 The Carvel Authors.
 // SPDX-License-Identifier: Apache-2.0
 
 package app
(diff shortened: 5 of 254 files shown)


cmd/kapp/kapp.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 VMware, Inc.
+// Copyright 2024 The Carvel Authors.
```
everettraven, line 1: I updated all the copyright years to 2024 as I figured that would be required. If folks feel that is unnecessary or is not best practice/allowed I'll revert the copyright years and only change from `VMWare, Inc` --> `The Carvel Authors`

kumaritanushree: May I know why do you feel it will be required to change year? As per my understanding we add the year as evidence of when the file has been written.

everettraven: I've seen this done in quite a few open source projects in the past, but I am by no means an expert in copyright which is why if there is someone more familiar with this that says this isn't necessary I am happy to revert it.

As far as I am aware though, copyright should be updated whenever changes are "published" to a file to make sure the changes are appropriately captured in the copyright. This is totally subjective, but since this is an open source project a commit being merged is, I believe, the same as that change being "published". I think in general it is easier to just do a wider sweep and update all the copyrights to the most recent year. From a bit of surface level research there are various valid formats for copyright updates. A brief set of what I found is valid:
- `Copyright 2020, 2022, 2024 ...`
- `Copyright 2020 - 2024 ...`
- `Copyright 2024 ...`

Just to re-iterate, I am not an expert and going based on prior experience and a very rudimentary understanding of copyright when it comes to open source projects. I am happy to revert the changes to the year if someone with more knowledge/experience says that the changes are unnecessary.

praveenrewar: I think it makes sense to switch to the 2024. @microwavables, @joaopapereira Do you have any thoughts/concerns with it?

pkg/kapp/config/default_test.go
```diff
@@ -105,14 +106,14 @@ spec:
   0,  3 +   annotations:
   0,  4 +     kapp.k14s.io/versioned: ""
   0,  5 +   name: test-configmap-ver-1
-  0,  6 + 
+  0,  6 +
```
praveenrewar: I think this is causing the tests to fail. Was this change intended?

cmd/kapp/kapp.go
joaopapereira: I am ok with the move to 2024

pkg/kapp/config/default_test.go
everettraven: Ah, not sure why that changed - probably an autoformatting thing with the editor I was using at the time :( - Will try to get that fixed in the next couple of days.

everettraven: This should be fixed with the latest push - thanks for catching! 

Lesson learned to not use my editor to do a find + replace and instead use `sed` or equivalent 🤦‍♂️

praveenrewar (approved):
LGTM! Thank you for the change 🚀

PR #921, opened 2024-04-09
Add handling of updates to a field's enum values to the CRD Upgrade Safety preflight check

### LINKED ISSUE #916: Add validation to `CRDUpgradeSafety` preflight check to allow adding more enum values to existing field
Now that there is a base `CRDUpgradeSafety` preflight check in place, we can continue adding validation logic based on the [`CRDUpgradeSafety` preflight check proposal](https://github.com/carvel-dev/carvel/blob/develop/proposals/kapp/001-crd-upgrade-safety/README.md).

This issue focuses on adding a validation to ensure that updates to an existing field are allowed when the update performed is adding more allowed enum values

As a potential source of inspiration, here is how a couple of the existing validations are implemented:
- Definitions: https://github.com/carvel-dev/kapp/blob/82f2a30f59c60b85039a921c39ac9df23f1cb3af/pkg/kapp/crdupgradesafety/validator.go#L71-L93
- Consumption: https://github.com/carvel-dev/kapp/blob/82f2a30f59c60b85039a921c39ac9df23f1cb3af/pkg/kapp/crdupgradesafety/preflight.go#L36-L39

>[!NOTE]
>It may be useful to consider a more generic check that allows for easily extending the validations that are run against updates to existing fields in a CRD schema

everettraven:
- Adds a generic, extensible validator for handling changes to existing fields in an existing version of a CRD schema
- Adds an initial field change validator for handling changes to an existing fields allowed enum values
    - Rejects when the existing version of the field does not have enum restrictions but the new version adds enum restriction
    - Rejects when enums from the existing version of the field are removed in the new version
    - Approves when enum values are added to the list of allowed enum values

Fixes #916 

```release-note
Add logic to the CRD Upgrade Safety pre-flight check to only allow additions to the valid enum values of an existing field in an existing version of a CRD schema
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/crdupgradesafety/change_validator.go
+++ b/pkg/kapp/crdupgradesafety/change_validator.go
@@ -1,0 +1,216 @@
+// Copyright 2024 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package crdupgradesafety
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+
+	"github.com/openshift/crd-schema-checker/pkg/manifestcomparators"
+	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ChangeValidation is a function that accepts a FieldDiff
+// as a parameter and should return:
+// - a boolean representation of whether or not the change
+// - an error if the change would be unsafe
+// has been fully handled (i.e no additional changes exist)
+type ChangeValidation func(diff FieldDiff) (bool, error)
+
+// EnumChangeValidation ensures that:
+// - No enums are added to a field that did not previously have
+// enum restrictions
+// - No enums are removed from a field
+// This function returns:
+// - A boolean representation of whether or not the change
+// has been fully handled (i.e the only change was to enum values)
+// - An error if either of the above validations are not satisfied
+func EnumChangeValidation(diff FieldDiff) (bool, error) {
+	// This function resets the enum values for the
+	// old and new field and compares them to determine
+	// if there are any additional changes that should be
+	// handled. Reseting the enum values allows for chained
+	// evaluations to check if they have handled all the changes
+	// without having to account for fields other than the ones
+	// they are designed to handle. This function should only be called when
+	// returning from this function to prevent unnecessary overwrites of
+	// these fields.
+	handled := func() bool {
+		diff.Old.Enum = []v1.JSON{}
+		diff.New.Enum = []v1.JSON{}
+		return reflect.DeepEqual(diff.Old, diff.New)
+	}
+
+	if len(diff.Old.Enum) == 0 && len(diff.New.Enum) > 0 {
+		return handled(), fmt.Errorf("enums added when there were no enum restrictions previously")
+	}
+
+	oldSet := sets.NewString()
+	for _, enum := range diff.Old.Enum {
+		if !oldSet.Has(string(enum.Raw)) {
+			oldSet.Insert(string(enum.Raw))
+		}
+	}
+
+	newSet := sets.NewString()
+	for _, enum := range diff.New.Enum {
+		if !newSet.Has(string(enum.Raw)) {
+			newSet.Insert(string(enum.Raw))
+		}
+	}
+
+	diffSet := oldSet.Difference(newSet)
+	if diffSet.Len() > 0 {
+		return handled(), fmt.Errorf("enum values removed: %+v", diffSet.UnsortedList())
+	}
+
+	return handled(), nil
+}
+
+// ChangeValidator is a Validation implementation focused on
+// handling updates to existing fields in a CRD
+type ChangeValidator struct {
+	// Validations is a slice of ChangeValidations
+	// to run against each changed field
+	Validations []ChangeValidation
+}
+
+func (cv *ChangeValidator) Name() string {
+	return "ChangeValidator"
+}
+
+// Validate will compare each version in the provided existing and new CRDs.
+// Since the ChangeValidator is tailored to handling updates to existing fields in
+// each version of a CRD. As such the following is assumed:
+// - Validating the removal of versions during an update is handled outside of this
+// validator. If a version in the existing version of the CRD does not exist in the new
+// version that version of the CRD is skipped in this validator.
+// - Removal of existing fields is unsafe. Regardless of whether or not this is handled
+// by a validator outside this one, if a field is present in a version provided by the existing CRD
+// but not present in the same version provided by the new CRD this validation will fail.
+//
+// Additionally, any changes that are not validated and handled by the known ChangeValidations
+// are deemed as unsafe and returns an error.
+func (cv *ChangeValidator) Validate(old, new v1.CustomResourceDefinition) error {
+	errs := []error{}
+	for _, version := range old.Spec.Versions {
+		newVersion := manifestcomparators.GetVersionByName(&new, version.Name)
+		if newVersion == nil {
+			// if the new version doesn't exist skip this version
+			continue
+		}
+		flatOld := FlattenSchema(version.Schema.OpenAPIV3Schema)
+		flatNew := FlattenSchema(newVersion.Schema.OpenAPIV3Schema)
+
+		diffs, err := CalculateFlatSchemaDiff(flatOld, flatNew)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("calculating schema diff for CRD version %q", version.Name))
+			continue
+		}
+
+		for field, diff := range diffs {
+			handled := false
+			for _, validation := range cv.Validations {
+				ok, err := validation(diff)
+				if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/kapp/crdupgradesafety/change_validator.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func (cv *ChangeValidator) Validate(old, new v1.CustomResourceDefinition) error {
+	errs := []error{}
+	for _, version := range old.Spec.Versions {
+		newVersion := manifestcomparators.GetVersionByName(&new, version.Name)
+		if newVersion == nil {
+			// if the new version doesn't exist skip this version
+			continue
+		}
```
everettraven, line 105: Thoughts on whether or not this should be an error instead? 

Ideally this is already handled by a previous validation since the ChangeValidator is focused specifically on handling _only_ changes to individual fields, not entire versions of the CRD. Thinking through this has also reminded me I should improve the godoc around the ChangeValidator type and its expectations. Will add more godoc comments over the next couple days or so

everettraven: godoc comments were added to make it a bit easier when using as a library to understand what each thing does.  Thinking about this a bit more I think this logic to skip if an existing version is not found in the new definition is OK because I don't _think_ it is considered a breaking change to remove a non-stored version (please let me know if I am wrong!) and we already have an existing validation that ensures no stored version is removed

everettraven:
Note: I'll squash commits after reviews

pkg/kapp/crdupgradesafety/change_validator.go
```diff
@@ -0,0 +1,214 @@
...
+// has been fully handled (i.e no additional changes exist)
+type ChangeValidation func(diff FieldDiff) (bool, error)
+
+// EnumChangeValidation ensures that:
+// - No enums are added to a field that did not previously have
+// enum restrictions
+// - No enums are removed from a field
+// If either of the above scenarios are not satisfied an error
+// representing the failure will be returned.
+// Additionally, a boolean representation of whether or not the change
```
rashmigottipati: Nit: It looks like the comment has been cut off abruptly.

everettraven: Good catch! I must've gotten distracted and forgot to finish it up 😅

everettraven: Should be fixed in https://github.com/carvel-dev/kapp/pull/921/commits/b192141d574799e7af39e98e4edd58dcf536cd93

pkg/kapp/crdupgradesafety/change_validator.go
```diff
@@ -0,0 +1,215 @@
+// Copyright 2024 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
+package crdupgradesafety
```
praveenrewar, line 4: nit: Add newline after copyright headers.

pkg/kapp/crdupgradesafety/change_validator.go
```diff
@@ -0,0 +1,215 @@
...
+
+// EnumChangeValidation ensures that:
+// - No enums are added to a field that did not previously have
+// enum restrictions
+// - No enums are removed from a field
+// If either of the above scenarios are not satisfied an error
+// representing the failure will be returned.
+// Additionally, a boolean representation of whether or not the change
+// is fully handled (i.e nothing but enum values were changed) is returned
+func EnumChangeValidation(diff FieldDiff) (bool, error) {
```
praveenrewar, line 32: nit: We are returning a `bool` and an `error`, however the doc above mentions `error` and `bool` (the order is reversed). I am not sure if there is any convention regarding the docs mentioning the returned params in the same order so not forcing to change this and leaving it upto you.

pkg/kapp/crdupgradesafety/change_validator.go
praveenrewar: > I don't think it is considered a breaking change to remove a non-stored version (please let me know if I am wrong!)

This is my understanding too.

+1 to skipping such versions.

pkg/kapp/crdupgradesafety/change_validator.go
```diff
@@ -0,0 +1,215 @@
...
+				if err != nil {
+					errs = append(errs, fmt.Errorf("version %q, field %q: %w", version.Name, field, err))
+				}
+				if ok {
+					handled = true
+					break
+				}
+			}
+
+			if !handled {
```
praveenrewar, line 128: I am not able to directly relate to the user persona here so I could be wrong, but maybe this should be configurable via the kapp config somehow? (Definitely not as part of this PR)

pkg/kapp/crdupgradesafety/change_validator_test.go
```diff
@@ -0,0 +1,429 @@
+// Copyright 2024 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
+package crdupgradesafety
```
praveenrewar: nit: newline after copyright headers

praveenrewar: Also, test package name should be `crdupgradesafety_test`. I think I missed this in previous PRs (apologies 😅).

test/e2e/preflight_crdupgradesafety_invalid_field_change_enums_added_test.go
```diff
@@ -0,0 +1,122 @@
...
+	appName := "preflight-crdupgradesafety-app"
+
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", appName})
+		RemoveClusterResource(t, "ns", testName, "", kubectl)
+	}
+	cleanUp()
+	defer cleanUp()
+
+	kapp.RunWithOpts([]string{"deploy", "-a", appName, "-f", "-"}, RunOpts{StdinReader: strings.NewReader(base)})
```
praveenrewar: nit: We can keep the base deploy closer to the update (inside logger)

praveenrewar (commented):
Looks pretty good apart from some minor comments.

I also wanted to mention one additional thing apart from the comments. In the Carvel world, we have been capitalising error strings (examples can be found across the projects). I know, I know, this is not as per the go style decisions, however since this is being done across the carvel tool suite, the error messages are consistent within Carvel. This might be something worth discussing outside the scope of this PR (maybe in the community meeting?). At this point I don't want to force you to capitalise error strings as even now when we get errors from some other library, they are not capitalised and the error messages are a mix of the two, which would be something similar with preflight checks.

pkg/kapp/crdupgradesafety/change_validator.go
everettraven: Yep! I believe this is something called out in the proposal that should _eventually_ be configurable by the kapp config as a "failure mode" of "open" (i.e if not handled don't error) or "closed" (i.e if not handled then error) that defaults to "closed". 

I'm happy to include the configuration in this PR but figured if the default is going to be "closed" then we can add the configuration options in a separate, smaller PR.

pkg/kapp/crdupgradesafety/change_validator.go
everettraven: I'm not sure if there are conventions regarding return value ordering but it was originally (error, bool) as the return ordering but the linter wasn't happy about that :). I'll update the comment to be consistent

test/e2e/preflight_crdupgradesafety_invalid_field_change_enums_added_test.go
everettraven: Should be done in all the newly added tests in https://github.com/carvel-dev/kapp/pull/921/commits/ce52021254d7ca39becbd0dd4711c11680b1db55

pkg/kapp/crdupgradesafety/change_validator_test.go
everettraven: Should be done in https://github.com/carvel-dev/kapp/pull/921/commits/ce52021254d7ca39becbd0dd4711c11680b1db55

Had to make some slight modifications to some types in the `crdupgradesafety` package to export fields needed for testing

pkg/kapp/crdupgradesafety/change_validator.go
everettraven: Re-ordered in https://github.com/carvel-dev/kapp/pull/921/commits/ce52021254d7ca39becbd0dd4711c11680b1db55

pkg/kapp/crdupgradesafety/change_validator.go
everettraven: Updated in https://github.com/carvel-dev/kapp/pull/921/commits/ce52021254d7ca39becbd0dd4711c11680b1db55

rashmigottipati (approved):
Changes look good to me!

pkg/kapp/crdupgradesafety/change_validator.go
praveenrewar: Thank you!
Separate PR makes sense.

praveenrewar (approved):
Looks good!!

PR #922, opened 2024-04-11
Add validation to CRDUpgradeSafety preflight check to prevent removal of existing fields

### LINKED ISSUE #912: Add validation to `CRDUpgradeSafety` preflight check to prevent removal of existing fields
Now that there is a base `CRDUpgradeSafety` preflight check in place, we can continue adding validation logic based on the [`CRDUpgradeSafety` preflight check proposal](https://github.com/carvel-dev/carvel/blob/develop/proposals/kapp/001-crd-upgrade-safety/README.md).

This issue focuses on adding a validation to ensure that no existing fields are removed from a particular version of a CRD's schema during an upgrade operation. 

As a potential source of inspiration, here is how a couple of the existing validations are implemented:
- Definitions: https://github.com/carvel-dev/kapp/blob/82f2a30f59c60b85039a921c39ac9df23f1cb3af/pkg/kapp/crdupgradesafety/validator.go#L71-L93
- Consumption: https://github.com/carvel-dev/kapp/blob/82f2a30f59c60b85039a921c39ac9df23f1cb3af/pkg/kapp/crdupgradesafety/preflight.go#L36-L39

rashmigottipati:
Since the base CRDUpgradeSafety preflight check is in place and merged into the develop branch, we are at a point where we can continue adding validation logic based on the [CRDUpgradeSafety preflight check proposal](https://github.com/carvel-dev/carvel/blob/develop/proposals/kapp/001-crd-upgrade-safety/README.md).
The validation logic in this PR focuses on adding a check using the CRD schema checker to ensure that no existing fields are removed from a particular version of a CRD's schema during an upgrade operation.

Fixes https://github.com/carvel-dev/kapp/issues/912

```
Add validation to `CRDUpgradeSafety` preflight check to prevent removal of existing fields from a particular version of a CRD schema during an upgrade operation
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/crdupgradesafety/validator.go
+++ b/pkg/kapp/crdupgradesafety/validator.go
@@ -6,7 +6,9 @@ package crdupgradesafety
 import (
 	"errors"
 	"fmt"
+	"strings"
 
+	"github.com/openshift/crd-schema-checker/pkg/manifestcomparators"
 	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
@@ -87,6 +89,32 @@ func NoStoredVersionRemoved(old, new v1.CustomResourceDefinition) error {
 		if !newVersions.Has(storedVersion) {
 			return fmt.Errorf("stored version %q removed", storedVersion)
 		}
+	}
+
+	return nil
+}
+
+func NoExistingFieldRemoved(old, new v1.CustomResourceDefinition) error {
+	reg := manifestcomparators.NewRegistry()
+	err := reg.AddComparator(manifestcomparators.NoFieldRemoval())
+	if err != nil {
+		return err
+	}
+
+	results, errs := reg.Compare(&old, &new)
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+
+	errSet := []error{}
+
+	for _, result := range results {
+		if len(result.Errors) > 0 {
+			errSet = append(errSet, errors.New(strings.Join(result.Errors, "\n")))
+		}
+	}
+	if len(errSet) > 0 {
+		return errors.Join(errSet...)
 	}
 
 	return nil
--- a/pkg/kapp/crdupgradesafety/validator_test.go
+++ b/pkg/kapp/crdupgradesafety/validator_test.go
@@ -7,8 +7,9 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	v1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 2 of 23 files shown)


rashmigottipati:
cc @everettraven

pkg/kapp/crdupgradesafety/validator.go
```diff
@@ -91,3 +93,30 @@ func NoStoredVersionRemoved(old, new v1.CustomResourceDefinition) error {
...
+
+	errSet := []error{}
+
+	for _, result := range results {
+		if len(result.Errors) > 0 {
+			errSet = append(errSet, errors.New(strings.Join(result.Errors, "\n")))
+		}
+	}
+	if len(errSet) > 0 {
+		baseErr := errors.New("existing field in the CRD schema cannot be removed for safe CRD upgrades")
```
everettraven: Nit: I'm not sure this base error is needed. The top level validator this is intended to be used with should return which individual validator failed, and looking at the errors returned by the crd-schema-checker package for the NoFieldRemoval comparator it returns an error of the format `crd/%v version/%v field/%v may not be removed` which seems like it would be descriptive enough.

WDYT?

pkg/kapp/crdupgradesafety/validator_test.go
```diff
@@ -174,3 +175,104 @@ func TestNoStoredVersionRemoved(t *testing.T) {
...
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			shouldError: true,
+		},
```
everettraven, line 271: Could we also add a test that verifies that if a new version is added to the CRD with the field removed that it doesn't return an error? IIUC it should be OK to add an entirely new version to the CRD with a different set of fields

everettraven (approved):
LGTM!

pkg/kapp/crdupgradesafety/validator_test.go
rashmigottipati: That's a good idea, I've addd a test in my [latest commit](https://github.com/carvel-dev/kapp/pull/922/commits/51b64005ce47e75e822f08f1bda3e656320bf0c3) that verifies that if a new version is added to the CRD with the removed field, it wouldn't return an error.

pkg/kapp/crdupgradesafety/validator.go
rashmigottipati: Addressed in [latest commit](https://github.com/carvel-dev/kapp/pull/922/commits/51b64005ce47e75e822f08f1bda3e656320bf0c3) as well. Originally added it to be able to use it to assert in the e2e test but ended up using the error returned by crd-schema-checker package, so the `baseErr` is not required anymore.

praveenrewar (approved):
LGTM! Thank you for your contribution 🙏🏻 
I added some comments on Bryce's PR that maybe relevant to this as well (package name for test files and capitalising error messages), although no action item needed as of now.

PR #931, opened 2024-04-17
Add configuration to permissions pre-flight check to use `SelfSubjectAccessReview` or `SelfSubjectRulesReview`

### LINKED ISSUE #930: Add ability to configure API used for permission validation when using the permission validation pre-flight check
Currently, the permission validation pre-flight check uses the `SelfSubjectAccessReview` API for performing permission validation. This API is highly accurate, but also introduces load to the API server when run against a very large set of manifests (at least one request is made per resource, but can be more depending on circumstances). I want to ensure that when using this check I am not introduce significant load against the Kubernetes API server.

Allow a user to configure the permission validation pre-flight check to instead use the `SelfSubjectRulesReview` API for performing permission validation. When done with a cache where the namespace is the index, this can significantly reduce load on the Kubernetes API server.

This should be possible to configure through the recently added pre-flight configuration path

---
Vote on this request

everettraven:
`SelfSubjectAccessReview`s are great for this pre-flight check's use case, but can a bit intense on the API server (for every resource we are sending at least one request to the API server). For users who are concerned about the load placed on the API server, it would be nice to allow the option to use a `SelfSubjectRulesReview` instead as it will give the set of permissions a user has for a namespace. Using this API, and caching the results per namespace, results in less requests sent directly to the API server. 

This PR:
- Refactors the `permissions` package to have a pluggable `PermissionValidator` interface that has a `ValidatePermissions` function instead of using a single, inextensible function as existed previously.
- Adds `PermissionValidator` implementations that validates  permissions using `SelfSubjectAccessReview` and `SelfSubjectRulesReview`
- Refactors validators and preflight check as necessary to make use of the new `PermissionValidator` interface

Fixes #930 

```release-note
Add configuration option to the PermissionValidation pre-flight check to specify use of either the `SelfSubjectAccessReview` or `SelfSubjectRulesReview` APIs for performing permission validation. `SelfSubjectAccessReview`s are more accurate but result in higher load on the API server. `SelfSubjectRulesReview`s are less accurate but result in lower load on the API server.
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


praveenrewar:
@rashmigottipati @tmshort Do y'all want to review this first? In case you already have and it looks good to you, please drop an LGTM 😄

rashmigottipati:
@praveenrewar that sounds good, I'll make sure to review it today/tomorrow. :)

pkg/kapp/permissions/preflight.go
```diff
@@ -37,7 +52,27 @@ func (p *Preflight) SetEnabled(enabled bool) {
...
+	}
+
+	err = json.Unmarshal(cfgBytes, pCfg)
+	if err != nil {
+		return fmt.Errorf("parsing permissions preflight config: %w", err)
+	}
+
+	switch pCfg.PermissionValidatorType {
+	case PermissionValidatorTypeSelfSubjectAccessReview:
+	case PermissionValidatorTypeSelfSubjectRulesReview:
```
praveenrewar: nit: should we add a comment to indicate that the intention is to set a default value and to do validation?
```suggestion
        // Valid, do nothing
	case PermissionValidatorTypeSelfSubjectAccessReview, PermissionValidatorTypeSelfSubjectRulesReview:
```

pkg/kapp/permissions/validator.go
```diff
@@ -48,18 +65,104 @@ func ValidatePermissions(ctx context.Context, ssarClient authv1client.SelfSubjec
...
+		cache:      make(map[string][]rbacv1.PolicyRule),
+		mu:         sync.Mutex{},
+	}
+}
+
+// ValidatePermissons will validate permissions for a ResourceAttributes object using SelfSubjectRulesReview.
+// An error is returned if there are any issues creating a SelfSubjectRulesReview (i.e can't determine permissions)
+// or if the SelfSubjectRulesReview is evaluated and the caller does not have the permission to perform the actions
+// identified in the provided ResourceAttributes.
+func (rv *SelfSubjectRulesReviewValidator) ValidatePermissions(ctx context.Context, resourceAttrib *authv1.ResourceAttributes) error {
```
praveenrewar, line 99: (optional) nit: `resourceAttributes` instead of `resourceAttrib`? (Also applicable to ValidatePermissions on  SelfSubjectAccessReviewValidator)

praveenrewar (commented):
Changes look good (added some optional nits). Should we also add a test for `SelfSubjectRulesReview`. I think the existing ones only tests `SelfSubjectAccessReview` as that is used by default.

everettraven:
>Should we also add a test for SelfSubjectRulesReview. I think the existing ones only tests SelfSubjectAccessReview as that is used by default.

I think that is reasonable. I _think_ this might be possible via adding a kapp Config resource to the appropriate tests and changing the value there, which I hadn't thought of before. I'll try to have a change made soon-ish for this

praveenrewar:
@everettraven Just checking if you are still working on it or your forgot to push your changes. (No rush)

everettraven:
@praveenrewar thanks for following up! I was working on the changes a bit yesterday. I'll finish them up and push today

pkg/kapp/permissions/validator.go
everettraven: Ah, I missed making this change. I can make this change if needed, but functionally it shouldn't make a difference

pkg/kapp/permissions/preflight.go
everettraven: Updated in the latest push

everettraven:
@praveenrewar @100mik rebased and pushed. I can squash the commits if needed

rashmigottipati (approved):
Looks good to me!

praveenrewar (approved):
LGTM!

Issue #945, opened 2024-04-29
Cleanup: preflight checks


praveenrewar:
Recently we have introduced the preflight flag in kapp and also added a couple of checks. This issue tracks a few of the items that we didn't cover during the reviews and it would be good to do some clean up.
- [ ] Package name for test files should be *_test ([registry_test.go](https://github.com/carvel-dev/kapp/blob/develop/pkg/kapp/preflight/registry_test.go))
- [ ] Remove the usages of `RemoveClusterResource` in tests that are no longer required ([example](https://github.com/carvel-dev/kapp/blob/0e94ff15321f67ba737df346bd29e82cd62ee057/test/e2e/preflight_permission_validation_escalation_test.go#L84))
- [ ] In Carvel, we capitalise first character of error messages along with colons, this is used in the ui library to help with printing [multiline error messages](https://github.com/carvel-dev/kapp/blob/0e94ff15321f67ba737df346bd29e82cd62ee057/cmd/kapp/kapp.go#L45). We should aim to keep the error format consistent ([example](https://github.com/carvel-dev/kapp/blob/0e94ff15321f67ba737df346bd29e82cd62ee057/test/e2e/apply_wait_error_test.go#L208) of how the error looks like with indentation)
- [ ] In some scenarios, there might be more than one errors that we want to report, we use [semi structured error](https://github.com/carvel-dev/kapp/blob/0e94ff15321f67ba737df346bd29e82cd62ee057/pkg/kapp/clusterapply/waiting_changes.go#L143) to ensure that each individual error gets it's own line making the overall error more clear. ([example](https://github.com/carvel-dev/kapp/blob/0e94ff15321f67ba737df346bd29e82cd62ee057/test/e2e/formatted_error_test.go#L53) of how such an error message looks like)
- [ ] We seem to be adding a lot of e2e tests for preflight checks (which is good). Currently all these tests are in the test/e2e directory and most of them have a dedicated file (which is also good as # of lines per file are less). Should we move them to a dedicated directory like `test/e2e/preflights` so that the other e2e tests are not overshadowed by preflight checks?
- [ ] Fix handled() function for some of the validators ([details](https://github.com/carvel-dev/kapp/pull/950#discussion_r1588263642))

praveenrewar:
@everettraven Let me know if this looks good.

everettraven:
Looks great, thanks for writing this up!

rashmigottipati:
[...]

PR #950, opened 2024-05-02
Add handling of updates to default values to the CRD Upgrade Safety preflight check

rashmigottipati:
## What this PR does / why we need it:

This PR adds another field change validation to the CRD Upgrade Safety checks to prevent changes to a field's default values for a given CRD version during an upgrade and ensures the below:
- No new value can be added as default that did not previously have a default value present
- Default value of a field cannot be changed
- Existing default value for a field cannot be removed

https://github.com/carvel-dev/kapp/issues/949

```release-note
Add logic to the CRD Upgrade Safety pre-flight check to ensure that default values are not changed for a given CRD version during an upgrade
```
 
##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/crdupgradesafety/change_validator.go
+++ b/pkg/kapp/crdupgradesafety/change_validator.go
@@ -4,6 +4,7 @@
 package crdupgradesafety
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"reflect"
@@ -362,6 +363,44 @@ func MaximumPropertiesChangeValidation(diff FieldDiff) (bool, error) {
 		newMax := *diff.New.MaxProperties
 		if newMax < oldMax {
 			return handled(), fmt.Errorf("maximum properties constraint decreased from %+v to %+v", oldMax, newMax)
+		}
+		fallthrough
+	default:
+		return handled(), nil
+	}
+}
+
+// DefaultValueChangeValidation adds a validation check to ensure that
+// default values are not changed in a CRD schema:
+// - No new value can be added as default that did not previously have a
+// default value present
+// - Default value of a field cannot be changed
+// - Existing default value for a field cannot be removed
+// This function returns:
+// - A boolean representation of whether or not the change
+// has been fully handled (i.e. the only change was to a field's default value)
+// - An error if either of the above criteria are not met
+func DefaultValueChangeValidation(diff FieldDiff) (bool, error) {
+	handled := func() bool {
+		diff.Old.Default = &v1.JSON{}
+		diff.New.Default = &v1.JSON{}
+		return reflect.DeepEqual(diff.Old, diff.New)
+	}
+
+	switch {
+	case diff.Old.Default == nil && diff.New.Default != nil:
+		newDefault := diff.New.Default
+		return handled(), fmt.Errorf("new value added as default when previously no default value existed: %+v", newDefault)
+
+	case diff.Old.Default != nil && diff.New.Default == nil:
+		oldDefault := diff.Old.Default.Raw
+		return handled(), fmt.Errorf("default value has been removed when previously a default value existed: %+v", oldDefault)
+
+	case diff.Old.Default != nil && diff.New.Default != nil:
+		oldDefault := diff.Old.Default.Raw
+		newDefault := diff.New.Default.Raw
+		if !bytes.Equal(diff.Old.Default.Raw, diff.New.Default.Raw) {
+			return handled(), fmt.Errorf("default value has been changed from %+v to %+v", oldDefault, newDefault)
 		}
 		fallthrough
 	default:
--- a/test/e2e/preflight_crdupgradesafety_invalid_field_change_default_added_test.go
+++ b/test/e2e/preflight_crdupgradesafety_invalid_field_change_default_added_test.go
@@ -1,0 +1,118 @@
+// Copyright 2024 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Adding a new value as default for a field that did not
+// previously have a default value present is invalid
+// and this test is covering that case.
+func TestPreflightCRDUpgradeSafetyInvalidFieldChangeDefaultAdded(t *testing.T) {
+	env := BuildEnv(t)
+	logger := Logger{}
+	kapp := Kapp{t, env.Namespace, env.KappBinaryPath, logger}
+
+	testName := "preflightcrdupgradesafetyinvalidfieldchangedefaultadded"
+
+	base := `
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.13.0
+  name: memcacheds.__test-name__.example.com
+spec:
+  group: __test-name__.example.com
+  names:
+    kind: Memcached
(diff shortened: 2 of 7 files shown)


rashmigottipati:
cc @everettraven

pkg/kapp/crdupgradesafety/change_validator.go
```diff
@@ -241,6 +241,40 @@ func MinimumPropertiesChangeValidation(diff FieldDiff) (bool, error) {
...
+
+	switch {
+	case diff.Old.Default == nil && diff.New.Default != nil:
+		return handled(), fmt.Errorf("new value added as default when previously no default value existed: %+v", diff.New.Default.Raw)
+
+	case diff.Old.Default != nil && diff.New.Default == nil:
+		return handled(), fmt.Errorf("default value has been removed when previously a default value existed: %+v", diff.Old.Default.Raw)
+
+	case diff.Old.Default != nil && diff.New.Default != nil:
+		if !reflect.DeepEqual(diff.Old.Default.Raw, diff.New.Default.Raw) {
```
everettraven: `slices.Equal` or `bytes.Equal` probably makes more sense to use here than `reflect.DeepEqual`. Reflection should only be used when necessary. The reason we use it when dealing with the full `v1.JSONSchemaProps` resources is because we need to ensure that the entire object is equivalent as opposed to comparing a specific field (which is significantly more complex to write and generalize for an API that can change)

pkg/kapp/crdupgradesafety/change_validator.go
```diff
@@ -241,6 +241,40 @@ func MinimumPropertiesChangeValidation(diff FieldDiff) (bool, error) {
...
+func DefaultValueChangeValidation(diff FieldDiff) (bool, error) {
+	handled := func() bool {
+		diff.Old.Default = &v1.JSON{}
+		diff.New.Default = &v1.JSON{}
+		return reflect.DeepEqual(diff.Old, diff.New)
+	}
+
+	switch {
+	case diff.Old.Default == nil && diff.New.Default != nil:
+		return handled(), fmt.Errorf("new value added as default when previously no default value existed: %+v", diff.New.Default.Raw)
```
everettraven: does using the value from the diff object after calling the `handled()` function in the return result in this returning an empty value?

everettraven: I think that we may have some other places in the code this may be a problem but I will plan to address them as part of #945

pkg/kapp/crdupgradesafety/change_validator.go
rashmigottipati: Ah that's good to learn, I've updated this to use `bytes.Equal` in the [latest commit](https://github.com/carvel-dev/kapp/pull/950/commits/2373d472df0407c0c7a718acc829dad5471f1cda)

pkg/kapp/crdupgradesafety/change_validator.go
rashmigottipati: Yes, good catch! I've fixed for now just for this particular change in the [latest commit](https://github.com/carvel-dev/kapp/pull/950/commits/2373d472df0407c0c7a718acc829dad5471f1cda), will capture it as an item as a part of https://github.com/carvel-dev/kapp/issues/945 for all other instances.

It looks like it isn't possible to ignore one field with `reflect.DeepEqual` and we would need to do a manual comparison, so I used extra variables for the swap.

test/e2e/preflight_crdupgradesafety_invalid_field_change_default_added_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Adding a new value as default for a field that did not
+// previously have a default value present is invalid
+// and this test is covering that case.
+func TestPreflightCRDUpgradeSafetyInvalidFieldChangeDefaultAdded(t *testing.T) {
```
100mik, line 16: nit: Can we move all the new tests to one file (`preflight_crdupgradesafety_defaultvalue_test.go` maybe) and have different methods?

100mik (commented):
Just a minor nit!

test/e2e/preflight_crdupgradesafety_invalid_field_change_default_added_test.go
praveenrewar: It will be covered as part of #945.

pkg/kapp/crdupgradesafety/change_validator.go
rashmigottipati: Reversed this change in the latest commit

everettraven: I think it probably makes sense to address this as part of this PR to prevent additional work in #945 now that we have identified this as a potential issue.

I would recommend doing something like this:
 ```suggestion
                 newDefault := *diff.New.Default.Raw
		return handled(), fmt.Errorf("new value added as default when previously no default value existed: %+v", newDefault)
```

While this introduces a new, temporary, variable which isn't the most ideal it results in a complete error message. IMO a complete error message is more important than reducing temporary variables because it provides a nicer user experience.

Comment (everettraven) [reply pkg/kapp/crdupgradesafety/change_validator.go]: @rashmigottipati It won't let me unresolve this thread, so tagging you directly

File: test/e2e/preflight_crdupgradesafety_valid_field_change_default_no_change_test.go [hunk 4]
```diff
@@ -0,0 +1,117 @@
...
+	base = strings.ReplaceAll(base, "__test-name__", testName)
+	appName := "preflight-crdupgradesafety-app"
+
+	cleanUp := func() {
+		kapp.Run([]string{"delete", "-a", appName})
+	}
+	cleanUp()
+	defer cleanUp()
+
+	update := `
```
everettraven, line 70: This "updated" CRD looks the same. Do we know for sure that this would trigger an "upgrade" operation via kapp since there is no diff (if I'm missing the diff please let me know!)? If it does _not_ trigger an upgrade operation via kapp then I'm not sure this test is really providing any value. Maybe there is another way we can update this in a way that it triggers an "upgrade"?

 Simplest one that comes to my mind is maybe updating the annotations to force a diff

pkg/kapp/crdupgradesafety/change_validator.go
rashmigottipati: Addressed in [latest commit](https://github.com/carvel-dev/kapp/pull/950/commits/97a258940768df26f84ba5297362f15de661d0da)

test/e2e/preflight_crdupgradesafety_valid_field_change_default_no_change_test.go
rashmigottipati: You're right, there's no diff so I don't think it will trigger an upgrade operation. I've forced the upgrade in the [latest commit](https://github.com/carvel-dev/kapp/pull/950/commits/97a258940768df26f84ba5297362f15de661d0da) via a change in annotation as you'd suggested.

praveenrewar:
@rashmigottipati Since I merged the other PR, this now has some merge conflicts.

everettraven (approved):

rashmigottipati:
Thanks @everettraven for the approval! 

@praveenrewar Thanks for merging the other PR. Rebased this PR and squashed commits.

100mik (approved):
Looks good to me!
@praveenrewar merge away if you are fine with this

praveenrewar (approved):
LGTM!

PR #996, opened 2024-08-05
Add filter for apps for which ownership can be overriden

### LINKED ISSUE #993: Selectively override ownership while using --dangerous-override-ownership-of-existing-resources
I want my `kapp` app to snatch ownership from existing resources belonging to another app. 
However, I only want this to happen if the  resource belongs to specific apps.

i would like to able to pick which `kapp` apps I snatch ownership from. Probably by using a "filter-in" filter.
For example. `kapp deploy --ownership-override-allowed-apps "oldapp,olderapp" --dangerous-override-ownership-of-existing-resources`.

We will run into a few limitations here.
First, as of today we cannot easily search for an "app name" if I know the app ID on a resource, without listing all the apps.
Secondly, to avoid privilege escalations, for the time being we will only be able to use this filter for apps in the same namespace.

---
Vote on this request

100mik:
Introduces a flag which allows users to specify flags which apps ownership can be snatched from while allowing ownership overrides. 

Fixes #993 

```release-note
Introduce flag to filter out apps which we can steal ownership from while deploying
```

The only catch today, is that it is a best effort listing of apps, limited to the namespace the new app itself is being deployed in. This is in line with encouraged kapp-controller patterns. 

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/cmd/app/deploy.go
+++ b/pkg/kapp/cmd/app/deploy.go
@@ -382,15 +382,27 @@ func (o *DeployOptions) existingResources(newResources []ctlres.Resource,
 		return ""
 	}
 
+	labelValAppMapResolver := func() map[string]string {
+		items, _ := apps.List(nil)
+		appLabelMap := map[string]string{}
+		for _, item := range items {
+			meta, _ := item.Meta()
+			appLabelMap[meta.LabelValue] = item.Name()
+		}
+		return appLabelMap
+	}
+
 	matchingOpts := ctlres.AllAndMatchingOpts{
 		ExistingNonLabeledResourcesCheck:            o.DeployFlags.ExistingNonLabeledResourcesCheck,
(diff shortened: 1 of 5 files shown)


praveenrewar (approved):
LGTM!
I am just trying to think if we can have a better name for the flag, otherwise looks good.

100mik:
Some options I had thought of, `ownership-override-allowed-apps`, `overridable-apps` (not a fan), `ownership-override-filter`
What do you think of these? I quite like the first one, since it is technically not a filter. Filters have mainly been for user supplied resources.

100mik:
From a convo offline, we are going with `ownership-override-allowed-apps` 🙌🏼

100mik:
Created #997 and #998 to track discussions that happened while we were working on this

PR #1001, opened 2024-08-12
KAPP-1000: Apply noop changes after everything else

### LINKED ISSUE #1000: Noop changes should observe change rules for upserting, or be performed last
I have a package that, among other things, include a kapp-controller PackageInstall which looks like this (superfluous information omitted):
```yaml
apiVersion: packaging.carvel.dev/v1alpha1
kind: PackageInstall
metadata:
  annotations:
    kapp.k14s.io/change-group: ccp.ipbps.cgi.com/contour
    kapp.k14s.io/change-rule.0: upsert after upserting ccp.ipbps.cgi.com/packages
    kapp.k14s.io/change-rule.1: delete before deleting ccp.ipbps.cgi.com/packages
  name: contour
spec:
  packageRef:
    refName: contour.tanzu.vmware.com
    versionSelection:
      constraints: 1.26.1+vmware.1-tkg.1
...
```
Together with a kapp-controller Package that looks like this (again, superfluous information omitted):
```yaml
apiVersion: data.packaging.carvel.dev/v1alpha1
kind: Package
metadata:
  annotations:
    kapp.k14s.io/change-group: ccp.ipbps.cgi.com/packages
  name: contour.tanzu.vmware.com.1.26.1+vmware.1-tkg.1
spec:
...
```

Then I update my package with a new version, that changes the version constraint in the PackageInstall and **removes** the Package with the old version as it simultaneously **adds** a new Package with the new version. There are some other changes in the package as well, among them some updated Secrets.

**What happened:**
The change to the PackageInstall happens immediately, and then it fails due to the fact that the new Package isn't available. The new Package won't be installed, since it awaits the Secrets to be deployed first. Then it gets stuck in a loop where the PackageInstall has a noop operation that fails due to the Package not existing, and the Package won't be created since it is in the next step in the dependency graph.

**What did you expect:**
Primarily I would expect the update to the PackageInstall to happen after the Package would be created, since the change rule indicates this. When the problem has already occurred and the PackageInstall is failing, I would expect the noop operation to observe the change rules so it's failure wouldn't stop reconciliation unless it fails after the Packages it is dependent on is ready.

Another workaround would be to handle all noops after the other changes in the change set - something I've implemented in a fork and which solved the problem for me. I'll add the PR for that once this issue is created

No

**Environment:**

- kapp version (use `kapp --version`): 0.63.2
- OS (e.g. from `/etc/os-release`): Ubuntu 22.04
- Kubernetes version (use `kubectl version`): v1.28.4+vmware.1

---
Vote on this request

OlofKalufs:
This is a workaround for the problem when a noop apply fails due to dependencies that is later in the dependency stage (even though the noop is declared to be dependent on them)

Fixes #1000 

```release-note
I'm not sure of the implications of this,
```

##### Review Checklist:

  a link to that PR
  change

```docs

```

### DIFF
--- a/pkg/kapp/clusterapply/applying_changes.go
+++ b/pkg/kapp/clusterapply/applying_changes.go
@@ -60,25 +60,28 @@ func (c *ApplyingChanges) Apply(allChanges []*ctldgraph.Change) ([]WaitingChange
 		// - "...: context canceled (reason: )"
 		applyThrottle := util.NewThrottle(c.opts.Concurrency)
 		applyCh := make(chan applyResult, len(nonAppliedChanges))
+		allChangesAreNoop := c.allChangesAreNoop(nonAppliedChanges)
 
 		for _, change := range nonAppliedChanges {
-			change := change // copy
+			if allChangesAreNoop || (change.Change.Op() != ctldgraph.ActualChangeOpNoop) {
+				change := change // copy
 
-			go func() {
-				applyThrottle.Take()
-				defer applyThrottle.Done()
+				go func() {
+					applyThrottle.Take()
+					defer applyThrottle.Done()
 
-				clusterChange := change.Change.(wrappedClusterChange).ClusterChange
-				retryable, descMsgs, err := clusterChange.Apply()
+					clusterChange := change.Change.(wrappedClusterChange).ClusterChange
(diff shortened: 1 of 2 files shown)


praveenrewar:
Closing this as the issue is closed, and we might need to iron out the solution before we make a change.

PR #1008, opened 2024-08-30
Make ownership override allowed apps flag dangerous

100mik:

This PR makes it so that `--dangerous-override-ownership-of-existing-resources` is no longer required to do a scoped override.
The flag for scoped override - `--dangerous-override-allowed-apps` is not itself a dangerous flag which accepts a list of apps that ownership can be snatched from.

```release-note
We need to update documentation with the new help message and flag name
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


praveenrewar (commented):
The changes looks good. I think we don't have a test for the existing `--dangerous-override-ownserhip-of-existing-resources`, can we please add one to ensure that there is no regression.

praveenrewar:
Also, please sign you commit :)

100mik:
Done and done!

praveenrewar (approved):

PR #1059, opened 2025-04-28
Fix CVEs and bump go version

devacts:
# What this PR does / why we need it:

- This PR bumps golang version to 1.27.8 and other modules to fix CVEs
- It fixes failing test cases in `test/e2e/wait_timeout_test.go`

## Which issue(s) this PR fixes:
All e2e test cases run on a minimum (1.27) and latest k8s release clusters.
After K8s 1.33 release, 2 tests were failing in the the latest k8s e2e test run.
The 2 tests were related to custom waitRules and checked the happy path as well as the timeout in waitRule introduced in PR https://github.com/carvel-dev/kapp/pull/994

### Why did the tests start failing?
Though the tests mainly check the timeout in custom waitRules, the config used for testing was following:

```
kind: Config
waitRules:
  conditionMatchers:
  - type: Ready
    status: "True"
    success: true
  - type: ContainersReady
    status: "False"
    timeout: 50s
  resourceMatchers:
  - apiVersionKindMatcher: {apiVersion: v1, kind: Pod
```
[Link](https://github.com/carvel-dev/kapp/blob/develop/test/e2e/wait_timeout_test.go#L48C1-L60C54)

This config had `supportsObservedGeneration` set to true as a waitRule.
When this waitRule is set to true, kapp checks that metadata.generation and status.observedGeneration values should be same: https://github.com/carvel-dev/kapp/blob/develop/pkg/kapp/resourcesmisc/custom_waiting_resource.go#L62
`s.waitRule.SupportsObservedGeneration && obj.Metadata.Generation != obj.Status.ObservedGeneration`

However, for pods observedGeneration was not supported till k8s release 1.32.x
With 1.33 release, [pods too support supportsObservedGeneration](https://github.com/kubernetes/kubernetes/pull/130650), but this needs to be enabled by a feature gate `PodObservedGenerationTracking`  for a cluster.
When this FG is enabled, then writers to pod status will update the status.observedGeneration.
[meatadata.generation is now also auto incremented](https://github.com/kubernetes/kubernetes/pull/130181) with each spec change in pods with k8s 1.33 release.

The kind cluster however, does not by default enable the `PodObservedGenerationTracking` feature flag, so even though the waitRules in config say `supportsObservedGeneration` is `true`, the observedGeneration is not updated as  PodObservedGenerationTracking is not set.
Due to this metadata.generation is incremented but observedGeneration is null, which breaks the [check](https://github.com/carvel-dev/kapp/blob/develop/pkg/kapp/resourcesmisc/custom_waiting_resource.go#L62).

This is the reason the 2 tests started to fail after k8s 1.33 release on latest cluster e2e run.

### Why did the test cases not fail earlier if observedGeneration was not supported for pods?
Prior to k8s 1.33, neither metadata.generation nor status.observedGeneration were updated, so both fields had value null.
Hence even the waitRules in the test said supportsObservedGeneration is true, the equality check for the two value returned true as both are null. So, the tests were passing.

### How has this been fixed?
kapp still supports supportsObservedGeneration for waitRules including pod but supportsObservedGeneration should be set to true for a pod resource only when the FG PodObservedGenerationTracking is enabled.
hence in tests or otherwise, this waitRules should be exercised with caution when using for pods.
Ideally, the tests in question were introduced to check timeout in custom waitRules and supportsObservedGeneration was not a required check.
Therefore the waitrule has been removed in the tests

```release-note
NONE
```

##### Review Checklist:

  a link to that PR
  change

```docs

```


joaopapereira (approved):
LGTM

